PR #2, opened 2016-08-16
Update Gradle wrapper to 2.14.1

jamierocks:
- Rename plugin to 'net.fabricmc.loom' from 'fabric-loom'

- Add .gitattributes

- Use the same .gitignore from fabric-base

- Add license headers

- Use -SNAPSHOT for builds


shadowfacts:
If you're going to update Gradle, update everything to 3.0?

build.gradle
```diff
@@ -1,26 +1,30 @@
...
+
+    id 'net.minecrell.licenser' version '0.2.1'
 }
 
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+group = 'net.fabric'
+archivesBaseName = project.name.toLowerCase()
+version = '0.0.1-SNAPSHOT'
```
modmuss50: I dont see the reason behind doing this. Can you explain this?

shadowfacts: http://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it

jamierocks:
3.0 is not released.

shadowfacts:
https://gradle.org/gradle-download/

> The latest release of Gradle is version 3.0, released on 15th August 2016.

modmuss50:
Have you tested it on 3?

jamierocks:
I tested building which works.

modmuss50:
We we are going to use gradle 3.0, all the projects should use it, so I will test this, with the base.

I am slightly wary of new things

PR #8, opened 2016-11-05
Fixed Deobf Typo

lclc98:


modmuss50:
Not sure how I didn't notice this, I will merge it but it will break things.

src/main/java/net/fabricmc/loom/util/ModRemapper.java
```diff
@@ -78,15 +78,15 @@ public static void remap(Project project) throws IOException {
 			.build();
 
 		OutputConsumerPath outputConsumer = new OutputConsumerPath(modJar.toPath());
-		//Rebof the debof jar
-		outputConsumer.addNonClassFiles(debofJar.toPath());
-		remapper.read(debofJar.toPath());
+		//Rebof the deobf jar
```
MrIbby: Rebof

PR #9, opened 2017-07-29
Start on access manipulator's

modmuss50:
This PR contains the stuff needed to allow mods/fabric-base to change the access of members of a class. This is needed in a few places to remove some old mixins that used to add getter + setters. This will also be needed for POMF to allow it to run without some janky make everything public transformer (see: https://github.com/FabricMC/fabric-base/blob/master/src/main/java/net/fabricmc/base/transformer/AccessTransformer.java)

I have made this a PR as I would like some input on the format of file. Im not 100% happy with it but it does the job. Here is an example of what it could look like: https://gist.github.com/modmuss50/ef61e4e8e5056adae2fe203840232520

The format has support for different mappings, in the future once intermediary mappings are in places this can be removed. I may look into making this a more compact format (based of tiny mappings?)


exaskye:
I think JSON is a bad fit for this kind of data. A plaintext file format like Tiny is probably much better suited.

PR #11, opened 2018-11-02
add gradle plugin marker

NikkyAI:
format build.gradle

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,9 @@
 plugins {
-    id 'java'
-    id 'groovy'
-    id 'maven'
-    id 'maven-publish'
-    id 'idea'
-    id 'eclipse'
+	id 'java'
+	id 'maven-publish'
+	id 'java-gradle-plugin'
+	id 'idea'
+	id 'eclipse'
 }
 
 sourceCompatibility = 1.8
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -79,30 +75,37 @@ task wrapper(type: Wrapper) {
...
+            url "http://mavenupload.modmuss50.me/"
+            if (project.hasProperty('mavenPass')) {
+                credentials {
+                    username 'buildslave'
+                    password project.getProperty('mavenPass')
+                }
+            }
+        }
+    }
+}
```
jamierocks, line 110: The newline should be retained, as should the tabs :)

It would seem that the buildscript currently uses a mix of tabs and spaces, so doing a format here could make sense :)

build.gradle
```diff
@@ -1,10 +1,10 @@
 plugins {
-    id 'java'
-    id 'groovy'
-    id 'maven'
-    id 'maven-publish'
-    id 'idea'
-    id 'eclipse'
+    id("java")
```
jamierocks: This isn't how the Gradle DSL is intended to be used, and furthermore that style of method calling is not used throughout the buildscript.

As you mention on IRC, this could be used in preparation of migrating to the Kotlin DSL - however, we've yet to make a decision on whether we are moving to that, and regardless until we do so we should maintain the standards of what we're using.

build.gradle
```diff
@@ -1,10 +1,10 @@
...
+    id("groovy")
```
jamierocks: We're not actually using groovy for this plugin, so this should be safe to remove.

jamierocks (changes requested):

build.gradle
NikkyAI: i added a .editorconfig and reverted it to tabs only

build.gradle
NikkyAI: reverted that to groovy style and groovy strings

build.gradle
copygirl: Still lacking newline at the end..?

PR #18, opened 2018-11-08
add mod extension

NikkyAI:
implemented are id and version properties as a proof of concept

this can be used like

```kotlin
mod {
    id = Constants.modid
    version = Constants.version + "-" + (System.getEnv("BUILD_NUMBER") ?: "local")
}
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,3 +3,6 @@ indent_style = tab
 
 [*.gradle]
 indent_style = tab
+
+[*.java]
+indent_style = tab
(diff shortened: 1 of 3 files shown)


asiekierka:
The problem I have with this approach is that the mod.json can include many, *many* more fields (check fabric-loader's ModInfo class!) than those two, setting aside the fact I wouldn't be surprised if people were going to make custom extensions.

If anything, I'd be keen on adding a way to propagate the version from Gradle to an existing mod.json, not generate a new mod.json like this and force us to maintain the structure in an additional, distinct location.

PR #19, opened 2018-11-08
add mod extension for editing mod.json

NikkyAI:
<s>this currently only works on a existing `mod.json`

properties that can be set: `id`, `version`

planned: everything else

properties currently just override what happened to be in the mod.json, this could be changed to allow appending or prepending values
I think this would be overengineered
such things can also just stay in the `build.gradle`
</s>

new approach: dumb token replacement

available methods on the mod extension:
- `replace(String key, String value)`
- `replace(Map<String, String> replacements)`

sample usage

```kotlin
mod {
   replace("VERSION", version + "-" + buildNumber)
}
```

`mod.json`
```json
{
  "id": "fabric-language-kotlin",
  "version": "VERSION"
}
```

### DIFF
--- a/src/main/java/net/fabricmc/loom/ModExtension.java
+++ b/src/main/java/net/fabricmc/loom/ModExtension.java
@@ -30,6 +30,6 @@ public class ModExtension {
     }
 
     public boolean getEnabled() {
-    	return id != null && version != null;
+    	return id != null || version != null;
 	}
 }
--- a/src/main/java/net/fabricmc/loom/ModExtension.java
+++ b/src/main/java/net/fabricmc/loom/ModExtension.java
@@ -1,0 +1,31 @@
+package net.fabricmc.loom;
+
+import org.gradle.api.Project;
+
+public class ModExtension {
+    private final Project project;
+    private String id = null;
+    private String version = null;
+    // TODO: add more properties
+    // TODO: add dependencies
+
+    public ModExtension(Project project) {
+        this.project = project;
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public String getVersion() {
+        return version;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
+    }
+}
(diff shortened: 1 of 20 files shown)


src/main/java/net/fabricmc/loom/ModExtension.java
```diff
@@ -0,0 +1,35 @@
+package net.fabricmc.loom;
+
+import org.gradle.api.Project;
+
+public class ModExtension {
+    private final Project project;
```
jamierocks: Just need to fix the formatting on this file (tabs).

jamierocks (changes requested):
Looking good :+1:

src/main/java/net/fabricmc/loom/ModExtension.java
NikkyAI: intereesting.. tabs and spaces seem to be used interchangeably in the project
net.fabricmc.loom.util.MixinRefmapHelper for example is spaces ?

NikkyAI: i fixed that and all other occurences of 4 spaces that should be tabs

NikkyAI:
sadly i cannot let idea autoformat because then it would basically touch every file.. i think someone should do that at some point though.. and maybe get a official fabric codestyle together
i suggest the intellij idea defaults with star imports banned

jamierocks:
There is one - see [fabric-docs](https://github.com/FabricMC/fabric-docs/blob/master/guidelines/intellij/FabricFormat.xml).

NikkyAI:
there is 2 other options we have for implementing this

:one:. do dumb string replacements
`$VERSION` -> `0.1.0-12`

:two:. pass a `Map<String, Any>`
  and handle Strings, Numbers and Maps recursively
  this options would probably be unable to cleanly extend a already existing map in the json

NikkyAI:
i implemented option :one: for now

ChloeDawn:
Why is this needed for the json? Gradle provides `processResources` for stuff like this

ChloeDawn:
```groovy
processResources {
  from(sourceSets.main.resources.srcDirs) {
    include 'mod.json'
    expand version: project.version
  }
  from(sourceSets.main.resources.srcDirs) {
    exclude 'mod.json'
  }
}
```
```json
{
  "id": "foobar",
  "name": "FooBar",
  "version": "$version",
  "side": "universal"
}
```

NikkyAI:
This gets handled just fine by gradle iself just as @InsomniaKitten pointed out

so i see no reason to keep this PR open

PR #24, opened 2018-12-11
Add -XstartOnFirstThread, required when running on macOS

shadowfacts:
Required when running on macOS, otherwise GLFW crashes trying to create a window:

```
java.lang.ExceptionInInitializerError
	at org.lwjgl.glfw.GLFW.glfwCreateWindow(GLFW.java:1823)
	at net.minecraft.client.util.Window.<init>(Window.java:88)
	at net.minecraft.class_3682.method_16038(class_3682.java:25)
	at net.minecraft.client.MinecraftClient.init(MinecraftClient.java:416)
	at net.minecraft.client.MinecraftClient.start(MinecraftClient.java:357)
	at net.minecraft.client.main.Main.main(Main.java:126)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at net.minecraft.launchwrapper.Launch.launch(Launch.java:135)
	at net.minecraft.launchwrapper.Launch.main(Launch.java:28)
Caused by: java.lang.IllegalStateException: GLFW windows may only be created on the main thread and that thread must be the first thread in the process. Please run the JVM with -XstartOnFirstThread. For offscreen rendering, make sure another window toolkit (e.g. AWT or JavaFX) is initialized before GLFW.
	at org.lwjgl.glfw.EventLoop$OffScreen.<clinit>(EventLoop.java:39)
	... 12 more
```

Right now, this isn't macOS specific, as there should be no harm in doing so on other platforms as well, and there's no need to unnecessarily complicate the code.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/RunClientTask.java
+++ b/src/main/java/net/fabricmc/loom/task/RunClientTask.java
@@ -79,6 +79,7 @@ public class RunClientTask extends JavaExec {
 		LoomGradleExtension extension = this.getProject().getExtensions().getByType(LoomGradleExtension.class);
 		List<String> args = new ArrayList<>();
 		args.add("-Dfabric.development=true");
+		args.add("-XstartOnFirstThread");
 		return args;
 	}
 
--- a/src/main/java/net/fabricmc/loom/task/RunServerTask.java
+++ b/src/main/java/net/fabricmc/loom/task/RunServerTask.java
@@ -75,6 +75,7 @@ public class RunServerTask extends JavaExec {
 		LoomGradleExtension extension = this.getProject().getExtensions().getByType(LoomGradleExtension.class);
 		List<String> args = new ArrayList<>();
 		args.add("-Dfabric.development=true");
+		args.add("-XstartOnFirstThread");
 		return args;
 	}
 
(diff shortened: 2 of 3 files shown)


asiekierka:
Oke! But first @modmuss50 needs to fix the runconfigs so they're... a bit smarter with adding in patches like this. Neither "never override" nor "always override" will do - I think you should parse the argstring and diff the argstring, so to say.

modmuss50:
Yes, this PR looks fine, and someone has come across this issue.

Just doing a simple diff on the arg string would mean custom args would get overridden. One option would be to generate a wrapper around the main class to handle this but that seems a bit unnecessary. Another option would be to just add a task to reset the run args, let me know what you think.

modmuss50:
I think this should be merged, as if you are having issues re-setting up isnt that bad. And it wont hurt not to have it.

Issue #35, opened 2018-12-14
Rebuilding Yarn in local repo does not cause rebuilt minecraft.jar 


codewarrior0:
When yarn is built outside of the usual CI system and installed in the local Maven repo, its jar is assigned a name like "yarn-18w50a-local.jar". This name does not change if yarn is rebuilt and installed again. Thus, after rebuilding yarn a second time, the changes are not detected when a mod (that is configured to fetch yarn from the local repo) is built with `--refresh-dependencies`, and so the `minecraft-18w50a-mapped-local.jar` is not rebuilt with the new mappings. To force a proper rebuild, both the mapped jar and the mappings must be removed from the cache folder.

The loom mechanisms (MinecraftMappedProvider, MappingsProvider, et al) for extracting the mappings from the yarn jar and rebuilding the mapped minecraft jar using those mappings only rely on the version numbers (or rather, filenames) of the mappings and the minecraft jar to decide whether to rebuild the mapped minecraft jar.

modmuss50:
running with `--refresh-dependencies` in 0.4+ will now do this.

PR #36, opened 2018-12-14
CleanLoomBinaries now removes mappings too

codewarrior0:
With this, I can now run `gradle cleanLoomBinaries` after rebuilding yarn into my local Maven repo, to cause the next `gradle build` to rebuild the mapped jar using the updated mappings.

Kind of a solution to #35


asiekierka:
Ah, I see.

Can't we make --rerun-tasks do the trick?

codewarrior0:
Not without turning all those DependencyProvider things into proper Gradle tasks, I imagine.

modmuss50:
Would it be worth also deleting some more stuff such as the mc jars? It might as well clean up everything?

asiekierka:
Separate command to clear everything. This is supposed to clear Loom processing outputs.

PR #52, opened 2019-01-09
Prevent remapJar task parameters from being overwritten

falkreon:
The best (well, most stable, safest from race conditions) way to get shadowJar and remapJar to play nicely is to set `jar` to the output from the shadow task and set dependsOn's accordingly. Something like,
```java
remapJar {
	jar = tasks.shadowJar.outputs.getFiles()[0]
}
```
...we can get into solutions for `.getFiles()[0]` later, but for now the issue is, the above closure doesn't work. `Task::apply` is run *after* the project configurations are run, meaning we always have the default jar parameter from `configureCompile`. Working around requires a kludge like,
```java
tasks.jar.doLast {
	tasks.remapJar.jar = tasks.shadowJar.outputs.getFiles()[0]
}
```
in order to run configuration code after `Task::apply`

This felt like the minimal diff that solves this problem. If access to the default setting of the jar param is desired, we could talk about hoisting the parameter to before `apply`, but that'd be more research, some side effects (it might not default to the output of a reconfigured jar task), and a bigger diff.


modmuss50:
Seems simple enough, thank you 👍

PR #64, opened 2019-04-07
[WIP] Start on 0.2.1

modmuss50:
Ideally this would have gone into 0.3 but there are still some issues with that so getting it into 0.2.1 will be quicker.

- [x] Allow nesting of jars, from maven or sub project

- [x] Mod compile should expand out nested jars

Dont merge this, when ready I can push to a new branch.


modmuss50:
Ok, I think this is all done now

modmuss50:
Pushed to a new branch

PR #66, opened 2019-04-09
Fix game launching in wrong folder with VS Code

copygirl:
It appears that a backslash in the working directory (in `launch.json`) is not correctly handled by the Java launcher or so, causing the game to run in the workspace folder instead of `run`.

**Note:** I don't have a Windows machine so please make sure this still works on there.

(edit: Also let me know if this is the right branch.)


modmuss50:
Branch is fine, I might pull this back into 0.2.0 as well. I will test on windows

modmuss50:
Works fine on windows, thanks. I will also back port this to 0.2

PR #72, opened 2019-04-18
Fix the handling of file paths

howtonotwin:
Some tasks used the `File(String)` constructor to attempt to open files in the project directory. Gradle does not have to (and it doesn't) use the project directory as the working directory. As a result, these tasks were broken. Switch to the correct way of creating `File`s under Gradle,
`Project#file`.


asiekierka (approved):
@modmuss50 ?

asiekierka:
The remaining changes have been done as part of a larger refactor a few commits ago.

PR #73, opened 2019-04-19
Download less each run

Chocohead:
Right now Loom doesn't cache the version manifests at all, and relies on the SHA-1 hashes for the client and server jars, and assets. This stores ETags and last modified times for all manifests, jars and assets. An average run will only save a little over 100kb of downloading from what I can tell, but every little helps right?

This doesn't remove the hash checking which could improve performance some more if disk access was proving a bigger bottleneck than networking. Can't see that it typically would though unless someone had a very slow HDD and CPU compared to their internet.

### DIFF
--- a/src/main/java/net/fabricmc/loom/providers/MinecraftLibraryProvider.java
+++ b/src/main/java/net/fabricmc/loom/providers/MinecraftLibraryProvider.java
@@ -28,11 +28,11 @@ import com.google.gson.Gson;
 import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.util.Checksum;
 import net.fabricmc.loom.util.Constants;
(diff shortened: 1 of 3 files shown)


asiekierka:
You're putting too much effort into the 0.2.x branch, Chocohead...

Chocohead:
You're over inflating how much effort it is to fix these things like this. So long as it's not having to deal with Gradle's internals it's not that different from achieving the same thing in any other Java project. As soon as you start having to deal with Gradle directly things get scary, which appears to be the root of many of 0.3's problems.

asiekierka (approved):
@modmuss50 LGTM, please review

modmuss50 (approved):
Looks fine to me, thanks for the help with this mess ;D

PR #77, opened 2019-04-20
Remap less

Chocohead:
Now checks if an output exists and has a last modified date older than the input that otherwise will be remapped. Makes Loom a whole load faster as a result. Forces going digging for the installer JSON as a result of not always remapping Fabric-Loader, but fairly certain I've handled that as it works now.


asiekierka (approved):
LGTM, @modmuss50 ?

modmuss50 (approved):
Looks good here, I think we always had the idea of not doing it every run (Could have been for testing it wasnt)

PR #79, opened 2019-04-21
add correct Maven dependencies to maven-publish mod POMs

asiekierka:
For supporting transitive mod dependencies, work should instead focus on Loom 0.3.x. This just means that mods built from now until then have correct dependencies in the POMs - and yes, it's a hack; and yes, I haven't figured out how to support "maven" (non-publish) yet.

### DIFF
--- a/src/main/java/net/fabricmc/loom/AbstractPlugin.java
+++ b/src/main/java/net/fabricmc/loom/AbstractPlugin.java
@@ -146,40 +146,7 @@ public class AbstractPlugin implements Plugin<Project> {
 			}
 		}
 
-		project.afterEvaluate((p) -> {
-			// add modsCompile to maven
-			MavenPluginConvention maven = p.getConvention().findPlugin(MavenPluginConvention.class);
-			if (maven != null) {
-				// TODO: non-functional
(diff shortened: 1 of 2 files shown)


JamiesWhiteShirt:
It's hacky but it should work for now. I don't think the "maven" plugin is even worth supporting. "maven-publish" is supposed to replace it.

src/main/java/net/fabricmc/loom/AbstractPlugin.java
```diff
@@ -127,9 +144,42 @@ public void apply(Project target) {
...
+				maven.getConf2ScopeMappings().addMapping(MavenPlugin.COMPILE_PRIORITY, compileModsConfig, Conf2ScopeMappingContainer.COMPILE);
+			}
+
+			// add modsCompile to maven-publish
+			PublishingExtension mavenPublish = p.getExtensions().findByType(PublishingExtension.class);
+			if (mavenPublish != null) {
+				System.out.println("YASSS");
+				mavenPublish.publications((publications) -> {
+					for (Publication publication : publications) {
+						System.out.println("yass " + publication.getName());
```
modmuss50: not sure you need this ;)

modmuss50 (commented):
Yes, I agree there shouldn't be any need to support the `maven` plugin

PR #80, opened 2019-04-21
make modCompile() transitive

asiekierka:
All glory to @covers1624 , though he'll cry at how I wedged his very good Gradle-y code into this mess.

This also bumps the version to 0.2.2.

Should we have a modApi() and modImplementation(), now, and deprecate modCompile()?


Pyrofab:
I do think we should move to modApi() and modImplementation() now. It would bring mod dependencies closer to the gradle standard, and avoid confusion regarding transitivity.

asiekierka:
How should they work with regards to POM generation?

asiekierka:
https://github.com/FabricMC/fabric-loom/commit/b4beaed968c59901ea88359777ab7e8eb2ad53fb

PR #108, opened 2019-06-11
Make RemapJarTask an AbstractArchiveTask and add tests

magneticflux-:


build.gradle
```diff
@@ -4,6 +4,8 @@ plugins {
 	id 'java-gradle-plugin'
 	id 'idea'
 	id 'eclipse'
+	id 'groovy'
+	id 'org.jetbrains.kotlin.jvm' version '1.3.31'
```
liach: uh, no need to grab extra groovy and kotlin just for this change

src/main/java/net/fabricmc/loom/AbstractPlugin.java
```diff
@@ -58,344 +67,346 @@
...
+    protected Project project;
+
+    public static boolean isRootProject(Project project) {
+        return project.getRootProject() == project;
+    }
 
-	private void extendsFrom(String a, String b) {
-		project.getConfigurations().getByName(a).extendsFrom(project.getConfigurations().getByName(b));
-	}
+    private void extendsFrom(String a, String b) {
```
liach: retain tabs please!

liach (changes requested):

src/main/java/net/fabricmc/loom/AbstractPlugin.java
magneticflux-: Discussion on the Discord about this here: https://discordapp.com/channels/507304429255393322/566418023372816394/587800276791197716

build.gradle
magneticflux-: Groovy is required for the tests in Spock. Kotlin was just a personal test I forgot to remove

src/main/java/net/fabricmc/loom/AbstractPlugin.java
Chocohead: The inconsistency in Loom is straying from all tabs to rogue spaces, not the other way around

magneticflux-: Then what's the point of the `.editorconfig`? I thought it would be standardized there if it was a thing

liach: @asiekierka has indicated that fabric projects will stick to tabs, from java files, yarn buildscript, and even enigma and tiny mappings.

Chocohead: There are no formal formatting declarations anywhere besides a few ancient (in modern Fabric terms) ones which have never really used. On the whole things use tabs, but there are exceptions like Stitch.

Whilst it doesn't cover `.java` files, the [Loom `.editorconfig`](https://github.com/FabricMC/fabric-loom/blob/dev/0.2.4/.editorconfig) does only state tabs too for what it's worth.

JamiesWhiteShirt:
This is great, I can't wait for this to be merged for the integration opportunities.

magneticflux-:
Closed due to branch rename due to version increment due to binary and source incompatibility.
Superseded by #110

PR #110, opened 2019-06-16
Make RemapJarTask an AbstractArchiveTask and add tests

magneticflux-:
Supersedes #108 
A branch named `dev/0.2.5` will need to be created so that I can change the base of this PR before merging.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Sat Nov 03 23:12:45 CET 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
```
modmuss50, line 3: Do we want to build against a newer gradle version? Might be best to keep this as 4.10.2 for now unless there is a good reason to update. 5.5 has some interesting features that might be of use to use but thats not out yet.

src/test/groovy/net/fabricmc/loom/BuildUtils.groovy
```diff
@@ -0,0 +1,213 @@
+package net.fabricmc.loom
```
modmuss50, line 1: any reason doing this over than having them as normal files and reading them in?

modmuss50 (commented):
Had a quick look over it, I can see the tests being really useful. Few questions/queries about the PR,

1) Any reason for updating gradle, I dont see a reason to force updating people's setups
2) BuildUtils could possibly be moved out into files might be a bit easier to manage?
3) The whole of the project is currently in java is there a reason for the tests being in groovy? Im not against it just want to check we are ok having it.

Other than that I think asie needs to have a closer look as this mainly touches stuff written by him over the past month or so.

Nice work 👍

gradle/wrapper/gradle-wrapper.properties
magneticflux-: Lazy Configuration and the new `Jar` task type's `getArchiveFile()` method required Gradle >= 5.1. I just bumped it to the latest version and added tests for older versions so the build itself would get the benefits of being on a newer version while still checking for compatibility as far back as possible.

This wires the `Property` instances together lazily so the jar output can be redirected and it follows unless explicitly set: https://github.com/FabricMC/fabric-loom/pull/110/files#diff-f3b497e3f3aa2f298110c07b1467b2f6R315

src/test/groovy/net/fabricmc/loom/BuildUtils.groovy
magneticflux-: Having files in the source directory would cause them to get built, which isn't desirable. This way, their contents may be generated in a temporary directory and "fresh" builds can be run as if the plugin was being consumed normally.

They could be separated into different files and trimmed down, but I pretty much just copied the current example mod as a base.

magneticflux-:
The tests being in Groovy is recommended [here](https://docs.gradle.org/current/userguide/test_kit.html#example_using_gradlerunner_with_spock) by Gradle because the Spock framework is extremely powerful and expressive for testing. I don't recommend using Groovy for the plugin itself though, because it's typically slower than a fully compiled language like Java or Kotlin. (Kotlin would be nice to write plugins in though, with the work Gradle has done on cross-compatibility and ease of use)

gradle/wrapper/gradle-wrapper.properties
modmuss50: Ok, if upgrading is needed adding a check to ensure a compatible version is being ran should be fine, its only 1 command to update the wrapper. Might be worth holding out until 5.5 as it has possibly useful features and is currently in RC.

src/test/groovy/net/fabricmc/loom/BuildUtils.groovy
modmuss50: Humm ok, I guess they wont change a lot and they dont seem to hard to maintain so thats fine.

modmuss50:
Ok, using groovy should be fine for the tests, im not familiar with that testing framework but I see it possibly being a bit of a PITA with java.

liach:
consider putting the test code files in test resources.

magneticflux-:
@liach I decided to make them multiline `GString`s in order to allow string interpolation in the files if a future test required that. It's only used to change versions right now, but it may be used in the future to inject more specific tests.

liach (approved):
Appears ready to me.

magneticflux-:
@modmuss50 @asiekierka Ping me when there's a `dev/0.2.5` branch and I'll change this PR's base.

NikkyAI (approved):
💯

modmuss50:
I merged this PR into https://github.com/FabricMC/fabric-loom/commits/dev/0.2.5

Not sure why its not showing as merged (prob becuase I did it in my IDE), sorry.

Thanks for the help 👍

PR #120, opened 2019-07-16
Fix crash in multi-project build

comp500:
When using Loom in a multi-project build, `getByName("compile")` fails on the root project (as it does not exist), throwing an UnknownConfigurationException, which breaks the remapJar task. This should ensure that it falls back correctly to buildscript configurations without crashing - although I haven't found a situation where getMixinDependency returns an actual Dependency, I'm not sure exactly what it does.

I tried this with 0.2.4 and 0.2.5 and it still failed, so it hasn't been fixed in newer versions. I'm submitting the PR to the 0.2.3 branch because that's what the example mod (and therefore my mod) uses.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -37,6 +37,7 @@ import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
 import org.gradle.api.artifacts.result.ResolvedArtifactResult;
+import org.gradle.api.artifacts.UnknownConfigurationException;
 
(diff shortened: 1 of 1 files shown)


MrIbby:
Yes! I was just dealing with this problem too!

modmuss50:
Is this happening when the root project does not have loom?

Are you usre just try-catching it is the correct soltuion?

comp500:
Yes. The structure I'm testing it with has a blank root project, and two subprojects with fabric and forge. This setup works when I add the try-catch, by publishing my fix to the local maven.

I don't know if it is exactly the correct solution as I don't know what getMixinDependency is used for, but the comment says it should fall back to the buildscript, so I don't think this will cause issues.

asiekierka:
That's not the right fix...

modmuss50:
Closing this, as no progress has been made to find the correct fix. This is just wrong sadly.

Thanks for looking into it.

Issue #118, opened 2019-07-06
RemapSourcesJarTask does not copy resources


JamiesWhiteShirt:
Pretty minor, but remapSourcesJar/RemapSourcesJarTask appears to only affect .java files. Perhaps it should just copy the remaining files?

modmuss50:
Fixed

PR #129, opened 2019-08-15
copy non-java files into remapped sources jar (fixes #126 and #118)

### LINKED ISSUE #126: Non-Java sources are excluded from source jar after remapping
The `remapSourcesJar` task only copies files ending in `*.java`. For non-Java mods, this means the source jar is incomplete.

Even if the sources can't be remapped, it may be useful to include them as-is.

natanfudge:
No-java sources files are copied into the remapped source jar as-is, both when building the source jar and when consuming it. This might result in seeing source code with different mappings, but that's minor (have you seen decompiled Kotlin code before?).


modmuss50 (approved):
This seems reasonble.

Player3324 (approved):

PR #133, opened 2019-09-26
Make generated Eclipse run configs more reliably pick the right working directory

Daomephsta:
The working directory for the generated Eclipse run configs is currently `${project_loc}/run`.
This causes problems because if no argument is passed, `${project_loc}` refers to the project containing the resource selected when the config is run. This means that the generated configs will not always launch the expected project, and will error on run if nothing is selected. This PR adds the project name as an argument, so that the config always launches the correct project and does not error if nothing is selected.


grondag:
Haven't tested this, but as an eclipse user, it really can''t make it any worse. :-)

grondag (approved):

Daomephsta:
I tested it before I made the PR. It worked as expected.

modmuss50 (approved):
Thanks for this, I know there are a few issues with the eclipse run configs, this is becuase I was given this template file and went with it :)

Ill merge it in as anything is better than what we have atm.

PR #135, opened 2019-10-06
Fix issues involving projects with modules that contain no java classes

zml2008:
This PR is an attempt to resolve build issues that came up while putting together a Fabric version of PEX. I'm not super familiar with the internals of Loom, so I'm very open to changing anything here as long as I can get out of here with a working PEX build.

- Loom expects there to be a compileClasspath configuration in every project Gradle is processing. Because this configuration is provided by the java plugin, it is only present on JVM builds -- this lets non-jvm builds be present in the project.

- Some entries in the compile classpath are empty directories -- for example the classes/java folder in a pure-kotlin project. TinyRemapper doesn't like being given empty/nonexistent directories in its classpath -- ignoring invalid entries more closely matches the behaviour of `javac` and gradle's builtin tasks, and gives a more forgiving build process

- Additionally, adds the output of remapJar as an artifact before the task executes to resolve a conflict with the signing plugin -- the signing plugin adjusts the dependencies of its task whenever an artifact is added, and tasks cannot have their dependencies changed while other tasks are executing.

### DIFF
--- a/src/main/java/net/fabricmc/loom/AbstractPlugin.java
+++ b/src/main/java/net/fabricmc/loom/AbstractPlugin.java
@@ -339,8 +339,8 @@ public class AbstractPlugin implements Plugin<Project> {
 				extension.addUnmappedMod(jarTask.getArchivePath().toPath());
 				remapJarTask.getAddNestedDependencies().set(true);
 
-				remapJarTask.doLast(task -> project1.getArtifacts().add("archives", remapJarTask.getArchivePath()));
-				remapJarTask.dependsOn(project1.getTasks().getByName("jar"));
+				project1.getArtifacts().add("archives", remapJarTask);
+				remapJarTask.dependsOn(jarTask);
 				project1.getTasks().getByName("build").dependsOn(remapJarTask);
 
 				Map<Project, Set<Task>> taskMap = project.getAllTasks(true);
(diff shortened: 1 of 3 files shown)


JamiesWhiteShirt (approved):

src/main/java/net/fabricmc/loom/AbstractPlugin.java
```diff
@@ -339,7 +339,7 @@ protected void configureCompile() {
 				extension.addUnmappedMod(jarTask.getArchivePath().toPath());
 				remapJarTask.getAddNestedDependencies().set(true);
 
-				remapJarTask.doLast(task -> project1.getArtifacts().add("archives", remapJarTask.getArchivePath()));
+				project1.getArtifacts().add("archives", remapJarTask.getArchivePath());
```
JamiesWhiteShirt:
```suggestion
				project1.getArtifacts().add("archives", remapJarTask);
```
Unrelated to the scope of this PR, but we might as well do it while we're changing this. [See what ArtifactHandler does when adding AbstractArchiveTasks](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/dsl/ArtifactHandler.html).

JamiesWhiteShirt (approved):

src/main/java/net/fabricmc/loom/AbstractPlugin.java
zml2008: Sure, I've incorporated this change -- also modified the line below to not fetch the jar task a second time while I was looking at the area.

modmuss50 (approved):

modmuss50:
Merged into 0.2.6, im open to back porting if wanted.

PR #138, opened 2019-11-06
Use Dev Launch Injector to ensure that the launch options are always kept upto date.

modmuss50:
Will remove the need to regen the run configs each time the game is updated.

This isnt fully complete yet as I want to add a way that force regens old run configs to this to reduce possible issues when updating.


src/main/java/net/fabricmc/loom/task/GenVsCodeProjectTask.java
```diff
@@ -90,7 +90,7 @@ public void add(RunConfig runConfig) {
 	private static class VsCodeConfiguration {
 		public String type = "java";
 		public String name;
-		public String request = "launch";
+		public String request = "net/fabricmc/launch";
```
modmuss50: I blame idea's refactoring for this.

src/main/java/net/fabricmc/launch/Launch.java
```diff
@@ -0,0 +1,36 @@
...
+
+package net.fabricmc.launch;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public final class Launch {
+	public static void main(String[] args) {
+		List<String> extraArgs = new ArrayList<>(Arrays.asList(args));
```
natanfudge: `Lists.newArrayList`

natanfudge: Lists.newArrayList if we have it

src/main/java/net/fabricmc/loom/providers/LaunchProvider.java
```diff
@@ -0,0 +1,275 @@
...
+					.argument("--assetIndex")
+					.argument(extension.getMinecraftProvider().versionInfo.assetIndex.getFabricId(extension.getMinecraftProvider().minecraftVersion))
+					.argument("--assetsDir")
+					.argument(new File(extension.getUserCache(), "assets").getAbsolutePath());
+
+			serverLaunchDetails
+					.property("fabric.development", "true");
+
+			final List<ZipEntrySource> entries = new ArrayList<>();
+			entries.add(generateZipEntry("net/fabricmc/launch/LaunchClient", clientLaunchDetails));
```
natanfudge: `new ZipEntrySource[]{generateZipEntry...., generateZipEntry...}` instead of this list ordeal

src/main/java/net/fabricmc/loom/providers/LaunchProvider.java
```diff
@@ -0,0 +1,275 @@
...
+
+		patchMethod(method, launchDetails);
+
+		return writeClassToBytes(classNode);
+	}
+
+	private static void patchMethod(MethodNode method, LaunchDetails launchDetails) {
+		final InsnList insnList = new InsnList();
+
+		for (Map.Entry<String, String> entry : launchDetails.getSystemProperties().entrySet()) {
```
natanfudge: `launchDetails.getSystemProperties().foreach((k,v) ->setProperty(insnList, k, v))`

src/main/java/net/fabricmc/loom/providers/LaunchProvider.java
```diff
@@ -0,0 +1,275 @@
...
+
+	@Override
+	public String getTargetConfig() {
+		return Constants.MINECRAFT_NAMED;
+	}
+
+	private static String getMainClass(String side, LoomGradleExtension extension) {
+		JsonObject installerJson = extension.getInstallerJson();
+
+		if (installerJson != null) {
```
natanfudge: `if (installerJson != null && installerJson.has("mainClass"))`

natanfudge (changes requested):

src/main/java/net/fabricmc/launch/Launch.java
modmuss50: I think we do have it, but dont want to relay on a 3rd party dep just to make the code slighty nicer in an auto generated class that no one should really have the need to touch.

natanfudge (approved):

modmuss50:
Updated to now use [dev-launch-wrapper](https://github.com/FabricMC/dev-launch-wrapper)

Config file looks like this:

```
commonProperties
	fabric.development=true
clientArgs
	--assetIndex
	19w45b-1.15
	--assetsDir
	C:\Users\<user>\.gradle\caches\fabric-loom\assets
clientProperties
	java.library.path=C:\Users\<user>\.gradle\caches\fabric-loom\natives\19w45b
	org.lwjgl.librarypath=C:\Users\<user>\.gradle\caches\fabric-loom\natives\19w45b
```

JamiesWhiteShirt (commented):
I am a bit unsure about the name "devlaunchwrapper". Could we perhaps name it something like "Loom Launcher" to avoid any confusion with LaunchWrapper?

JamiesWhiteShirt (commented):
What if we generated a launcher config per launch (server/client) instead of combining it in one file? It would make the launcher a lot simpler.

Player3324 (approved):

Player3324:
Launchwrapper is pretty much dead, I don't think that's a problem. A single config file is nicer to handle imo, the implementation is really simple already.

asiekierka:
I would still consider renaming it, as while it's dead it's still a very recognized name.

modmuss50:
Im still not sure what better name should be used

natanfudge:
[Warp](https://en.m.wikipedia.org/wiki/Warp_and_weft)

modmuss50:
Should be all good now.

PR #140, opened 2019-11-07
Improve performance of the SourceRemapper when working on jars.

modmuss50:
Untested, waiting on https://github.com/CadixDev/Mercury/pull/3


modmuss50:
Doesnt work, and cannot fix.

https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java#L1013

PR #146, opened 2019-11-11
Fix V2 regressions

### LINKED ISSUE #21: Parameters with no names are emitted, which breaks the kotlin compiler
Relevant youtrack issue https://youtrack.jetbrains.com/issue/KT-34834

### LINKED ISSUE #143: V2 regressions
- Gradle often pulls version 0.1 of tiny-mappings-parser instead of 0.2 for loader
- If you just update loom 0.2.6 and do nothing else, loom will remap the minecraft jar with generated field names, but won't update the remapped dependencies like Fabric API. The mappings then turn out different which causes a FieldNotFoundError.

natanfudge:
Upgrading stitch fixes part 1 of #143 , not excluding dependencies from mixin compile extensions should fix part 2 of #143 . 
Upgrading tiny remapper and setting `renameInvalidLocals` to true should fix FabricMC/tiny-remapper#21.
Fixes #143.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ dependencies {
 	implementation ('com.google.guava:guava:28.0-jre')
 
 	// game handling utils
-	implementation ('net.fabricmc:stitch:0.3.0.66') {
+	implementation ('net.fabricmc:stitch:0.4.0.67') {
 		exclude module: 'enigma'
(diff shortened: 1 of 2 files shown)


modmuss50 (approved):
Looks good, thanks for this.

PR #154, opened 2019-11-29
Javadocs

modmuss50:
Requires https://github.com/FabricMC/intellij-fernflower/pull/4


natanfudge (approved):
hmm, not so hard was it

liach:
This doesn't handle param docs yet, but we can take care of that later

Player3324 (approved):

PR #156, opened 2019-11-30
[DRAFT] Adding transformer support to be able to use remapped dependencies in any configuration.

marchermans:
### Current situation
In the current situation a custom configuration is used to adapt and remap the mod dependencies to the mapping as defined by the user.
Even though this works _as intended_ it is not entirely compatible with Gradle spec. 

### New situation
This implements transformers that interact with all dependencies added to the project and will transform them accordingly.
The transformers call the old methods, which where previously looping over a configuration.
Each transformer will initially detect if the artifact passed in a valid Fabric Mod before performing either a remap of a library jar, or a sources jar.


liach:
must you get the project in such nasty ways?

marchermans:
@liach Sadly yes, gradle uses a custom injector system to inject data into the transformers itself, as such what can be injected is extremely limited. (Basically only its inputs). Additionally Transformers need to perform a pure action and their parameters need to be isolateble. In this context that means that the entire parameter set needs to be serializable.

In the current context this means that to get the external data (the data needed to remap, for example the MC dependency or the mapping data) from the project config there is literally only one way.

That said I am currently looking into a technology stack which would pass all of that information through to us without having to use the project at all. But yeah as of right now this PR also needs an update of TinyRemapper since that can not handle being uses in a multithreaded environment. And just crashes. Hence it being a draft as well.

marchermans:
For everybody reading this: This PR needs a bug fix in TinyRemapper to be usefull: https://github.com/FabricMC/tiny-remapper/issues/7 is the hanging issue.

src/main/java/net/fabricmc/loom/transformers/CompiledJarRemappingTransformer.java
```diff
@@ -0,0 +1,144 @@
...
+				FileUtils.copyFile(inputFile, outputFile);
+			} catch (IOException e) {
+				lifecycle("Failed to copy not remappable file to output.", e);
+			}
+            return;
+		}
+
+		//Rudimentary check for sources. Which need to be remapped separately.
+        //Potentially we need to check zip file contents.
+		if (inputFile.getName().contains("-sources"))
```
natanfudge: Considering this one and the sources jar one are extremely similiar,  I think some common code can be extracted.

src/main/java/net/fabricmc/loom/transformers/TransformerProjectManager.java
```diff
@@ -0,0 +1,51 @@
...
+package net.fabricmc.loom.transformers;
+
+import com.google.common.collect.Maps;
+import org.gradle.api.Project;
+
+import java.util.Map;
+
+public class TransformerProjectManager
+{
+    private static final TransformerProjectManager INSTANCE = new TransformerProjectManager();
```
natanfudge, line 34: add a private constructor if you plan on keeping this

src/main/java/net/fabricmc/loom/transformers/TransformerProjectManager.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public static TransformerProjectManager getInstance() {
+        return INSTANCE;
+    }
+
+    private Map<String, Project> projects = Maps.newConcurrentMap();
+
+    public void register(Project project)
+    {
```
natanfudge, line 43: wrong formatting

src/main/java/net/fabricmc/loom/util/FabricModUtils.java
```diff
@@ -0,0 +1,56 @@
...
+import org.gradle.api.logging.Logger;
+
+import java.io.File;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public final class FabricModUtils
+{
+
+    private FabricModUtils()
+    {
```
natanfudge, line 38: formatting

src/main/java/net/fabricmc/loom/util/FabricModUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        throw new IllegalStateException("Tried to initialize: FabricModUtils but this is a Utility class.");
+    }
+
+    /**
+     * Checks if an artifact is a fabric mod, according to the presence of a fabric.mod.json.
+     */
+    public static boolean isFabricMod(Project project, Logger logger, File artifact, String notation) {
+        AtomicBoolean fabricMod = new AtomicBoolean(false);
+        project.zipTree(artifact).visit(f -> {
```
natanfudge, line 47: Naming this lambda param `file` would be beneficial in my opinion

src/main/java/net/fabricmc/loom/util/InstallerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+
+                try (InputStream inputstream = jarFile.getInputStream(entry)) {
+                    jsonStr = IOUtils.toString(inputstream, StandardCharsets.UTF_8);
+                }
+            }
+
+            JsonObject jsonObject = GSON.fromJson(jsonStr, JsonObject.class);
+            return jsonObject;
+        } catch (IOException e) {
+            e.printStackTrace();
```
natanfudge, line 89: Might want something better than this

src/main/java/net/fabricmc/loom/util/InstallerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+{
+
+    private static final Gson GSON = new Gson();
+
+    private InstallerUtils()
+    {
+        throw new IllegalStateException("Tried to initialize: InstallerUtils but this is a Utility class.");
+    }
+
+    static JsonObject readInstallerJson(File file, Project project) {
```
natanfudge, line 53: Is this intentionally package private?

src/main/java/net/fabricmc/loom/util/InstallerUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        try {
+            LoomGradleExtension extension = project.getExtensions().getByType(LoomGradleExtension.class);
+            String launchMethod = extension.getLoaderLaunchMethod();
+
+            String jsonStr;
+            int priority = 0;
+
+            try (JarFile jarFile = new JarFile(file)) {
+                ZipEntry entry = null;
```
natanfudge, line 62: I don't like how much mutable state is in this code, but I don't see an easy way around it so don't worry about it too much.
Also formatting.

src/main/java/net/fabricmc/loom/util/ModProcessor.java
```diff
@@ -58,74 +61,68 @@
...
+                return;
+            }
 
-		if (entry == null) {
-			throw new RuntimeException(String.format("%s was not found in %s", fileName, parentJar.getName()));
-		}
+            JsonArray jsonArray = json.getAsJsonArray("jars");
 
-		File nestedFile = new File(extension.getNestedModCache(), fileName.substring(fileName.lastIndexOf("/")));
+            for (int i = 0; i < jsonArray.size(); i++) {
```
natanfudge, line 97: can't you use a `for : ` (foreach) loop here?

src/main/java/net/fabricmc/loom/util/SourceUtils.java
```diff
@@ -0,0 +1,79 @@
...
+
+public final class SourceUtils {
+
+    private SourceUtils() {
+        throw new IllegalStateException("Tried to initialize: SourceUtils but this is a Utility class.");
+    }
+
+    public static File findSources(DependencyHandler dependencies, ResolvedArtifact artifact) {
+        @SuppressWarnings("unchecked") ArtifactResolutionQuery query = dependencies.createArtifactResolutionQuery()//
+                                                                                       .forComponents(artifact.getId().getComponentIdentifier())//
```
natanfudge: IDK why but the formatting here looks awful on github

natanfudge (changes requested):
remember to rebase

src/main/java/net/fabricmc/loom/util/FabricModUtils.java
liach: Stop using project here. Use java 8 zip file sysstem api or zip entry. also for project logger use sys out println instead

src/main/java/net/fabricmc/loom/util/GradleSupport.java
```diff
@@ -34,24 +34,6 @@
...
-	private static RegularFileProperty getfilePropertyModern(Project project) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
-		ObjectFactory objectFactory = project.getObjects();
-		Method method = objectFactory.getClass().getDeclaredMethod("fileProperty");
-		method.setAccessible(true);
-		return (RegularFileProperty) method.invoke(objectFactory);
-	}
-
-	private static RegularFileProperty getfilePropertyLegacy(Project project) {
-		return project.getLayout().fileProperty();
+		return project.getObjects().fileProperty();
```
Chocohead: This class could be removed completely if there's no backwards compatibility being gained from it anymore

src/main/java/net/fabricmc/loom/util/ModProcessor.java
Chocohead: It would iterate over `JsonElement`s rather than `JsonObject`s so it's hardly much better

src/main/java/net/fabricmc/loom/util/InstallerUtils.java
Chocohead: It's package private in `ModProcessor` so is probably just a holdover from copy-pasting

src/main/java/net/fabricmc/loom/util/ModProcessor.java
natanfudge: It would have the same benefits as in any other for loop, 
``` JsonObject jsonObject = jsonElement.getAsJsonObject();```

File: src/main/java/net/fabricmc/loom/util/GradleSupport.java (hunk 11 above)
Comment (liach) [reply src/main/java/net/fabricmc/loom/util/GradleSupport.java]: Yep, should remove this completely.

Comment (marchermans) [reply src/main/java/net/fabricmc/loom/util/GradleSupport.java]: Okey will do.

File: src/main/java/net/fabricmc/loom/util/ModProcessor.java (hunk 9 above)
Comment (marchermans) [reply src/main/java/net/fabricmc/loom/util/ModProcessor.java:97]: TBH This is original code, i did not change this. But I plan to go over all of this and clean this kind of stuff up in an upcoming PR later on. This will probably stay in this PR, getting this to behave is already kind of difficult.

File: src/main/java/net/fabricmc/loom/util/InstallerUtils.java (hunk 8 above)
Comment (marchermans) [reply src/main/java/net/fabricmc/loom/util/InstallerUtils.java:62]: The way we do this is really nasty. TBH The installer dependencies should not be retroactively injected in the classpath based on some weird spec nobody knows.

I will see what I can do to propose a Change here in the future, then this will probably Change as well.

File: src/main/java/net/fabricmc/loom/util/InstallerUtils.java (hunk 7 above)
Comment (marchermans) [reply src/main/java/net/fabricmc/loom/util/InstallerUtils.java:53]: Yep.
InstallerUtils comes from the ModProcessor class, so yeah I will need to give These Kinds of Things a brush over.

File: src/main/java/net/fabricmc/loom/util/InstallerUtils.java (hunk 6 above)
Comment (marchermans) [reply src/main/java/net/fabricmc/loom/util/InstallerUtils.java:89]: Again original code does this. Will fix it.

File: src/main/java/net/fabricmc/loom/util/FabricModUtils.java (hunk 5 above)
Comment (OrionDevelopment) [reply src/main/java/net/fabricmc/loom/util/FabricModUtils.java]: I did not write this, however I totally agree, we should change this.

File: src/main/java/net/fabricmc/loom/transformers/CompiledJarRemappingTransformer.java (hunk 1 above)
Comment (marchermans) [reply src/main/java/net/fabricmc/loom/transformers/CompiledJarRemappingTransformer.java]: Yep totally agree.
Some lambda logic and a builder would probably go a long way.
But I would love to write this up a bit better so that it works in nearly all cases perfectly before I go mayham on cleaning loom up and making it more stable.

File: src/test/groovy/net/fabricmc/loom/BuildUtils.groovy [hunk 12]
```diff
@@ -24,10 +24,10 @@ dependencies {
 	//to change the versions see the gradle.properties file
 	minecraft "com.mojang:minecraft:\${project.minecraft_version}"
 	mappings "net.fabricmc:yarn:\${project.yarn_mappings}"
-	modCompile "net.fabricmc:fabric-loader:\${project.loader_version}"
+	compile "net.fabricmc:fabric-loader:\${project.loader_version}"
```
natanfudge, line 27: :O

marchermans: What?
That is the goal of this whole ordeal :D

liach: Well, this should have been `compile` in the example mod, given how the loader touches no minecraft code at all

modmuss50:
is this ready for reviews and testing?

marchermans:
I am still working on it.
That damn IDEA task is kind of difficult to deal with. Also christmas, and 3 birthdays in one week kind of took its toll. 
Getting back on my horse though.

modmuss50:
No, dont rush or worry at all. I was just checking you werent waiting on me or something stupid 👍

natanfudge:
I wouldn't worry too much about the IDEA task. You can still just import Gradle instead.

modmuss50:
Going to close this PR as its very outdated, and unlikely something we will be doing. It was worth experimenting with back then. Thanks.

PR #158, opened 2019-12-06
Change mixin refmap default name to use archivesBaseName instead of the project name

Juuxel:
Using the project name for the default refmap name can cause problems when libraries are pulled in with Jitpack. When building, Jitpack clones projects in a directory called `build`, which sets the project name to `build` if not manually defined. The resulting refmap, `build-refmap.json`, can conflict with other mods' refmaps, leading to mixin crashes.

Unlike the project name or the refmap name, the `archivesBaseName` of a project is usually set manually [like in the example mod](https://github.com/FabricMC/fabric-example-mod/blob/705d4da55b9a514e5e7279a2ea5becaf039cf2d0/gradle.properties#L13), so conflicts should be more rare.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -48,6 +48,7 @@ import net.fabricmc.loom.providers.MappingsProvider;
 import net.fabricmc.loom.providers.MinecraftMappedProvider;
 import net.fabricmc.loom.providers.MinecraftProvider;
 import net.fabricmc.loom.util.LoomDependencyManager;
+import org.gradle.api.plugins.BasePluginConvention;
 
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):
Seems a resonable change

liach (approved):

liach:
Should fix FabricMC/fabric-loader#140

asiekierka:
Ideally, we'd fix refmap loading to actually keep in mind the JAR file being used, but we don't live in a perfect world.

asiekierka (approved):

Player3324 (approved):

LemmaEOF:
Can this be merged in soon?

ghost (approved):

coderbot16 (approved):
time to change `archivesBaseName` to `build` :thinking:

PR #160, opened 2019-12-12
Allow more granularly setting nested dependencies

### LINKED ISSUE #159: Allow more granularly setting nested dependencies
Right now I have a setup where I publish 2 variants of jars:
- One without dependencies (modpack)
- One with dependencies (standalone)

Now I would like to extract a module out of my mod to be used in another one. There's not much point distributing the module because it will likely only be used by me. So what I would like to do is include it in _both_ variants. However, if I set `addNestedDependencies` to `true`, all dependencies will be added, and if I set it to false, none of them will be added. I need a way to include _only_ the one shared module. 
I think the best way of going about these is having a `forceAddNestedDependencies` (or a better name if you can think of one) on the `RemapJarTask` that is a list of `archivesBaseNames` that will be included even when `addNestedDependencies` are true.

natanfudge:
Closes #159

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
+++ b/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
@@ -29,6 +29,7 @@ import java.io.FileNotFoundException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
@@ -126,7 +127,7 @@ public class RemapJarTask extends Jar {
 			project.getLogger().debug("Transformed mixin reference maps in output JAR!");
 		}
 
-		if (NestedJars.addNestedJars(project, output, getForcedNestedDependencies().getOrElse(new ArrayList<>()),
+		if (NestedJars.addNestedJars(project, output, getForcedNestedDependencies().getOrElse(Collections.emptyList()),
 				getAddNestedDependencies().getOrElse(false))) {
 			project.getLogger().debug("Added nested jar paths to mod json");
 		}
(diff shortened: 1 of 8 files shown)


natanfudge:
On second thought, wouldn't it make more sense to have `include` as a concept be tied to the `Jar` task, instead of something you apply to everything? This way you just choose exactly what dependencies you want to include for each `Jar` task.

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -122,10 +126,9 @@ public void doTask() throws Throwable {
 			project.getLogger().debug("Transformed mixin reference maps in output JAR!");
 		}
 
-		if (getAddNestedDependencies().getOrElse(false)) {
-			if (NestedJars.addNestedJars(project, output)) {
-				project.getLogger().debug("Added nested jar paths to mod json");
-			}
+		if (NestedJars.addNestedJars(project, output, getForcedNestedDependencies().getOrElse(new ArrayList<>()),
```
liach: use collections emptylist call

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -29,6 +29,7 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
```
liach: Remove this import

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -28,11 +28,14 @@
 import java.io.FileNotFoundException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
```
natanfudge:
```suggestion
```

liach (approved):

natanfudge:
Probably not relevant if you use a modpack flattener

PR #167, opened 2019-12-27
Brute force solution to the "no obfuscation mapping" problem

natanfudge:


src/main/java/net/fabricmc/loom/util/SetupIntelijConfigs.java
```diff
@@ -86,5 +80,21 @@ private static void generate(Project project) throws IOException {
...
+			MinecraftAssetsProvider.provide(extension.getMinecraftProvider(), project);
+			MinecraftNativesProvider.provide(extension.getMinecraftProvider(), project);
+
+			// IDEA refuses to build when you have any mixins when annotation processing is enabled
+			turnOffAnnotationProcessing(projectDir);
+		}
+	}
+
+	private static void turnOffAnnotationProcessing(File projectDir) throws IOException {
+		File compilerConfigs = new File(projectDir, "compiler.xml");
```
modmuss50: Make sure the file exists

natanfudge: ok

modmuss50:
I think I sloved this issue? Not heard about it in a long while

PR #168, opened 2020-01-01
Add mixins that target the class as a comment.

modmuss50:
2am code, so prob got some issues. I plan on expanding this to add the comment to methods as well.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/MixinTargetScanner.java
+++ b/src/main/java/net/fabricmc/loom/util/MixinTargetScanner.java
@@ -35,6 +35,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipException;
 import java.util.zip.ZipFile;
 
 import com.google.common.reflect.TypeToken;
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/util/MixinTargetScanner.java
```diff
@@ -0,0 +1,216 @@
...
+
+	public void scan() {
+		classMixins = new HashMap<>();
+
+		Set<File> filesToScan = configuration.getResolvedConfiguration().getFiles();
+		filesToScan.forEach(this::scanFile);
+	}
+
+	private void scanFile(File file) {
+		//is this enough?
```
coderbot16: Why not just check if it is a valid zip file?

modmuss50:
I will most likely get this finished off without methods for now. Methods will take a bit more work

modmuss50:
should be good now.

asiekierka:
This would've been better implemented as an IDEA plugin or something, IMO. Decompilation of the Minecraft JAR was meant to be done once per version/mapping, not once per dependency change as this change seems to imply.

modmuss50:
I would 1/2 agree with that, however it does help slove an issue people have been voicing for a while, without too much work.

As for it changing per dep, I do now see an issue with this (as JamiesWhiteShirt also pointed out), as if you have more than one project things will be fighting across them.

modmuss50:
I think for now im going to revert it

PR #169, opened 2020-01-01
Add mixins that target the class as a comment.

modmuss50:
Same as #168 

A few issues with this:

- Causes issues with more than one project, as diffrent projects will have diffrent deps but use the same source jar.


modmuss50:
Going to close this PR as its badly out of date, and I think an IDE plugin may be able to provide a better solution to this issue.

PR #170, opened 2020-01-08
Speed up genSources by 3000%*

natanfudge:
*provided `genSources` was run before on a sufficiently similar Minecraft jar.

## How it works
- On the first time `genSources` runs, on version A, decompile the entire jar normally. Save the compiled jar that has the original Mojang linemaps (the _original Yarn jar_) - #171.
- On subsequent runs, on version B, compare the _original yarn jar_ of version B to that of version A. Produce a list of class files in version B that are the same as those in version A (_unchanged classes_), and a list of class files that have either changed or do not have an equivalent in version A (_changed classes_) - [here](https://github.com/natanfudge/fabric-loom/blob/61652bcd4682d13bff7a17fcf6769753c1a29baf/src/main/java/net/fabricmc/loom/task/fernflower/IncrementalDecompilation.java#L133). 
- Decompile only _changed classes_ - [here](https://github.com/natanfudge/fabric-loom/blob/61652bcd4682d13bff7a17fcf6769753c1a29baf/src/main/java/net/fabricmc/loom/task/fernflower/IncrementalDecompilation.java#L89).
- Add all _unchanged classes_ from the source jar of version A into the result of the decompilation (of version B) - [here](https://github.com/natanfudge/fabric-loom/blob/61652bcd4682d13bff7a17fcf6769753c1a29baf/src/main/java/net/fabricmc/loom/task/fernflower/FernFlowerTask.java#L164).
- Linemap the compiled Yarn jar of version B is per usual.
- Copy over all _unchanged classes_ from the linemapped jar of version A to the linemapped jar of the result linemapped jar (of version B) - [here](https://github.com/natanfudge/fabric-loom/blob/61652bcd4682d13bff7a17fcf6769753c1a29baf/src/main/java/net/fabricmc/loom/task/ApplyLinemappedJarTask.java#L87) .

I dub this process **Incremental Decompilation**.

## Benefit
99.9% of time is spent inside FernFlower decompiling classfiles. This means that decompiling 50% less classfiles means the process will be twice as fast. 
From yarn version 1.15.1+build.14 to 1.15.1+build.23 **only ~140 out of ~5000 classes need to be redecompiled. This is a speedup of 3500%.**  From version 1.15.1+build.23 to 1.15+build.1 only ~800 out of ~5000 need to be redcompiled. A modest speedup of 650%. 

## Possible issues
It's possible that quality of decompilation will go down as a result. However, I haven't observed any significant changes in output.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ targetCompatibility = 1.8
 
 group = 'net.fabricmc'
 archivesBaseName = project.name
-version = '0.2.9-SNAPSHOT'
+version = '0.2.7-SNAPSHOT'
 
 def build = "local"
 def ENV = System.getenv()
--- a/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
@@ -30,25 +30,13 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Locale;
 
+import net.fabricmc.loom.task.*;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.tasks.TaskContainer;
 
 import net.fabricmc.loom.providers.MappingsProvider;
 import net.fabricmc.loom.providers.MinecraftLibraryProvider;
-import net.fabricmc.loom.task.AbstractDecompileTask;
-import net.fabricmc.loom.task.CleanLoomBinaries;
-import net.fabricmc.loom.task.CleanLoomMappings;
-import net.fabricmc.loom.task.DownloadAssetsTask;
-import net.fabricmc.loom.task.GenEclipseRunsTask;
-import net.fabricmc.loom.task.GenIdeaProjectTask;
-import net.fabricmc.loom.task.GenVsCodeProjectTask;
-import net.fabricmc.loom.task.MigrateMappingsTask;
-import net.fabricmc.loom.task.RemapJarTask;
-import net.fabricmc.loom.task.RemapLineNumbersTask;
-import net.fabricmc.loom.task.RemapSourcesJarTask;
-import net.fabricmc.loom.task.RunClientTask;
-import net.fabricmc.loom.task.RunServerTask;
 import net.fabricmc.loom.task.fernflower.FernFlowerTask;
 
 public class LoomGradlePlugin extends AbstractPlugin {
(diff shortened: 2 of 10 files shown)


modmuss50:
Not got the time to read the code yet, seems like a good idea. I would add a cleanSources task that cleans it all up so the next genSources run is a full run.

JamiesWhiteShirt:
~~The idea of incremental decompilation might be good, but I would not be surprised if it is more complex than this due to potential dependencies between classes during decompilation. I don't trust this to work correctly until I see diffs (or a lack of diffs) of pure decompilation and incremental decompilation between two significantly different jars like a 1.14 snapshot and 1.15.1.~~

I realize I may have misunderstood the scope of this. I was initially under the impression that this was across game versions, but the example is actually across mapping versions on the same version of the game. Thus, there are differences only in names, and I find it much easier to trust that this works at least on the same version of the game.

modmuss50:
Im still not convinced this is a real solution. I am worried for things to get messed up, I think the real fix here is making the decompiler faster/replacing it with a faster decompiler.

natanfudge:
No decompiler speedups will achieve the level of improvement that this PR gives. The real solution is to test this solution to see if it works (in addition to my findings that it works).

natanfudge:
Lets compare the decompilation output of Minecraft 1.15.1 with build.1 of Yarn, decompiled normally, and decompiled incrementally when using 1.14.4-build.1 as comparison. 
When using incremental we get this message:
image
Therefore I would expect the difference to be extremely minimal. 
I used [WinMerge](https://winmerge.org/) to compare the outputs. Here's an overall view of the diff:
overall_diff
I looked at differences of 2 different files in `com/mojang`, and 4 different files in `net/minecraft`. 
#### FileDownload
file_download
The change here involves moving the same logic around. It looks like the output on the right (the incremental one) is actually better, not producing a label. 
#### RealmsDownloadLatestWorldScreen
realmsdownloadlatestworldscreen
Fernflower opts to using the `if-return` strategy instead of putting the code in one big if. A small stylistic change.  
#### CarvedPumpkingBlock
carvedpumpkinblock
The arguments passed as varargs are represented as an explicit String array allocation, perhaps slightly worse, but not significant. 
#### AdvancementScreen
advancementsscreen
FernFlower does not explicitly qualify `Listener`. 
#### EndPortalFrameBlock
endportalframeblock
Explicit varargs array.
#### TrueTypeFont
truetypefont
The same logic coded slightly differently.

And finally, here's an overview of all changed files:
fulldiff
As expected, the changes are minimal. Only around 20 files are different out of over 5000, and those that are different aren't significantly so.

modmuss50:
This has been mostly superseeded by fully threaded fernflower. Thanks for the support, but I dont think this was ever the real solution.

PR #171, opened 2020-01-08
Save yarn jars with original line mappings for incremental decompilation

natanfudge:
Part of #170, I've split up this part so we can merge this earlier, and then sources generated since this PR was merged will be able to be used once #170 is merged.


modmuss50:
No longer needed, thanks for the help

PR #172, opened 2020-01-12
Fix being unable to refresh gradle while the game is running

natanfudge:


modmuss50:
Id rarther crash here, and print out a warning to ensure the game is closed.

natanfudge:
Why crash? Refreshing should be able to work just fine.

modmuss50:
it could fail silently in other cases, loom shouldnt be used when the game is open, there is no reason for it to.

natanfudge:
I want to genSources while the game is running. Loom always allowed refreshing when the game is running, it just stopped working when the natives were added.

immibis:
I think it would be better to not extract the file if it hasn't changed.

modmuss50:
Closing this as its not the fix. the fix is to not exact the natives each run.

PR #173, opened 2020-01-18
Replace dots in artifact names when generating mod jsons for JIJ

Juuxel:
This should fix using deps like `org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.0`.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/NestedJars.java
+++ b/src/main/java/net/fabricmc/loom/util/NestedJars.java
@@ -194,7 +194,7 @@ public class NestedJars {
 	private static String getMod(Dependency dependency) {
 		JsonObject jsonObject = new JsonObject();
 		jsonObject.addProperty("schemaVersion", 1);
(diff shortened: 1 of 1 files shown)


natanfudge:
I'd rather crash here, and print out a warning you can't place dots in artifact names.

Chocohead:
Nothing says you cannot, they are perfectly valid Maven artifact names. It is Loader which doesn't allow dots in mod IDs thus it's on Loom to make them conform when generating `fabric.mod.json`s.

natanfudge:
It could fail silently in other cases, artifact names shouldn't contains dots, there are no reasons to allow them.
(based off of this https://github.com/FabricMC/fabric-loom/pull/172)

natanfudge:
Obviously I have no problem with this issue, I just think the reasoning provided by Modmuss against #172 is unreasonable.

Chocohead:
There is a difference between whether you should be running Gradle under Minecraft on the same files presuming which ones Loom can get over not modifying, and whether the Maven spec allows something. One is up to Loom so can fairly be argued either way, the other is not.

natanfudge:
Fair.

modmuss50:
Looks good. Thanks. I will backport to 0.2.6 shortly with some other fixes.

PR #174, opened 2020-01-21
Pretty-print the modified fabric.mod.json in ModResolver

AlexIIL:
...Because this one didn't used to, and it's a lot nicer to read than if it's on a single line.

If there's any reason why this *isn't* wanted then I'd quite like to add a comment here, as it's probably going to annoy me again in the future :p

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/ModProcessor.java
+++ b/src/main/java/net/fabricmc/loom/util/ModProcessor.java
@@ -38,6 +38,7 @@ import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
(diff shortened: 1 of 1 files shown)


asiekierka:
Non-pretty-printed JSONs are probably slightly smaller compressed and slightly faster to read, though the difference probably doesn't even hit the milliseconds.

liach:
do pretty print please. using gson already made things slow, we don't mind if it gets slower

modmuss50:
I see no issue in this, its loom, its not like every milisecond counts.

liach:
well nowadays people don't care about such tiny size differences and the mod json is load only once on game launch.

AlexIIL:
This only affects the fabric.mod.json that loom transforms *after* downloading from maven, but before it's opened by the dev - released jars have always been pretty-printed AFAIK.

PR #175, opened 2020-01-27
Multithreaded asset downloading, should be a lot quicker

modmuss50:

### DIFF
--- a/src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
+++ b/src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
@@ -127,7 +127,7 @@ public class MinecraftAssetsProvider {
 						}
 
 						project.getLogger().debug(":downloading asset " + assetName);
-						progressLogger.progress(assetName + " - " + sha1);
+						progressLogger.progress(String.format("%-30.30s", assetName) + " - " + sha1);
 
 						try {
 							DownloadUtil.downloadIfChanged(new URL(Constants.RESOURCES_BASE + sha1.substring(0, 2) + "/" + sha1), file, project.getLogger(), true);
--- a/src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
+++ b/src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
@@ -28,8 +28,9 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.net.URL;
+import java.util.Deque;
 import java.util.Map;
-import java.util.Stack;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
```diff
@@ -76,18 +80,18 @@ public static void provide(MinecraftProvider minecraftProvider, Project project)
 			}
 		}
 
-		ProgressLogger progressLogger = ProgressLogger.getProgressFactory(project, MinecraftAssetsProvider.class.getName());
-		progressLogger.start("Downloading assets...", "assets");
+		project.getLogger().lifecycle(":downloading assets...");
+
+		Stack<ProgressLogger> loggers = new Stack<>();
+		ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() / 2);
```
dblsaiko: Pretty sure `Runtime.getRuntime().availableProcessors() / 2` will give you 0 when on a single core processor.

modmuss50:
I have aligned the hashes so it looks a lot better when downloading

src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
modmuss50: Sloved it.

dblsaiko (approved):

natanfudge:
How is a network-bound task sped up by multithreading?

modmuss50:
When downloading 1000s of small files a lot of time is spent establishing the network connection. This allows for the bandwidth to be maxed out, in some cases drastically increasing download speed

src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
```diff
@@ -76,18 +80,18 @@ public static void provide(MinecraftProvider minecraftProvider, Project project)
 			}
 		}
 
-		ProgressLogger progressLogger = ProgressLogger.getProgressFactory(project, MinecraftAssetsProvider.class.getName());
-		progressLogger.start("Downloading assets...", "assets");
+		project.getLogger().lifecycle(":downloading assets...");
+
+		Stack<ProgressLogger> loggers = new Stack<>();
```
liach: Deque please

liach (changes requested):

src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java
modmuss50: Why, whats the issue with stack?

Earthcomputer: Stack is deprecated in newer versions of Java and will be removed eventually. Deque is the preferred replacement.

modmuss50: ah, didnt know that. I will sort that out later. Thanks

Chocohead: `Stack` (and all its other pre-collections friends) are obsolete but [not deprecated](http://hg.openjdk.java.net/jdk/jdk/file/e4fc1e578b3a/src/java.base/share/classes/java/util/Stack.java). Java still tries for backwards compatibility where possible so I wouldn't ever see them being removed given they've lasted the best part of 22 years not even being deprecated.

Also if you're using it for the thread safety you'll have to be careful the collection you replace it with is safe too (which `Deque` as an interface doesn't guarantee for example).

modmuss50: moved to a `ConcurrentLinkedDeque`, will test once im at home.

modmuss50:
Tested again, im happy with the changes.

PR #178, opened 2020-02-17
0.2.7 refactors

modmuss50:
This PR contains a bit of cleanup to loom, as well as almost drop in support for access winders, and fixes for idea 2020.

This has the abbility to generate project based jars.

PR #182, opened 2020-02-28
Access widener loom support

modmuss50:
Moved this into its own PR, from #178


modmuss50:
This is now fully inline with the loader PR and has been tested.

PR #186, opened 2020-03-31
Use a custom dev time log4j config. Enables color output and improves overall logging.

modmuss50:
Color output is not enabled by default for eclipse users.

This is based off https://github.com/FabricMC/fabric-loader/pull/216 Another solution for runtime logging will be looked at.

Color output tested in vscode, idea, cmd and eclipse (disabled)

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -60,7 +60,6 @@ public class LoomGradleExtension {
 	public boolean extractJars = false;
 	public String customManifest = null;
 	public File accessWidener = null;
-	public boolean forceAnsiConsole = false;
 	public Function<String, Object> intermediaryUrl = mcVer -> "https://maven.fabricmc.net/net/fabricmc/intermediary/" + mcVer + "/intermediary-" + mcVer + "-v2.jar";
 
 	private List<Path> unmappedModsBuilt = new ArrayList<>();
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -60,6 +60,7 @@ public class LoomGradleExtension {
 	public boolean extractJars = false;
 	public String customManifest = null;
 	public File accessWidener = null;
+	public boolean forceAnsiConsole = false;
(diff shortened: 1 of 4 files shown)


liach (approved):

dblsaiko (changes requested):
Already talked about this on Discord, but I think determining which IDE is currently in use by checking which files exist in the filesystem is very fragile and should be avoided at all costs in favor of better ways of detecting the used IDE.

Instead, I would suggest adding an IDE flag to new run configs so that loader can determine which IDE is in use based on that. This would eliminate improper detection for:

 - using multiple IDEs in parallel
 - repositories containing eclipse project files
 - any other IDE that may be using the same file names

For existing run configurations that do not have that flag yet, a fallback IDE type should be assumed, which uses the lowest common denominator of features (in this case, no ANSI codes by default, like for Eclipse). It should also warn about this on startup so that users can add the new flag to their existing run configurations.

New run configurations should of course be generated with the new flag, for example `-Dfabric.dli.ide=idea` or `-Dfabric.dli.ide=eclipse`, which loader can then determine the IDE from.

The existing flag for force enabling ANSI codes (like it does right now for Eclipse) can persist.

modmuss50:
the ide run args might be useful, but I dont see how they can easily be hooked into this. It would require some changes to DLI to support ide specific options. What happens if someone is using an unsupported IDE?

I dont think its perfect solution either. Im also not a big fan of re-generating the run configs again. its already caused headaces for a lot of people.

dblsaiko:
> What happens if someone is using an unsupported IDE?

Same as for run configurations that don't have the IDE flag, it does not use any features that not everything can handle by default.

> Im also not a big fan of re-generating the run configs again.

They don't have to be regenerated, as I said, what I suggested would be fully backwards compatible with current run configurations. If users don't have the new flag in their arguments, it will fall back to not using ANSI codes since that will work for everything, and warn them about not having the argument. They can just ignore it if they don't care about having colored messages.

Also, in general, adding workarounds just so that existing setups will continue to work as opposed to fixing the issue properly isn't the right way to go in the long term run imo. This is partly how unmaintainable codebases with decades of workarounds that nobody knows anymore why it was done this way and everyone is afraid of touching it because it might break, get created. Especially since it's not even a problem in this case, existing setups won't break, they just won't have ANSI colors. And that only until the next time the project gets set up again.

modmuss50:
I know the detection for ide's is far from great, but its not something I want to change now. Loom is full of hacks like this, so im not too bothered. If there are issues I can address those.

PR #196, opened 2020-04-24
Update to use threaded fernflower

modmuss50:
Requires: https://github.com/FabricMC/intellij-fernflower/pull/6

The max number of threads can be set in the build.gradle using the following:

```groovy
genSourcesDecompile {
	numThreads = 4
}
```

This may produce slighly diffrent results to source, however im not sure if this is cuased by running it multithreaded, or an existing issue in ff.

On my machine (i9-7900x) this shows a 10x decrease in decompile time.

A lot of the ground work for this was already in loom thanks to covers1624.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ dependencies {
 
 	// decompilers
 	implementation ('net.fabricmc:procyon-fabric-compilertools:0.5.35.13')
-	implementation ('org.jetbrains:intellij-fernflower:1.1.0.11')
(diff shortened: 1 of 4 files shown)


immibis:
Is it possible for the setting to be on the command line? This has nothing to do with building the mod, so putting it in the build script doesn't seem ideal.

modmuss50:
Humm, thats a good point. I can prob add a system property for it.

PR #198, opened 2020-04-28
Make the Minecraft dependency non-transitive to subprojects

### LINKED ISSUE #187: Access wideners on subprojects do not attach the correct jar in IDEA
Loom: 0.2.7 build 33
IDEA: 2019.3.3

This happens when an AW is attached to a fabric-api module, for example.

Juuxel:
Fixes #187. This is literally shorter than the workaround 😛 

Tested with FabLabs' `ScreenHandlers` and my Fabric API `ScreenHandler` PR by removing all of the workaround exclusions, and it seems to work fine.

### DIFF
--- a/src/main/java/net/fabricmc/loom/AbstractPlugin.java
+++ b/src/main/java/net/fabricmc/loom/AbstractPlugin.java
@@ -124,7 +124,8 @@ public class AbstractPlugin implements Plugin<Project> {
 			}
 		}
 
-		extendsFrom("compile", Constants.MINECRAFT_NAMED);
+		extendsFrom("compileClasspath", Constants.MINECRAFT_NAMED);
+		extendsFrom("runtimeClasspath", Constants.MINECRAFT_NAMED);
 
 		if (!extension.ideSync()) {
 			extendsFrom("annotationProcessor", Constants.MINECRAFT_NAMED);


liach:
No... if a subproject has a minecraft with custom access wideners, it should just redeclare the minecraft dependency to include the access widener.

Juuxel:
@liach This does... just that?

This PR's goal is to prevent AWs bleeding into other Gradle projects within the same workspace via transitive dependencies, so you don't have to exclude Minecraft when working with subprojects.

Before:
```groovy
// Project A
minecraft {
    accessWidener = file("something")
}
dependencies {
    minecraft "com.mojang:minecraft:20w17a"
}

// Project B
dependencies {
    minecraft "com.mojang:minecraft:20w17a"
    implementation(project(":a")) {
        exclude module: "minecraft" // No! I don't want your AWs here.
    }
}
```

After:
```groovy
// Project A
minecraft {
    accessWidener = file("something")
}
dependencies {
    minecraft "com.mojang:minecraft:20w17a"
}

// Project B
dependencies {
    minecraft "com.mojang:minecraft:20w17a"
    implementation project(":a") // Project A knows how to keep its AWs to itself. :)
}
```

modmuss50:
accessWidener's are desgined to be specific to the project, at runtime they will be re-applyed so it wont be an issue there.

modmuss50 (approved):

JamiesWhiteShirt (approved):
Looks reasonable and the risk of breaking projects seems to be low. I don't see any better way to make it non-transitive.

PR #199, opened 2020-04-28
Move all Loom tasks to the "fabric" group

Juuxel:
This moves all Loom tasks to the same place in Gradle's `tasks` output or IDEA's Gradle tool window. The two tasks with no group were given the `fabric` group, and the run tasks were moved from `minecraftMapped` to `fabric`.

**Exact changes:**
- `cleanLoom`: [no group] -> fabric
- `remapJar`: [no group] -> fabric
- Removed the "minecraftMapped" group that the run tasks were in:
  - `runClient`: minecraftMapped -> fabric
  - `runServer`: minecraftMapped -> fabric


Pyrofab:
Maybe specify "fabric loom" instead of fabric ? There may be other fabric gradle plugins in the future.

modmuss50:
Might be slighly better, but its not a big issues, and not a big issue if another plugins tasks go into the same group

modmuss50 (approved):

natanfudge (approved):

Prospector (approved):

shedaniel (approved):

PR #203, opened 2020-05-06
Update license year to 2020

haykam821:


valoeghese (commented):
ok

liach:
Imo we should just use the initial year 2016

modmuss50:
no, this needs a propper solution like fabric api.

PR #206, opened 2020-05-11
 Use idea-ext plugin to generate run configurations

### LINKED ISSUE #204: JetBrains' gradle-idea-ext-plugin might be able to be used to fix run configurations generation
Specifically the issue with them not showing up until you reopen the project. According to [this][1] it has a way to define run configurations, but I'm not sure if this can be accessed from other gradle plugins and not only from a build script.

https://github.com/jetbrains/gradle-idea-ext-plugin

(Similarly, it might also be able to be used to set the default build method to IntelliJ instead of Gradle instead of having to set that every time opening a project. I do it in my buildscript here, but it would be nice if this could be integrated into Loom: https://github.com/2xsaiko/retrocomputers/commit/6ded51f7808ed27a17e9eeedd49bf508d38ec148)

[1]: https://github.com/JetBrains/gradle-idea-ext-plugin/wiki/DSL-spec-v.-0.1#project-level-settings

dblsaiko:
Use JetBrains' gradle-idea-ext-plugin (https://github.com/jetbrains/gradle-idea-ext-plugin) to create run configurations instead of editing the .idea directory directly. Also sets the build mode to IntelliJ instead of Gradle (see https://github.com/FabricMC/fabric-example-mod/pull/53)

Closes #204


build.gradle
```diff
@@ -29,6 +29,9 @@ repositories {
 		name = "Fabric"
 		url = 'https://maven.fabricmc.net/'
 	}
+	maven {
+		url = 'https://plugins.gradle.org/m2/'
```
modmuss50, line 33: I assume this isnt needed when being used by mods?

modmuss50:
Looks good, I will test this before merging it.

modmuss50 (approved):

build.gradle
liach: just use plugin block

modmuss50: No, not here. As this is saying that loom should depend on it. Not for it to be used in loom's buildscript

liach:
do we still need the idea plugin?

build.gradle
dblsaiko: It's not, this is just so that I can pull the plugin into the classpath when developing. The `project.getPluginManager().apply("org.jetbrains.gradle.plugin.idea-ext");` line does everything when actually running gradle (as far as I can see it's the equivalent of `plugins { id 'org.jetbrains.gradle.plugin.idea-ext' version '0.7' }`
Actually, why does this not need me to specify a version... not sure if I should.

dblsaiko:
> do we still need the idea plugin?

Yeah it's used elsewhere, I already checked if I can get rid of it.

modmuss50:
Can we be sure that this plugin support gradle 4.9? as its being built against 5.4

modmuss50:
Seems to still work when importing idea, would be nice if it could set the compiler output as this is a right pain if you dont know how to fix it.

dblsaiko:
Yeah it is. Maybe in addition to the issue on the idea-ext repo an issue on the IDEA bugtracker on that would be nice because that's definitely been a regression for a while.
Also, the fabric wiki says how to fix that compiler path not set so I don't think it's too big of an issue especially since you can just copy a fixed string in that box as I've found (`$PROJECT_DIR$/out`)

modmuss50:
I have used this as a base for 2nd attempt at using this, however it still seems to suffer from similar issues as before.

See my WIP branch here: https://github.com/modmuss50/fabric-loom-1/tree/idea-ext

Setting the compiler output dir is still not supported (With no progress on your issue), I would also like to support Idea's 2021.3 feature of customisable classpaths (To remove client only libs from the server), this also does not appear to be supported. In my testing it still doesnt fix the issue of the run configs not showing up without restarting idea, so right now I see little benefit to this approach.

Im going to close this PR as there is no point contiuning with it, it was a great resource for my attempt at using it. I might try and use a mixture of this plugin, and modifying the project XML to get the desired affects.

Thanks 👍

PR #207, opened 2020-05-11
Don't remapJar when running runClient or runServer

natanfudge:
This saves about a second in a hello world project, probably a lot more in bigger projects.
Task graphs for reference:
image

Please discuss.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
@@ -164,7 +164,7 @@ public class LoomGradlePlugin extends AbstractPlugin {
 		tasks.register("remapSourcesJar", RemapSourcesJarTask.class);
 
 		tasks.register("runClient", RunClientTask.class, t -> {
-			t.dependsOn("assemble", "downloadAssets");
+			t.dependsOn("classes", "downloadAssets");
 			t.setGroup("minecraftMapped");
 		});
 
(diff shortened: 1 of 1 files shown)


JamiesWhiteShirt:
A reasonable change. Does this end up putting the right classes and resources on the classpath? It would break if the run tasks currently puts the mod _jar_ (what we want to avoid) on the runtime classpath and not the mod _classes_ (what we want). It's probably OK, but worth checking.

modmuss50 (changes requested):
This will break things badly. It must depend on the `jar` task as the jar is added to the classpath.

See here: https://github.com/FabricMC/fabric-loom/blob/dev/0.2.7/src/main/java/net/fabricmc/loom/AbstractPlugin.java#L296

And here: https://github.com/FabricMC/fabric-loom/blob/dev/0.2.7/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java#L70

JamiesWhiteShirt:
This confirms my suspicions. However, the run tasks do not _require_ depending on the `jar` task, as the classes and resources can be added to the classpath as directories (like common IDE run configurations operate) instead of as a jar.

The current state of the PR will break things, though.

modmuss50:
Yes, thanks for thinking of that.

The real fix would be to change how this works to run it on the directory, however I just making it depend on jar would be better than nothing.

natanfudge:
I changed it to jar.

modmuss50 (approved):

PR #208, opened 2020-05-12
Improve remapping speed.

modmuss50:
This is made possible by the updated tiny remapper. This is currenly only 1/2 done.


src/main/java/net/fabricmc/loom/util/ModProcessor.java
```diff
@@ -58,82 +55,34 @@
...
-
-	private static void handleNestedJars(File input, Project project, Configuration config, ResolvedArtifact artifact) throws IOException {
-		try (JarFile jarFile = new JarFile(input)) {
-			JarEntry modJsonEntry = jarFile.getJarEntry("fabric.mod.json");
-
-			if (modJsonEntry == null) {
-				return;
+		for (ModDependencyInfo info : processList) {
+			if (info.getRemappedOutput().exists()) {
+				info.getRemappedOutput().delete();
```
natanfudge, line 73: Does this get called every time you refresh gradle or is there something preventing it after the first time?

src/main/java/net/fabricmc/loom/util/ModProcessor.java
```diff
@@ -58,82 +55,34 @@
...
-			FileUtils.copy(jarStream, nestedFile);
 		}
 
-		File remappedFile = new File(extension.getRemappedModCache(), fileName.substring(fileName.lastIndexOf("/")));
+		remapJars(project, processList);
 
-		processMod(nestedFile, remappedFile, project, config, artifact);
+		for (ModDependencyInfo info : processList) {
+			if (!info.getRemappedOutput().exists()) {
+				throw new RuntimeException("Failed to remap mod" + info);
```
natanfudge, line 81: This won't show why it failed to remap, some error should be thrown earlier that explains why it failed instead of this.

natanfudge (changes requested):

src/main/java/net/fabricmc/loom/util/ModProcessor.java
modmuss50: This should never be reached. I should remove it or throw here.

src/main/java/net/fabricmc/loom/util/ModProcessor.java
modmuss50: This is a catch all, ideally something will throw sooner but there isn’t any more info I can add to this. The log may provide more info.

natanfudge: Alright, you're missing a space there then
```suggestion
				throw new RuntimeException("Failed to remap mod " + info);
```

modmuss50:
In my basic testing this has saved 21 seconds off the setup time of the example mod.

Issue #209, opened 2020-05-14
Can't use Kotlin Annotation Processor 


natanfudge:
- Add kapt to gradle project
```
    id "org.jetbrains.kotlin.kapt" version "1.3.72"
```
- Use some annotation processor e.g.
```
kapt "com.google.auto.service:auto-service:1.0-rc7"
```
- Boom
![image](https://user-images.githubusercontent.com/25990014/81966251-ec412180-9621-11ea-82ee-86c250bd3d1e.png)

[Full stacktrace](https://hatebin.com/pdmwybqfyy)

liach:
Somehow kapt is including the compile classpath in its annotation processor classpath. As it suggests in log, add `kapt.includeCompileClasspath = false` to disable that and try again.

Also blame mixin for shipping that annotation processor 😓

natanfudge:
Wow, that actually did it. Thanks!

natanfudge:
It appears setting that option made it not remap the method `init()`, breaking it in prod.

natanfudge:
Actually, just applying the `kapt` plugin makes it fail to remap.

liach:
Hmm, so you should have loom which should add mixin compile extensions to your classpath. Mind run a `gradlew dependencies` (if you are using modern gradle) and post the dependency tree under `annotationProcessor`

natanfudge:
`annotationProcessor` https://hatebin.com/cmrpiassui
`kapt` https://hatebin.com/hoyyrnilzq

liach:
Hmm, does mixin ap emit any message in gradle? usually it should produce some message if it is ever run

natanfudge:
Without kapt plugin:
```
> Task :compileJava
:setting java compiler args
Could not find refmap definition, will be using default name: ladder-ktx-refmap.json
Note: SpongePowered MIXIN Annotation Processor Version=0.8
Note: ObfuscationServiceFabric supports type: "official:intermediary"
Note: ObfuscationServiceFabric supports type: "official:named"
Note: ObfuscationServiceFabric supports type: "intermediary:official"
Note: ObfuscationServiceFabric supports type: "intermediary:named"
Note: ObfuscationServiceFabric supports type: "named:official"
Note: ObfuscationServiceFabric supports type: "named:intermediary"
Note: ObfuscationServiceMCP supports type: "searge"
Note: ObfuscationServiceMCP supports type: "notch"
[...]

PR #211, opened 2020-05-17
Allow applying kapt and using mixins in other source sets

natanfudge:
This PR fixes the following issues regarding refmaps: 
- Applying kapt makes you unable to build the project, see https://github.com/FabricMC/fabric-loom/issues/209.
- Mixin would not produce refmaps for source sets other than `main`. 
- There was a weird warning regarding refmaps. It is now only logged as `info`. 

Note: This PR might allow Kotlin mixins but that option is not supported.


dblsaiko (approved):
LGTM

Although I would make AnnotationProcessorInvoker<T> and everything extending it be static methods instead of classes since it looks like they get allocated, then you call configureMixin on them, and then they get immediately deleted and not stored anywhere, as far as I can see. But really that's just personal preference.

src/main/java/net/fabricmc/loom/util/mixin/AnnotationProcessorInvoker.java
```diff
@@ -0,0 +1,118 @@
...
+				// Pass named MC classpath to mixin AP classpath
+				processorConfig.extendsFrom(
+								configs.getByName(Constants.MINECRAFT_NAMED),
+								configs.getByName(Constants.MOD_COMPILE_CLASSPATH_MAPPED),
+								configs.getByName(Constants.MAPPINGS_FINAL)
+				);
+
+				// Add Mixin and mixin extensions (fabric-mixin-compile-extensions pulls mixin itself too)
+				project.getDependencies().add(processorConfig.getName(),
+								"net.fabricmc:fabric-mixin-compile-extensions:" + Constants.MIXIN_COMPILE_EXTENSIONS_VERSION);
```
modmuss50, line 104: I know im still against kotlin mixins, but is this really needed on the kotlin annotation processor? Does kapt work without this.

src/main/java/net/fabricmc/loom/util/mixin/AnnotationProcessorInvoker.java
```diff
@@ -0,0 +1,118 @@
...
+
+		for (T task : invokerTasks) {
+			passMixinArguments(task);
+		}
+	}
+
+	static Stream<SourceSet> getNonTestSourceSets(Project project) {
+		return project.getConvention().getPlugin(JavaPluginConvention.class).getSourceSets()
+						.stream()
+						.filter(sourceSet -> !sourceSet.getName().equals("test"));
```
modmuss50, line 116: Why is this needed? Should it be excluding custom test source sets like `testMod` in fabric api?

modmuss50 (changes requested):
This will need some testing, need to make sure the old issues in idea dont come back with this.

src/main/java/net/fabricmc/loom/util/mixin/AnnotationProcessorInvoker.java
natanfudge: When kapt it applied it invokes java mixins instead of the java plugin, so it needs to have mixin and the mc stuff on its classpath

modmuss50: Ok, thats what I assumed.

src/main/java/net/fabricmc/loom/util/mixin/AnnotationProcessorInvoker.java
natanfudge: There is similar code in the old code that excludes sourcesets that include the work `test` in them, so I thought I would do the same (I assume the thought was that there's no point to applying mixin on the test sourceset). I made it only target specifically `test` though because having mixins arbitraily not work because your sourceset has the word `test` in them (like `testMod`) doesn't make sense.

modmuss50: Ok, that sounds resonable. As they are needed on the testMod sourceset so that makes sense. Does it cause any harm if they are ran on the test source set?

natanfudge: No, it doesn't cause any issues. This code is just here to make it more similar to the intention of the old behavior.

src/main/java/net/fabricmc/loom/util/mixin/KaptApInvoker.java
```diff
@@ -0,0 +1,113 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import kotlin.Unit;
```
liach, line 35: Hmm, is kotlin available in gradle environment?

natanfudge: Note that imports don't exist in JVM bytecode; kotlin.Unit is only accessed after an if check made earlier `hasPlugin("org.jetbrains.kotlin.kapt")`, and kapt plugin pulls Kotlin with it.

liach: Just concerned of `ClassNotFoundException` at runtime, asking for a safety check. Should be good since the method (or even this class) theoretically won't be loaded if the kapt plugin was not deteced

liach (approved):

Juuxel (approved):

modmuss50 (approved):
Tested and seems to not break things.

PR #218, opened 2020-06-06
Test across java and gradle versions with github actions.

modmuss50:
This PR moves from travis to github actions, making use of the matrix build feature.

This allows us to test diffrent combinations of gradle and java versions.

I have included a test that runs daily against the current, and current release candiate of gradle (nighly is possibly as well), this should provide a good indicator of when an upcoming gradle version will break loom. Im not sure how well the daily github actions work so may end up removing this.

The code changes too loom are:

- Move to relfection when getting the file property all the time, not just on modern gradle versions.

- Run the tests on the current runtime gradle version, not sure why this was done this way to begin with

- Output the test output to console

This is my first play around with github actions, seems to work well, could be a good replacement for travis for now. Will need to play around with it some more to replace jenkins tho ;)

An example run can be seen here: [https://github.com/modmuss50/fabric-loom-1/actions/runs/127164618](https://github.com/modmuss50/fabric-loom-1/actions/runs/127164618)


modmuss50:
I will of course squish my test commits away when merging

Player3324 (approved):

PR #227, opened 2020-06-21
Rengenerate and download all files with --refresh-dependencies

modmuss50:
This should act as a good way to prevent needing to delete files when things go wrong.

I think I have got everything covered.


modmuss50:
we need this now

PR #228, opened 2020-06-24
Less verbosity + code cleanup

RDIL:
Some minor tweaks.

- Make some messages debug only
- Improved the readability of some comments (and typo fix)
- Sync quote style in `build.gradle`
- Fixed the editorconfig

Signed-off-by: Reece Dunham <me@rdil.rocks>

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -58,7 +58,6 @@ public class LoomGradleExtension {
 	public String loaderLaunchMethod;
 	public boolean remapMod = true;
 	public boolean autoGenIDERuns = true;
-	public boolean verbose = true;
 	public String customManifest = null;
 	public File accessWidener = null;
 	public Function<String, Object> intermediaryUrl = mcVer -> "https://maven.fabricmc.net/net/fabricmc/intermediary/" + mcVer + "/intermediary-" + mcVer + "-v2.jar";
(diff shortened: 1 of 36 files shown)


src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -58,6 +58,7 @@
 	public String loaderLaunchMethod;
 	public boolean remapMod = true;
 	public boolean autoGenIDERuns = true;
+	public boolean verbose = true;
```
modmuss50: Whats the use of this? Should you just use --info if you want more verbose logging?

We might need to tweak some log outputs to be lower

modmuss50:
Seems fine overall, not a fan of the verbose thing, using `--info` or `--debug` should be used to get more log output

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
RDIL: Yeah, I think that right now loom outputs a bit too much info for what it is doing.

RDIL:
The one thing I do want to silence that I can't find the source to is `Note: Writing refmap to /workspace/Cherry/build/classes/java/main/CherryClient-refmap.json` - it is repeated like 15 times and its super annoying

modmuss50 (approved):

modmuss50:
Merge conflicts and then this should be good to go

RDIL:
Fixed.

RDIL:
@modmuss50 all good now!

modmuss50:
Fixed conflicts and merged in my IDE. thanks.

PR #232, opened 2020-06-28
Supply the configuration to the run Exec directly

octylFractal:
This allows the implicit task dependencies from project() dependencies to flow into the run tasks.

For example, if you did `implementation(project("worldedit-core"))`, it would not build `worldedit-core.jar`, despite using it on the classpath. Gradle tracks this information, so using the Configuration directly also adds the `worldedit-core:jar` task as a dependency. This also allows us to remove the dependency on the `jar` task.

### DIFF
--- a/src/main/java/net/fabricmc/loom/AbstractPlugin.java
+++ b/src/main/java/net/fabricmc/loom/AbstractPlugin.java
@@ -49,12 +49,14 @@ import org.gradle.api.tasks.javadoc.Javadoc;
 import org.gradle.plugins.ide.idea.model.IdeaModel;
 
 import net.fabricmc.loom.providers.LaunchProvider;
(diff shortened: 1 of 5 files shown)


natanfudge (approved):
Good shit

modmuss50 (approved):
Looks good! 

I will do some testing to make sure everything still works as expected before merging.

liach (approved):

modmuss50:
It doesnt seem to be picking up changes to the code when re-running the task. Adding back the requirement to the `jar` task fixes this.

natanfudge:
Right it needs to depend on `classes`

modmuss50:
I first tried `classes` and it still didnt seem to want to do it.

modmuss50 (approved):
Tested, works as expected. Thanks a lot for this.

PR #236, opened 2020-07-12
Allow IDEA run configs to be generated for non-root projects

Johni0702:
See individual commits.

Default behavior should be almost unchanged, only difference being that `autoGenIDERuns` now defaults to `false` for subprojects (for backwards compat) and only when it's set to `true`, will run configs be generated (as opposed to before where they were never generated).


Johni0702:
The build failure seems random and unrelated to the changes afaict.

modmuss50 (approved):
Looks good, will merge into 0.5

PR #238, opened 2020-07-17
Improve error handling in the case mappings or minecraft dependency was not provided or provided too much

natanfudge:
Currently it throws an NPE somewhere random.


natanfudge:
Actually, it seems like my NPE is not this, hold on

natanfudge:
Ok it's good now

modmuss50 (approved):

PR #240, opened 2020-07-19
Use jetbrains annotations and remap mc to use them.

modmuss50:
This would need to go into a 0.5 loom as it does have a slim chance of breaking as some people included jsr in their buildscript to work around some compiler oddities.

### DIFF
--- a/src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
+++ b/src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
@@ -24,26 +24,30 @@
 
 package net.fabricmc.loom.providers;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Map;
 import java.util.function.Consumer;
 
+import com.google.common.collect.ImmutableMap;
 import org.gradle.api.Project;
 
 import net.fabricmc.loom.util.TinyRemapperMappingsHelper;
 import net.fabricmc.tinyremapper.OutputConsumerPath;
 import net.fabricmc.tinyremapper.TinyRemapper;
-import net.fabricmc.mapping.tree.TinyMappingFactory;
-import net.fabricmc.mapping.tree.TinyTree;
 import net.fabricmc.loom.util.Constants;
 import net.fabricmc.loom.util.DependencyProvider;
 
 public class MinecraftMappedProvider extends DependencyProvider {
+	private static final Map<String, String> jsrToJetbrains = new ImmutableMap.Builder<String, String>()
+			.put("javax/annotation/Nullable", "org/jetbrains/annotations/Nullable")
+			.put("javax/annotation/Nonnull", "org/jetbrains/annotations/NotNull")
+			.put("javax/annotation/concurrent/Immutable", "org/jetbrains/annotations/Unmodifiable")
+			.build();
+
 	private File minecraftMappedJar;
 	private File minecraftIntermediaryJar;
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
```diff
@@ -118,8 +122,15 @@ private void mapMinecraftJar() throws IOException {
 	}
 
 	public TinyRemapper getTinyRemapper(String fromM, String toM) throws IOException {
+		TinyTree annotationMappingsTree;
+
+		try (BufferedReader reader = new BufferedReader(new InputStreamReader(MinecraftMappedProvider.class.getClassLoader().getResourceAsStream("jsrToJetbrains.tiny")))) {
```
natanfudge: I would have done 
```
Map<String,String> jsrToJetbrains = new HashMap {{
   put("javax/annotation/Nullable", "org/jetbrains/annotations/Nullable");
   ...
}};
```
And then
```
.withMappings(acceptor -> jsrToJetbrains.forEach((jetbrains,jsr) -> acceptor.acceptClass(jetbrains,jsr)))
```

natanfudge (changes requested):

src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
modmuss50: Yes, thats prob better. I expected there to be more to remap possibly including methods.

Player3324 (approved):

shedaniel (approved):

liach:
Imo this should really be done on yarn's side and annotation location needs adjustment as well. Suffices as a bridging practice though.

liach (approved):

Earthcomputer (approved):

src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
```diff
@@ -40,6 +42,12 @@
 import net.fabricmc.loom.util.DependencyProvider;
 
 public class MinecraftMappedProvider extends DependencyProvider {
+	private static final Map<String, String> jsrToJetbrains = new ImmutableMap.Builder<String, String>()
```
Juuxel:
```suggestion
	private static final Map<String, String> JSR_TO_JETBRAINS = new ImmutableMap.Builder<String, String>()
```

Juuxel (changes requested):

src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
JamiesWhiteShirt: Every time you use double brace initialization, god kills a kitten

src/main/java/net/fabricmc/loom/providers/MinecraftMappedProvider.java
liach: guess so, this map is effectively immutable

natanfudge (approved):

ramidzkh (approved):

liach:
@ramidzkh why is it bad to do it on yarn's side? these nullability preferably should be moved to type annotations' locations, and we should really make yarn a documentation system that document extra annotations that *should have been present* in the code.

Earthcomputer:
> that document extra annotations that should have been present in the code

and also change the annoying ones like `@Nullable ClientPlayerEntity player;` and `@Deprecated` from those Block methods (replace with `@OverrideOnly`)

i509VCB:
Really the `@OverrideOnly` handling should be part of yarn imo. Someone should write an RFC for adding extra annotations to tiny spec for use in yarn or however else they could see it working. I am 100% sure you will have to manually catalog all methods and fields that.

> `@Nullable ClientPlayerEntity player;`

That field in MinecraftClient has the intended annotation. The ClientPlayerEntity is only present if in game. Some javadoc for yarn specifying when those nullable fields are available would be nice to add.

Pyrofab (approved):

Earthcomputer:
> That field in MinecraftClient has the intended annotation

But it's annoying to have the IDE warning because most of the time you're using it you know that the player exists.

i509VCB:
I asusme you know you can suppress the warning.

PR #245, opened 2020-07-27
Filter non-existent directories out of the Mercury classpath

### LINKED ISSUE #29: Unable to remap source jar + migrateMappings
As mentioned in the cadix Discord, ever since switching to a setup with split mixin/accessor/main source sets, my remapSourcesJar task has failed with [this error](https://paste.gg/p/anonymous/8d1f04636d62483c8417f13c5fe8662b).

This is occurring in the [Confabricate](https://github.com/zml2008/confabricate) project. My build setup is a bit funky -- the most relevant piece of code for classpath shenanigans is probably https://github.com/zml2008/gradle-plugins/blob/05b31098fad6c008b19818f5b19852397eab1669/opinionated-fabric/src/main/kotlin/ca/stellardrift/build/fabric/OpinionatedFabric.kt#L143.

I'd appreciate some help tracking down what the source of the error is :)

zml2008:
Fixes CadixDev/Mercury#29

This issue was triggered by some of my customizations to loom settings that create separate source sets for mixins and accessors, and add those to the classpath. JDT will not accept classpath entries that don't exist and throws an error, while what's standard for most Gradle tasks is to ignore such classpath entries.

This change makes sure directories exist before they're added to the Mercury classpath. From testing on one of my affected project, this change appears to resolve the issue. However, I am unsure how this would react in a situation where the `Mercury` instance has been created (from `LoomGradleExtension#getOrCreateSrcMappingCache`), but some classpath entries are only created later. Is such a situation even possible?

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/SourceRemapper.java
+++ b/src/main/java/net/fabricmc/loom/util/SourceRemapper.java
@@ -36,6 +36,7 @@ import org.cadixdev.lorenz.MappingSet;
 import org.cadixdev.mercury.Mercury;
 import org.cadixdev.mercury.remapper.MercuryRemapper;
 import org.gradle.api.Project;
+import org.gradle.api.plugins.JavaPlugin;
 import org.zeroturnaround.zip.ZipUtil;
 
 import net.fabricmc.loom.LoomGradleExtension;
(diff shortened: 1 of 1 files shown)


modmuss50:
This doesnt feel right. the real question is why arent those deps there.

The configurations might need resloving

zml2008:
The specific issue isn't the dependencies -- it's directories in my own project that don't exist -- because they were output directories for a source set that had no classes or resources. Running with a debugger attached, these were the specific errors I was getting from JDT before the change: https://i.imgur.com/a8sTTsb.png

Checking for the existence of dependencies isn't necessarily a required part of the patch -- i mostly did those for consistency.

ramidzkh (approved):

zml2008:
I've been poking through some of the Gradle issues targeted for 6.8, and it looks like this might be the correct way to solve things.

https://github.com/gradle/gradle/pull/11764 solves a similar issue in the same way -- and seems to mostly be waiting on more tests to be merged.

zml2008:
Done in https://github.com/FabricMC/fabric-loom/commit/1956c03df2d2a166c118d8440e8cfc5cc420d363

PR #225, opened 2020-06-20
Make the JarProcessor system extensible

ramidzkh:
Rewrites the existing JarProcessor system so that additional ones can be provided (with `LoomGradleExtension#addProcessor`)

PR #247, opened 2020-08-02
Allow adding additional jar processors

natanfudge:


ramidzkh (changes requested):
Doesn't let jar processors modify the final remapped jar. It is useful because e.g. remapped access wideners needs to be placed in the final mod archive

natanfudge:
It doesn't allow processing mod jars as well which I would also like at some point but I would like to get this in soon.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -67,6 +68,7 @@
 	private final ConfigurableFileCollection unmappedMods;
 
 	final List<LoomDecompiler> decompilers = new ArrayList<>();
+	final List<JarProcessor> jarProcessors = new ArrayList<>();
```
ramidzkh: Why package private?

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -84,6 +86,10 @@ public void addDecompiler(LoomDecompiler decompiler) {
 		decompilers.add(decompiler);
 	}
 
+	public void addJarProcessor(JarProcessor processor) {
```
ramidzkh, line 96: Document API methods?

src/main/java/net/fabricmc/loom/processors/JarProcessor.java
```diff
@@ -31,7 +31,13 @@
 public interface JarProcessor {
```
ramidzkh, line 29: https://github.com/FabricMC/fabric-loom/pull/225/files#r460656239

natanfudge: ok

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
natanfudge: ok

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
natanfudge: I thought of using it in LoomGradlePlugin but that didn't go well

TheGlitch76 (approved):
not being able to process dependency jars or remapped jars is an issue, but that can be added later.
This works for 80% of use cases and imo defining the extended API would work better in a different PR anyways to avoid :bike::house:

modmuss50 (approved):

PR #248, opened 2020-08-02
Start on CFRDecompiler line mapping

modmuss50:
This doesnt seem to be working quite right, ive ran out of time to look into this now.


modmuss50:
Im starting again with this.

PR #259, opened 2020-08-19
Mojmap support

ramidzkh:


src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,185 @@
...
+
+	public static Dependency mojmap(Project project, LoomGradleExtension extension, String minecraftVersionRaw) throws IOException {
+		File userCache = extension.getUserCache();
+
+		File mappingsDir = new File(userCache, "mappings");
+		File steps = new File(mappingsDir, "steps");
+
+		File manifests = new File(userCache, "version_manifest.json");
+		DownloadUtil.downloadIfChanged(new URL("https://launchermeta.mojang.com/mc/game/version_manifest.json"), manifests, project.getLogger());
+		String versionManifest = Files.asCharSource(manifests, StandardCharsets.UTF_8).read();
```
liach: Use `MoreFiles` and import jdk `Files`

liach (approved):

valoeghese (approved):
mojmap good and best in game

hYdos (approved):

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+public class MojmapCreator {
+	private static final Gson GSON = new Gson();
+	private static final ClassLoader CLASS_LOADER = null;
+	private static final Map<String, Object> CREATE = new HashMap<>();
+
+	static {
+		CREATE.put("create", "true");
+	}
+
+	public static Dependency mojmap(Project project, LoomGradleExtension extension, String minecraftVersionRaw) throws IOException {
```
shartte, line 72: I'd make this synchronized for good measure, unless DownloadUtil already synchronizes somehow.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+		File userCache = extension.getUserCache();
+
+		File mappingsDir = new File(userCache, "mappings");
+		File steps = new File(mappingsDir, "steps");
+
+		File manifests = new File(userCache, "version_manifest.json");
+		DownloadUtil.downloadIfChanged(new URL("https://launchermeta.mojang.com/mc/game/version_manifest.json"), manifests, project.getLogger());
```
shartte, line 79: Does this tool ensure the target directory structure exists?

Runemoro (approved):

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -98,6 +100,28 @@ public LoomGradleExtension(Project project) {
...
+	 * risk of using them. You may copy and use the mappings for development purposes, but you may not redistribute
+	 * the mappings complete and unmodified. Microsoft makes no warranties, express or implied, with respect to the
+	 * mappings provided here.  Use and modification of source code for Minecraft: Java Edition is governed by the
+	 * Minecraft End User License Agreement available at https://account.mojang.com/documents/minecraft_eula. </code>
+	 * <br>
+	 *
+	 * @param version Minecraft version to lookup
+	 * @return The dependency. This still needs to be applied to an appropriate configuration
+	 * @throws IOException If any I/O exception occurs while downloading, manipulating or writing mappings
+	 */
```
modmuss50: Tottaly pointless here, should be in the console using the actaull license pulled from the mappings. With a bit warning. See discord for a good example/

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -98,6 +100,28 @@ public LoomGradleExtension(Project project) {
...
+	public Dependency mojmap(String version) throws IOException {
```
modmuss50, line 122: Why pass the version in, this should always match the mc version?

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+import org.gradle.api.internal.artifacts.dependencies.DefaultSelfResolvingDependency;
+import org.gradle.api.internal.file.FileCollectionInternal;
+import org.gradle.internal.component.external.model.DefaultModuleComponentIdentifier;
+
+import net.fabricmc.loom.LoomGradleExtension;
+import net.fabricmc.lorenztiny.TinyMappingFormat;
+
+public class MojmapCreator {
+	private static final Gson GSON = new Gson();
+	private static final ClassLoader CLASS_LOADER = null;
```
modmuss50: Dont need this, just cast to null

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+	private static final Map<String, Object> CREATE = new HashMap<>();
```
modmuss50: Better name for this? Or even remove it and just use singletonMap

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+
+	static {
+		CREATE.put("create", "true");
+	}
+
+	public static Dependency mojmap(Project project, LoomGradleExtension extension, String minecraftVersionRaw) throws IOException {
+		File userCache = extension.getUserCache();
```
modmuss50, line 71: All of this processing stuff should be moved to when the dep is actaully being resolved ideally.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+
+		File mappingsDir = new File(userCache, "mappings");
+		File steps = new File(mappingsDir, "steps");
+
+		File manifests = new File(userCache, "version_manifest.json");
+		DownloadUtil.downloadIfChanged(new URL("https://launchermeta.mojang.com/mc/game/version_manifest.json"), manifests, project.getLogger());
+		String versionManifest = MoreFiles.asCharSource(manifests.toPath(), StandardCharsets.UTF_8).read();
+		ManifestVersion mcManifest = GSON.fromJson(versionManifest, ManifestVersion.class);
```
modmuss50: Loom does this in other places, reuse the existing stuff.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+
+		ManifestVersion.Versions version = mcManifest.versions.stream().filter(v -> v.id.equals(minecraftVersionRaw)).findAny().get();
+		String id = version.id;
+
+		File mappingsJar = new File(mappingsDir, "mojmap-" + id + ".jar");
+
+		if (!mappingsJar.exists()) {
```
modmuss50: Should respect refresh deps

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+			File minecraftJson = new File(userCache, "minecraft-" + id + "-info.json");
+			DownloadUtil.downloadIfChanged(new URL(version.url), minecraftJson, project.getLogger());
+
+			MinecraftVersionInfo info = GSON.fromJson(MoreFiles.asCharSource(minecraftJson.toPath(), StandardCharsets.UTF_8).read(), MinecraftVersionInfo.class);
```
modmuss50: Again I think we already have this data ready.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -98,6 +100,28 @@ public LoomGradleExtension(Project project) {
...
 	}
 
+	/**
+	 * Creates a dependency for the official Minecraft Mappings.
+	 * <br>
+	 * Note: It is the full responsibility of the user to bear these mappings. Users must read and accept the
+	 * license on the header of the official Minecraft mappings. As of current, the license header is provided below,
+	 * but may not be correct.
+	 * <br>
+	 * <code>© 2020 Microsoft Corporation. All rights reserved. These mappings are provided "as-is" and you bear the
```
i509VCB, line 111: Maybe specify the date the license was quoted at.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+					ProGuardReader serverReader = new ProGuardReader(new FileReader(serverMappings))) {
+				clientReader.read(mappings);
+				serverReader.read(mappings);
+			}
+
+			MappingSet officialToNamed = mappings.reverse();
+			MappingSet officialToIntermediary;
+
+			File intermediaryJar = new File(steps, "v2-intermediary-" + id + ".jar");
+			DownloadUtil.downloadIfChanged(new URL(extension.getIntermediaryUrl().apply(id)), intermediaryJar, project.getLogger());
```
modmuss50: Again, already done in other palces.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+
+			try (FileSystem fileSystem = FileSystems.newFileSystem(intermediaryJar.toPath(), CLASS_LOADER)) {
+				officialToIntermediary = TinyMappingFormat.DETECT.createReader(fileSystem.getPath("mappings/mappings.tiny"), "official", "intermediary").read();
+			}
+
+			try (FileSystem fileSystem = FileSystems.newFileSystem(URI.create("jar:" + mappingsJar.toURI().toURL()), CREATE)) {
```
modmuss50: IMO its cleaner to just the the ZipUtil lib we already have.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+				Path path = fileSystem.getPath("mappings", "mappings.tiny");
+				Files.createDirectories(path.getParent());
+
+				try (PrintWriter writer = new PrintWriter(Files.newBufferedWriter(path))) {
```
modmuss50: Move this out into another method, or even class.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+					writer.println("v1\tofficial\tintermediary\tnamed");
```
modmuss50: Use v2

i509VCB (approved):

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,186 @@
...
+							writer.println("METHOD\t" + officialName + "\t" + signature.getDescriptor() + "\t" + intermediaryMethod.getObfuscatedName() + "\t" + intermediaryMethod.getDeobfuscatedName() + "\t" + mojmapMethod.getDeobfuscatedName());
+						}
+					});
+
+					writer.println();
+				}
+			}
+		}
+
+		DefaultModuleComponentIdentifier gradleId = new DefaultModuleComponentIdentifier(DefaultModuleIdentifier.newId("loom-generated", "mojmap-" + id), id);
```
modmuss50, line 176: Im not familar with this part of gradle's api, will review again when I have it in front of me.

modmuss50 (changes requested):
Quite a lot of work is required. 

Its not how I imagined doing it as im still unsure if we are happy with converting the mojang mappings into another format. I will have to check with @sfPlayer1 as this was a corncern he raised.

A few other things that would be good:

- [ ] A a simple test

- [ ] Migrate mapping support

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
modmuss50: ~~Just thinking we will need to make sure we dont auto gen names on this later down the pipeline, not too sure how to stop that.~~

Edit: using v2 stops this.

liach: v2 is troubled because v2 always tries to do stupid merging with official -> intermediary even when official exists in a v2 mapping

modmuss50: using v1 isnt the solution to that

liach: Why? mojmap does not contain any information v1 doesn't support

modmuss50: v1 is technically deprecated. Pulling in already outdated code seems pointless. I’m not too sure what the issue you were taking about is? Is it something that can be fixed?

liach: Yeah, shouldn't be too hard to fix. It's just make v2 mappings skip the merge process if an official namespace already exists

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
ramidzkh: That requires the minecraft configuration to already have a dependency on it, which isn't how it is right now

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Idea loves to automatically clean that up but sure

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: That requires touching more Gradle internals

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Resolved later in MinecraftProvider

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Resolved later in MinecraftProvider

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
modmuss50: Ok, it really needs doing tho as it allows you to clean up a lot of code, other option is to expose it using one of the existing file mavens loom registers.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
modmuss50: Yes, becuase its running too soon, see the other comment.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Moved into MinecraftProvider

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Only these two lines are common. In MappingProvider, it is read with a ZipFS and instantly merged. Here, it is read with ZipUtil and then lorenz-tiny is used

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: > All of this processing stuff should be moved to when the dep is actaully being resolved ideally.

I'm guessing you mean into MappingProvider or something like that. Well, then 1. You cannot use the artifact created by mojmap to be used outside of the mappings configuration and 2. That would hardcore behavior there, which would be more hidden

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Also it is a mess to extract common code from `MinecraftProvider` as it also can use the custom manifest from there. Now that I think of it, this should also use the custom manifest if present

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -98,6 +101,28 @@ public LoomGradleExtension(Project project) {
...
+	 * Note: It is the full responsibility of the user to bear these mappings. Users must read and accept the
+	 * license on the header of the official Minecraft mappings. As of 15th August 2020, the license header is provided
+	 * below, but may not be correct.
+	 * <br>
+	 * <code>© 2020 Microsoft Corporation. All rights reserved. These mappings are provided "as-is" and you bear the
+	 * risk of using them. You may copy and use the mappings for development purposes, but you may not redistribute
+	 * the mappings complete and unmodified. Microsoft makes no warranties, express or implied, with respect to the
+	 * mappings provided here.  Use and modification of source code for Minecraft: Java Edition is governed by the
+	 * Minecraft End User License Agreement available at https://account.mojang.com/documents/minecraft_eula. </code>
+	 * <br>
```
modmuss50, line 116: remove this its outdated already.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -355,7 +380,7 @@ public boolean ideSync() {
 	// Ideally this should use maven, but this is a lot easier
 	public Function<String, String> getIntermediaryUrl() {
 		// Done like this to work around this possibly not being a java string...
-		return s -> intermediaryUrl.apply(s).toString();
+		return s -> intermediaryUrl.apply(UrlEscapers.urlFragmentEscaper().escape(s)).toString();
```
modmuss50, line 383: No, I wouldnt do this here? why this unrelated change?

src/main/java/net/fabricmc/loom/providers/MappingsProvider.java
```diff
@@ -165,8 +164,7 @@ private void storeMappings(Project project, MinecraftProvider minecraftProvider,
 			// These are unmerged v2 mappings
 
 			// Download and extract intermediary
-			String encodedMinecraftVersion = UrlEscapers.urlFragmentEscaper().escape(minecraftVersion);
-			String intermediaryArtifactUrl = getExtension().getIntermediaryUrl().apply(encodedMinecraftVersion);
+			String intermediaryArtifactUrl = getExtension().getIntermediaryUrl().apply(minecraftVersion);
```
modmuss50, line 167: undo please try and keep PRs consise.

src/main/java/net/fabricmc/loom/providers/MinecraftProvider.java
```diff
@@ -119,41 +121,7 @@ private void initFiles() {
...
-		if (getExtension().customManifest != null) {
-			ManifestVersion.Versions customVersion = new ManifestVersion.Versions();
-			customVersion.id = minecraftVersion;
-			customVersion.url = getExtension().customManifest;
-			optionalVersion = Optional.of(customVersion);
-			getProject().getLogger().lifecycle("Using custom minecraft manifest");
-		}
-
-		if (!optionalVersion.isPresent()) {
-			optionalVersion = mcManifest.versions.stream().filter(versions -> versions.id.equalsIgnoreCase(minecraftVersion)).findFirst();
```
modmuss50, line 155: Just store this value in field with a getter? so much easier

src/main/java/net/fabricmc/loom/util/DownloadUtil.java
```diff
@@ -60,7 +61,14 @@ public static void downloadIfChanged(URL from, File to, Logger logger) throws IO
 	 * @throws IOException If an exception occurs during the process
 	 */
 	public static void downloadIfChanged(URL from, File to, Logger logger, boolean quiet) throws IOException {
-		HttpURLConnection connection = (HttpURLConnection) from.openConnection();
+		URLConnection urlConnection = from.openConnection();
+
+		if (!(urlConnection instanceof HttpURLConnection)) {
+			Files.asByteSink(to).writeFrom(urlConnection.getInputStream());
```
modmuss50, line 67: No this is awful, this is for downloading files. Again please keep the PR minimal.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
modmuss50: No, you can run this when the dep is resloved. Ill be honest I forget how to do it but im 95% sure there is a better time to handle this.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,244 @@
...
+				DownloadUtil.downloadIfChanged(new URL(serverMappingsUrl), serverMappings, project.getLogger());
+
+				MappingSet mappings = MappingSet.create();
+
+				// Print license headers
+				String clientMappingsString = MoreFiles.asCharSource(clientMappings.toPath(), StandardCharsets.UTF_8).read();
+				String serverMappingsString = MoreFiles.asCharSource(serverMappings.toPath(), StandardCharsets.UTF_8).read();
+
+				{
+					project.getLogger().warn("USING OFFICIAL MAPPINGS. YOU ARE USING THESE MAPPINGS AT YOUR OWN RISK UNDER THE FOLLOWING LICENSE:");
```
modmuss50, line 106: this might need some better wording but I guess it will do

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,244 @@
...
+				String serverMappingsUrl = info.downloads.get("server_mappings").url;
+
+				DownloadUtil.downloadIfChanged(new URL(clientMappingsUrl), clientMappings, project.getLogger());
+				DownloadUtil.downloadIfChanged(new URL(serverMappingsUrl), serverMappings, project.getLogger());
+
+				MappingSet mappings = MappingSet.create();
+
+				// Print license headers
+				String clientMappingsString = MoreFiles.asCharSource(clientMappings.toPath(), StandardCharsets.UTF_8).read();
+				String serverMappingsString = MoreFiles.asCharSource(serverMappings.toPath(), StandardCharsets.UTF_8).read();
```
modmuss50, line 103: Use buffered readers, I think you can mark and then reset back to the beggining of the file after the license has been printed.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,244 @@
...
+					serverReader.read(mappings);
+				}
+
+				officialToNamed = mappings.reverse();
+			}
+
+			// Download and parse intermediaries
+			{
+				File intermediaryJar = new File(steps, "v2-intermediary-" + id + ".jar");
+				DownloadUtil.downloadIfChanged(new URL(extension.getIntermediaryUrl().apply(id)), intermediaryJar, project.getLogger());
```
modmuss50, line 131: This is done else where, just use what has already been done. Depends on moving this logic to a later time.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
```diff
@@ -0,0 +1,244 @@
...
+					}
+				});
+			});
+
+			// Write
+			{
+				ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+
+				try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8))) {
+					new TinyWriter(writer, "intermediary", "named").write(intermediaryToNamed);
```
modmuss50, line 169: Can we somehow copy the license over to the new file?

modmuss50 (changes requested):
Still not happy with this. It should not do all that downloading at evaulation time. Im fairly sure there is a better way to do it if not, ill think again.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
ramidzkh: Related since `MojmapCreator` uses this to download the intermediaries, and it would need to escape the version in both `MappingsProvider` and `MojmapCreator` anyway

src/main/java/net/fabricmc/loom/util/DownloadUtil.java
ramidzkh: I was testing custom manifests. Will remove

src/main/java/net/fabricmc/loom/providers/MappingsProvider.java
ramidzkh: See above comment

modmuss50:
@ramidzkh I have got a local copy of this PR with some improvements, I will be pushing soon. Thanks for the help I will ensure to credit you.

src/main/java/net/fabricmc/loom/util/MojmapCreator.java
ramidzkh: Marking with a large limit would be equivalent to just reading it to a buffer (this case a string) anyway, I don't see any benefits to this

modmuss50:
Thanks for the help, I based off this PR and cleaned it up a bit while fixing some issues.

https://github.com/FabricMC/fabric-loom/commit/9029a699d93df18c5c845a5afc34b9dbf380f82b

ramidzkh:
🦀

TheMadHau5:
🦀

PR #267, opened 2020-08-29
Split out and organize some constants

i509VCB:
This pull request reorganizes parts of the constants class into categories based on sub-classes and moves some constants into this class


zml2008 (approved):
This is a good change to help get the codebase organized. I'd like to see this in sooner rather than later so that the structure is there, even if it doesn't hit every constant just yet.

Once this PR provides the structure, it'll be easier for any PRs in flight to add to the constants without as many merge conflicts.

i509VCB:
Ah yes conflicts again

modmuss50 (approved):
Looks good, fix conflicts and then this can go in.

i509VCB:
conflicts should be fixed

Issue #191, opened 2020-04-11
Question: remapJar with "official" mappings


LeafHacker:
I have a mod that needs to target non-fabric environments, however it'd still be nice to use fabric-loom for developing it. To this end I'd like to write a JarTask that generates an `official` mapped jar complete with a `named:official` mixin refmap.

I can see fabric-mixin-compile-extensions lists `ObfuscationServiceFabric supports type: "named:official"` and tiny-remapper of course supports this too. I'm just wondering what the best way to approach this would be, e.g. can this be done by just configuring fabric-loom in a specific way or should I be writing a remapJar inspired task from scratch?

In essence I'm asking how you would approach this as I'm fairly unfamiliar with the ecosystem and implementation details.

modmuss50:
Looking at the code it currenly seems like this is hardcoded, but it shouldnt be too hard for me to expand this out to support changing the output namespace.

LeafHacker:
That'd be great, thanks

modmuss50:
This was done with the rewrite of remapping jars.

PR #268, opened 2020-08-29
Allow specification of namespaces to use in RemapJarTask

i509VCB:
This is a draft since setting `toM` as `official` has mapping conflicts, this is probably out of our control tbh.

But for the future this may have additional use for exporting to other mapping sets (such as mojmap) in the future.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
+++ b/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Set;
 
 import com.google.common.base.Preconditions;
@@ -50,6 +51,7 @@ import net.fabricmc.loom.util.NestedJars;
 import net.fabricmc.loom.util.TinyRemapperMappingsHelper;
 import net.fabricmc.loom.util.accesswidener.AccessWidenerJarProcessor;
 import net.fabricmc.loom.util.JarRemapper;
+import net.fabricmc.mapping.tree.TinyTree;
 import net.fabricmc.stitch.util.Pair;
 import net.fabricmc.tinyremapper.OutputConsumerPath;
 import net.fabricmc.tinyremapper.TinyRemapper;
@@ -59,6 +61,8 @@ public class RemapJarTask extends Jar {
 	private final RegularFileProperty input;
 	private final Property<Boolean> addNestedDependencies;
 	private final Property<Boolean> remapAccessWidener;
+	private final Property<String> fromM;
+	private final Property<String> toM;
 	public JarRemapper jarRemapper;
 
 	public RemapJarTask() {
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -176,8 +197,20 @@ public void scheduleRemap() throws Throwable {
...
+		List<String> namespaces = mappings.getMetadata().getNamespaces();
+
+		// Validate the namespaces are supported in the tiny tree
+		if (!namespaces.contains(fromM)) {
+			RemapJarTask.throwMissingNamespaceException(project, fromM, namespaces);
+		}
+
+		if (!namespaces.contains(toM)) {
+			RemapJarTask.throwMissingNamespaceException(project, toM, namespaces);
+		}
```
Juuxel:
```suggestion
		if (!namespaces.contains(fromM)) {
			throwMissingNamespaceException(project, fromM, namespaces);
		}

		if (!namespaces.contains(toM)) {
			throwMissingNamespaceException(project, toM, namespaces);
		}
```
There's no need to specify the class name here.

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -91,8 +100,20 @@ public void doSingleRemap() throws Throwable {
...
+		List<String> namespaces = mappings.getMetadata().getNamespaces();
+
+		// Validate the namespaces are supported in the tiny tree the namespaces are supported in the tiny tree
+		if (!namespaces.contains(fromM)) {
+			RemapJarTask.throwMissingNamespaceException(project, fromM, namespaces);
+		}
+
+		if (!namespaces.contains(toM)) {
+			RemapJarTask.throwMissingNamespaceException(project, toM, namespaces);
+		}
```
Juuxel:
```suggestion
		if (!namespaces.contains(fromM)) {
			throwMissingNamespaceException(project, fromM, namespaces);
		}

		if (!namespaces.contains(toM)) {
			throwMissingNamespaceException(project, toM, namespaces);
		}
```
Same here.

Juuxel (changes requested):
Looks nice and this would be useful in one of my Loom fork's branches :eyes:

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
i509VCB: I qualify these things so I know whether a method is static or belongs to a class file (usually via `this`)

i509VCB: Of course no conventions exist on whether member references should be qualified or not

LeafHacker:
This looks like it would resolve #191, could you clarify what you mean by _“official has mapping conflicts, this is probably out of our control tbh”_? Are you saying Tiny remapper has issues mapping a jar to `official` because the member names are sometimes duplicated?

modmuss50:
This is also missing changing the output of the AW remapping, and the mixin refmap + output mappings

modmuss50:
This was handled when I rewrote this. Thanks.

PR #271, opened 2020-08-31
Mark auto-generated mods with a custom value

Haven-King:
Make the barebones `fabric.mod.json` generated for non-mod dependencies an API for ModMenu purposes to prevent mods with a large number of dependencies from cluttering the menu.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/NestedJars.java
+++ b/src/main/java/net/fabricmc/loom/util/NestedJars.java
@@ -212,7 +212,7 @@ public class NestedJars {
 		jsonObject.addProperty("name", dependency.getModuleName());
 
 		JsonObject custom = new JsonObject();
-		custom.addProperty("modmenu:api", true);
+		custom.addProperty("fabric-loom:generated", true);
 		jsonObject.add("custom", custom);
 
 		return GSON.toJson(jsonObject);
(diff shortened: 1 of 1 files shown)


JamiesWhiteShirt:
As a more general solution, I would suggest exposing the JSON document to be configurable in the build script. We should not have to hardcode support for a third-party mod.

modmuss50:
Yes, we dont want mod specific handling in loom. Either we need to expand the loader api to support this or have a way to specify the generated json. Ideally per dep but im not sure how that would work.

Haven-King:
I agree that this isn't what a final solution would look like, and that it shouldn't explicitly hardcode support for a third-party mod. I mostly wanted to start the conversation on this topic so that those with more experience with Gradle than I could talk about possible solutions.

shedaniel:
Just name it `"fabric-loom:generated": true`, and it is all fixed.

Haven-King:
> Just name it `"fabric-loom:generated": true`, and it is all fixed.

I like this idea.

Player3324 (approved):
IMO this works for now, don't want to give any guarantees that the property will always be present - it is just a hint!

Martmists-GH (approved):

modmuss50 (approved):
Looks good.

LambdAurora (approved):

TwilightFlower (approved):

PR #277, opened 2020-09-13
Export rumtime remap classpath to DLI

modmuss50:
required for: https://github.com/FabricMC/fabric-loader/pull/241


modmuss50:
Going to go with this for now, it may have some issues with subprojects we can smooth those issues out over time. It should work well for the most common uses cases however.

PR #281, opened 2020-09-19
Made migrateMappings source-set aware and added an inPlace option

shartte:


src/main/java/net/fabricmc/loom/task/MigrateMappingsTask.java
```diff
@@ -58,23 +70,23 @@
...
-	private Path inputDir;
-	private Path outputDir;
+	private List<String> sourceSets;
 	private String mappings;
+	private boolean inPlace;
 
 	public MigrateMappingsTask() {
-		inputDir = getProject().file("src/main/java").toPath();
-		outputDir = getProject().file("remappedSrc").toPath();
+		sourceSets = new ArrayList<>();
```
zml2008, line 78: Why not use the `Property` and `DomainObjectSet` API? Gradle seems to be preferring that api for its newer tasks, and some portion of the API exists in 4.10 already: https://docs.gradle.org/4.10.2/userguide/lazy_configuration.html

src/main/java/net/fabricmc/loom/task/MigrateMappingsTask.java
```diff
@@ -87,25 +99,67 @@ public void doTask() throws Throwable {
...
 
-		if (!Files.exists(inputDir) || !Files.isDirectory(inputDir)) {
-			throw new IllegalArgumentException("Could not find input directory: " + inputDir.toAbsolutePath());
+		File mappings = loadMappings();
+		MappingsProvider mappingsProvider = extension.getMappingsProvider();
+
+		MappingSet mappingSet = loadMappingSet(mappings, mappingsProvider);
+
+		Path projectDir = project.getProjectDir().toPath();
+		Path outputBaseDir = project.getBuildDir().toPath().resolve("remapped");
```
zml2008, line 110: This should probably match up with the name of the task -- `mappingsMigrated` or something?

Also would be good to make this configurable.

zml2008 (changes requested):
A few small things that stood out to me.

I was originally thinking each source set should be its own gradle task -- but loading, resolving, and merging mappings for every source set probably knocks out any efficiency gains that'd come from working in parallel.

This also fixes #246

shartte:
It's so far out of date, it doesn't make a lot of sense to update this 😓

PR #282, opened 2020-09-24
Add ability to configure a separate testmod source set

zml2008:
This is a common build element in library mods, including fabric API. The test mod is designed for easily testing functionality at dev time.

A modified version of the example mod that uses this functionality is [available here](https://github.com/FabricMC/fabric-example-mod/compare/master...zml2008:feature/testmod)

This is based on the configurations that [myself](https://github.com/zml2008/gradle-plugins/blob/73d5db701471201c1ae5a6048e6b6cc6ce2f4bbd/opinionated-fabric/src/main/kotlin/ca/stellardrift/build/fabric/OpinionatedFabric.kt#L56) and [i509](https://github.com/i509VCB/Escalate/blob/master/escalate-fabric/src/main/kotlin/me/i509/build/fabric/EscalateFabricExtension.kt#L72) have done for our own mods. The main addition here is the beginnings of support for producing a remapped jar for testmods. This is also the part that is currently incomplete.

The main blocker for remapped jars is properly configuring the Mixin annotation processor to generate different refmaps per source set. Currently the same refmap is written to for each compile task, so when creating a production jar one of the main or testmod jars will be invalid. This will involve deprecating and replacing the `refmapName` field on LoomGradleExtension, and figuring out a way to link compile tasks with remap tasks. Perhaps the AP invokers should get a bit of a redesign...

One other limitation that will be complicated to resolve is testmod-specific dependencies and JiJ. To avoid that, `Constants.MOD_COMPILE_ENTRIES` would have to be replaced with something that dynamically creates the `mod*` configurations for new source sets. However, most testmods don't need their own dependencies, so that can be kicked down the road a bit.

testmod-specific access wideners are also not supported, and would need further refactoring to support. I don't think it's worth the effort, since that substantial of a test mod probably warrants a subproject.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -48,6 +48,13 @@ import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.plugins.BasePluginConvention;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.SourceSetContainer;
+import org.gradle.api.tasks.TaskProvider;
+import org.gradle.api.tasks.bundling.AbstractArchiveTask;
+import org.gradle.jvm.tasks.Jar;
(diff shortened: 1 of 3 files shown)


modmuss50:
This has now all been taken care of 0.6 and somewhat more in 0.7. Its not as provided on a plate as this is but still should be trivial for a mod to do on their own.

Thanks 👍

PR #285, opened 2020-09-26
eclipse: Fix run configuration generation

### LINKED ISSUE #283: Eclipse run configs are broken
The working directory is being set to `${project_loc:%MODULE%}/run` instead of `${project_loc}/run`. This is not valid.

zml2008:
Fixes #283

Main issue here is it looks like someone changed the placeholders the run config generator uses without updating the template.

I've also changed to `workspace_loc` rather than `project_loc` -- which may make no difference, but may make things more reliable if eclipse deselects the project associated with the run config somehow.


OroArmor:
Please merge this in, both I and Pepper from the discord have encountered this issue, and it makes running eclipse projects very difficult.

OroArmor (approved):

PR #296, opened 2020-11-11
Side stripping Jar processor

ramidzkh:
Allows users to remove client or server only members from their local development Minecraft Jar

### DIFF
--- a/src/main/java/net/fabricmc/loom/processors/side/ClassStripper.java
+++ b/src/main/java/net/fabricmc/loom/processors/side/ClassStripper.java
@@ -1,0 +1,80 @@
+/*
+ * This file is part of fabric-loom, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2016, 2017, 2018 FabricMC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


ramidzkh:
Usage:
```groovy
import net.fabricmc.loom.processors.side.SideStripperJarProcessor

minecraft {
	addJarProcessor(SideStripperJarProcessor.SERVER) // For keeping only common or server-only members
        // addJarProcessor(SideStripperJarProcessor.CLIENT) For keeping only common or client-only members
}
```

NickAcPT:
Just wondering, does this affect the Mixins configuration at all? If it doesn't, is it needed?

ramidzkh:
Nope, just removes client/server only classes so you don't need to deal with it. My purpose is so that I can fully remove client only members from [CannedTater](https://github.com/FukkitMC/CannedTater) and make it actually launch without anonymous classes ruining it for everyone. Also, my reason for making this a feature PR is that more people can use it (lets say you want to develop a server-side only mod), and that implementing a processor like this in your own buildscript would be painful

hYdos (approved):

Ueaj-Kerman (approved):

modmuss50:
Going to close this PR, it should now be trivial to do properly using the new changes I added to support splitting the jars.

PR #304, opened 2020-12-05
Respect reproducibleFileOrder and preserveFileTimestamps

CoolMineman:
Makes loom respect reproducibleFileOrder and preserveFileTimestamps.
Allows for reproducible builds of all libs in the example mod with the following modification:
```
// Make Jars Reproducable
tasks.withType(AbstractArchiveTask) {
  preserveFileTimestamps = false
  reproducibleFileOrder = true
}
```
Tested locally with the example mod.
scheduleRemapSources has been modified to take reproducibleFileOrder and preserveFileTimestamps as parameters.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
+++ b/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
@@ -40,6 +40,7 @@ public class ZipReprocessorUtil {
 		if (!reproducibleFileOrder && preserveFileTimestamps) {
 			return;
 		}
+
 		try (ZipFile zipFile = new ZipFile(file)) {
 			ZipEntry[] entries;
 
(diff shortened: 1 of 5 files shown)


src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipOutputStream;
+
+public class ZipReprocessorUtil {
+	private ZipReprocessorUtil() { }
+
+	public static void reprocessZip(File file, boolean reproducibleFileOrder, boolean preserveFileTimestamps) throws IOException {
```
modmuss50, line 39: Looks good, I think there was talk about solving this in tiny remapper but this will do for now.

modmuss50 (approved):

PR #305, opened 2020-12-06
EnumWidener

nnym:
A **compilation time-only** JAR processor that unsets the `enum` modifier from enums and their fields and fixes enum constructor signatures to make name and ordinal parameters accessible. Further, it makes the synthetic enum array visible. Combined with AccessWidener, it allows enums to be instantiated (via the `new` opcode; the JLS does not forbid it) and extended.

Since unsetting the `enum` modifier at runtime is dangerous, compilation and runtime dependencies have to be separated. This separation is achieved by replacing the previous `minecraftNamed` configuration by 2 new configurations: `minecraftNamedCompile` and `minecraftNamedRuntime` that correspond to separate JAR dependencies, and adding appropriate archive classifiers in order to differentiate them.

usage in `build.gradle`: 
```groovy
minecraft {
	enumWidener("net/minecraft/SomeEnum")
}
```


src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -151,6 +152,14 @@ public JsonObject getInstallerJson() {
 		return installerJson;
 	}
 
+	public void enumWidener(String klass) {
+		if (this.enumWidener == null) {
+			this.enumWidener = new ArrayList<>();
```
ramidzkh: Maybe initialize the field in the constructor instead? Would look neater

src/main/java/net/fabricmc/loom/processors/EnumWidenerJarProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+	public void process(File file) {
+		if (this.classes != null && file.equals(this.getTarget())) {
+			this.project.getLogger().lifecycle(String.format("EnumWidener(tm) v0 is in action on %s.", file));
+
+			ZipUtil.transformEntries(file, this.classes.stream()
+						.map(klass -> new ZipEntryTransformerEntry(klass.replace('.', '/') + ".class", new EnumWidenerTransformerEntry(this.project, klass)))
+						.toArray(ZipEntryTransformerEntry[]::new)
+			);
+
+			ZipUtil.addEntry(file, HASH_FILE_NAME, ByteBuffer.allocate(4).putInt(this.classes.hashCode()).array());
```
ramidzkh: Access wideners do a SHA-256 hash instead

src/main/java/net/fabricmc/loom/processors/EnumWidenerJarProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	@Override
+	public boolean isInvalid(File file) {
+		if (this.classes == null || !file.equals(this.getTarget())) {
```
ramidzkh: ~~If the enumwideners were removed, the file would be marked not invalid.~~ Seems to be a flaw with the current design of the current JarProcessor system really, as the same is true if all access wideners are removed but another JarProcessor is present.

~~I would also remove the getTarget methods and its associated checks~~ Maybe make JarProcessors aware if they are operating on the compile or runtime jar?

src/main/java/net/fabricmc/loom/processors/MinecraftProcessedProvider.java
```diff
@@ -50,23 +53,27 @@ public MinecraftProcessedProvider(Project project, JarProcessorManager jarProces
 
 	@Override
 	protected void addDependencies(DependencyInfo dependency, Consumer<Runnable> postPopulationScheduler) {
-		if (jarProcessorManager.isInvalid(projectMappedJar) || isRefreshDeps()) {
+		if (jarProcessorManager.isInvalid(projectMappedCompileJar) || isRefreshDeps()) {
```
ramidzkh: Check if both are invalid

asiekierka:
How does this interact with enum switch statements? Particularly, a case with a non-void method which has its return statements in an enum switch-case; does this tackle that problem, or is it left to the programmer to avoid such cases?

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassNode.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package net.fabricmc.loom.util.enumwidener;
+
+import org.objectweb.asm.FieldVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.ClassNode;
+
+public class EnumWidenerClassNode extends ClassNode {
```
i509VCB: Class visitor?

Ueaj-Kerman:
https://github.com/SpongePowered/Mixin/issues/387

According to 🌍🖥️ it's not a problem. no idea how it works, but it works, and if oracle says adding enums doesn't break binary compatibility, you best believe it doesn't. oracle will break your legs if you break binary compat

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassNode.java
Ueaj-Kerman: no, I think he's trying to save on object here. kindof a microoptimization but whatever

Ueaj-Kerman (approved):
if it works, looks good

i509VCB:
Wouldn't this need a runtime component? or does it not need that.

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassNode.java
i509VCB: You'd still have to allocate the classreader and classwriter though. Just the visitor would be lower overhead.

Arc-blroth:
As far as I know, this would require a runtime component - access hasn't added that yet as far as I know

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassNode.java
nnym: I had a ClassVisitor earlier but I needed a ClassNode in order to shift the parameter annotations due to the 2 additional parameters in constructor signatures.

src/main/java/net/fabricmc/loom/AbstractPlugin.java
```diff
@@ -115,8 +115,10 @@ public void apply(Project target) {
 		Configuration modCompileClasspathMappedConfig = project.getConfigurations().maybeCreate(Constants.Configurations.MOD_COMPILE_CLASSPATH_MAPPED);
 		modCompileClasspathMappedConfig.setTransitive(false);
 
-		Configuration minecraftNamedConfig = project.getConfigurations().maybeCreate(Constants.Configurations.MINECRAFT_NAMED);
-		minecraftNamedConfig.setTransitive(false); // The launchers do not recurse dependencies
+		Configuration minecraftNamedCompileConfig = project.getConfigurations().maybeCreate(Constants.Configurations.MINECRAFT_NAMED_COMPILE);
```
modmuss50: It should only use the split configurations if required.

Arc-blroth:
> the JLS does not forbid it

Actually, re-reading access' original post, this would not require a runtime component

src/main/java/net/fabricmc/loom/processors/MinecraftProcessedProvider.java
```diff
@@ -50,23 +53,27 @@ public MinecraftProcessedProvider(Project project, JarProcessorManager jarProces
...
 		}
 
 		getProject().getRepositories().flatDir(repository -> repository.dir(getJarDirectory(getExtension().getProjectPersistentCache(), PROJECT_MAPPED_CLASSIFIER)));
 
-		getProject().getDependencies().add(Constants.Configurations.MINECRAFT_NAMED,
-				getProject().getDependencies().module("net.minecraft:minecraft:" + getJarVersionString(PROJECT_MAPPED_CLASSIFIER)));
+		getProject().getDependencies().add(Constants.Configurations.MINECRAFT_NAMED_COMPILE,
+				getProject().getDependencies().module("net.minecraft:minecraft:" + getJarVersionString(PROJECT_MAPPED_COMPILE_CLASSIFIER)));
+		getProject().getDependencies().add(Constants.Configurations.MINECRAFT_NAMED_RUNTIME,
+					getProject().getDependencies().module("net.minecraft:minecraft:" + getJarVersionString(PROJECT_MAPPED_RUNTIME_CLASSIFIER)));
```
modmuss50: Again here, only use the split jar if required.

src/main/java/net/fabricmc/loom/processors/MinecraftProcessedProvider.java
```diff
@@ -50,23 +53,27 @@ public MinecraftProcessedProvider(Project project, JarProcessorManager jarProces
...
-				FileUtils.copyFile(super.getMappedJar(), projectMappedJar);
+				FileUtils.copyFile(super.getMappedJar(), projectMappedCompileJar);
+				FileUtils.copyFile(super.getMappedJar(), projectMappedRuntimeJar);
 			} catch (IOException e) {
 				throw new RuntimeException("Failed to copy source jar", e);
 			}
 
-			jarProcessorManager.process(projectMappedJar);
+			jarProcessorManager.process(projectMappedCompileJar);
+			jarProcessorManager.process(projectMappedRuntimeJar);
```
modmuss50: No need to run the same processors on 2 jars.

I think the jar processor API should get a bit of an overhall to support this in a much nicer way.

i509VCB:
> enumWidener("net/minecraft/SomeEnum")

Should every enum have to be specified in loom or do we want to do something like a new file or make it part of AW and just ignore it at runtime.

Ueaj-Kerman:
I don't think you'll be widening enough enums to justify a separate file

dexman545:
How are `default` cases of enum switcghes handled, particularly the ones that throw an exception? Should they just be avoided? If so, in a multimod environment that could lead to issues - some mod adds another `Hand`, but another mod decided to throw on any unexpected `Hand` value.

i509VCB:
For exceptions you could just inject at NEW and cancel? However for cases where there is more logic afterwards that is not suitable.

Whats does jvm do if you try to throw null?

nnym:
> How does this interact with enum switch statements? Particularly, a case with a non-void method which has its return statements in an enum switch-case; does this tackle that problem, or is it left to the programmer to avoid such cases?

This pull request does not tackle that problem.

> > enumWidener("net/minecraft/SomeEnum")
> 
> Should every enum have to be specified in loom or do we want to do something like a new file or make it part of AW and just ignore it at runtime.

I considered using AccessWidener, but as it currently is, adding a single transformation could require the addition of multiple (up to 4?) `ClassAccess` constants. Besides that, EnumWidener has to do additional transformations to fix constructor signatures. Since it exists only during compilation, I think that an entry in the build script for each enum should suffice.

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassNode.java
Chocohead: What was wrong with just returning a `MethodVisitor` which added two to `visitParameterAnnotation`'s `parameter`s when the method is `<init>`?

modmuss50 (commented):
As for the loom side of it, I do see the use of having a split compile/runtime jar but it needs to be a bit more inteligent than this. (Only using 2 jars when required, and only applying processors when needed not twice). - Should be seprate PR IMO, when I have some more time I can possibly write down my thoughts on this, or even do the work.

As for the enum widener (needs another name) bit of it, I like the idea; but another possible apporch would be to have a data driven way of adding enum entries (would remove the need of the split jars) Most of the time enum values are constructed via basic to no inputs (strings, ints and what not) these could be quite easily defined in a custom format. Failing that it could be pointed to a static method in another class.

```
net/minecraft/ExampleTargetEnum
   VALUE
      1
      "A string"
      my/mod/EnumUtils.aStaticMethod
```

This is just what ive come up with now as a basic example of what this could look like. The int, string and method call would be passed into the enums constructor. This would get remapped to intermediary on build.

As with access widners I dont think the logic should go into loom but into another lib but thats something thats not an issue at this stage.

The final question is, does this fit into fabric by default? Im not 100% sure it does, this wil need discussion and evluation of what others thing. I am happy to look into expanding the loom API to allow other plugins to have greater control over how things work, however I imagine loader's (lack of) API around this area will be the biggest pain point.

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassNode.java
nnym: > What was wrong with just returning a `MethodVisitor` which added two to `visitParameterAnnotation`'s `parameter`s when the method is `<init>`?

Nothing; I had not thought that it would be possible. It should be better now.

src/main/java/net/fabricmc/loom/processors/EnumWidenerJarProcessor.java
nnym: Should I also do that? Is something wrong with `hashCode`?

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerClassVisitor.java
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	@Override
+	public MethodVisitor visitMethod(int access, String name, String descriptor, String signature, String[] exceptions) {
+		return new EnumWidenerMethodVisitor(
+			Constants.ASM_VERSION,
+			super.visitMethod(access, name, descriptor, name.equals("<init>") ? signature.replace("(", "(Ljava/lang/String;I") : signature, exceptions),
+			name.equals("<init>")
+		);
+	}
```
Chocohead: You might as well only use the `EnumWidenerMethodVisitor` for `<init>` as there's optimisations within `ClassReader` to pass straight to a `MethodWriter` (i.e. if there's no intermediary `MethodVisitor`s)

src/main/java/net/fabricmc/loom/util/enumwidener/EnumWidenerTransformerEntry.java
```diff
@@ -0,0 +1,57 @@
...
+	private final String klass;
+
+	public EnumWidenerTransformerEntry(Project project, String klass) {
+		this.project = project;
+		this.klass = klass;
+	}
+
+	@Override
+	protected byte[] transform(ZipEntry zipEntry, byte[] input) {
+		ClassWriter writer = new ClassWriter(0);
```
Chocohead: If you keep the `ClassReader` then pass it to this as a constructor argument it can reuse the symbol table rather than have to remake it

.editorconfig
```diff
@@ -1,2 +1,4 @@
 [*.{gradle,java}]
 indent_style = tab
+ij_continuation_indent_size = 8
+ij_java_imports_layout = $*,|,java.**,|,*,|,net.fabricmc.**
```
modmuss50: Going to steel this and put it in on its own. Thanks.

TheCommunist (approved):

Mysterious-Dev (approved):

Player3324:
Sorry for the late reply, I don't usually work on Loom..

IMO we should have enum extension support, but I'd rather leave the enums intact and use a technique similar to what modmuss described instead of breaking them open. This reduces the risk of java and jit compiler problems (including later java versions) since it can generate the exact same bytecode, gives better guarantees about the usual enum invariants by keeping them sealed and doesn't expose enum implementation details to mods where changes in Javac's compilation strategy would introduce breakage that a simple loader update could fix otherwise.

The access widener approach sounds quite suitable, they have the same challenges with making java/ides and the runtime happy. Maybe it should even end up in a common library with AW. Nothing should btw rely on loader's current "make everything public" hack.

shartte:
How does this interact with serializing enums between server and client? I.e. using the read/write enum methods on packet buffers?

Wouldn't this mean that server/client need the same enum mods and also in the same order?

Ueaj-Kerman:
I think you could circumvent the order question (eg. by passing an identifier, and sorting them, or simply sorting by name), but you'd still need server/client. Though that begs the question, isn't that always the case? Like if the server adds an enum that the client doesn't have, and it's being networked, well then it will surely fail on the client right?

PR #306, opened 2020-12-07
Environment-aware JAR processors

nnym:
This branch allows `JarProcessor` to specify its processing `Environment` (`COMPILE`, `RUNTIME` or `BOTH`). Based on that, `MinecraftProcessedProvider` determines whether to use a single JAR or to split it. In the latter case, common `JarProcessor`s are applied to a single JAR before it is split; afterward, environment-specific `JarProcessor`s are applied to each.

It is based on some of the modifications done in #305, without the EnumWidener-specific changes.

Also, it adds a few EditorConfig rules for IDEA that fix formatting to work with the Checkstyle configuration. I didn't know if I should make it a separate pull request, but I am willing to do so.


.editorconfig
```diff
@@ -1,2 +1,4 @@
 [*.{gradle,java}]
 indent_style = tab
+ij_continuation_indent_size = 8
+ij_java_imports_layout = $*,|,java.**,|,*,|,net.fabricmc.**
```
modmuss50: Nice! I didnt know this was possible. 👍

src/main/java/net/fabricmc/loom/processors/MinecraftProcessedProvider.java
```diff
@@ -50,23 +57,54 @@ public MinecraftProcessedProvider(Project project, JarProcessorManager jarProces
...
-		getProject().getDependencies().add(Constants.Configurations.MINECRAFT_NAMED,
-				getProject().getDependencies().module("net.minecraft:minecraft:" + getJarVersionString(PROJECT_MAPPED_CLASSIFIER)));
+		String compileClassifier = this.split ? PROJECT_MAPPED_COMPILE_CLASSIFIER : PROJECT_MAPPED_CLASSIFIER;
+		String runtimeClassifier = this.split ? PROJECT_MAPPED_RUNTIME_CLASSIFIER : PROJECT_MAPPED_CLASSIFIER;
+
+		getProject().getDependencies().add(Constants.Configurations.MINECRAFT_NAMED_COMPILE,
+				getProject().getDependencies().module("net.minecraft:minecraft:" + getJarVersionString(compileClassifier))
+		);
+		getProject().getDependencies().add(Constants.Configurations.MINECRAFT_NAMED_RUNTIME,
+				getProject().getDependencies().module("net.minecraft:minecraft:" + getJarVersionString(runtimeClassifier))
```
modmuss50: As I said in the other PR it should only use the split jars when required.

modmuss50 (changes requested):

src/main/java/net/fabricmc/loom/processors/MinecraftProcessedProvider.java
nnym: That is what it does. If there is not a need to split the JARs, then `compileClassifier.equals(runtimeClassifier)` and this adds the single JAR to compilation and runtime class paths.

modmuss50: Ah ok, I see. as long as IDE's dont show it as 2 entries (duplicated search for example) then this should be fine.

nnym:
Being based on #305, the commit history is somewhat of a mess. Should I make a new branch with a clean history or leave it like this?

modmuss50:
I dont think thats an issue, it will get squished into 1 commit when merged.

.editorconfig
Juuxel: It's a pretty nice IDEA-specific editorconfig feature. (Could be added to Fabric API as well since its import layout quite far from IDEA's default)

src/main/java/net/fabricmc/loom/util/mixin/AnnotationProcessorInvoker.java
```diff
@@ -94,7 +94,7 @@ public void configureMixin() {
 				project.getLogger().info("Adding mixin to classpath of AP config: " + processorConfig.getName());
 				// Pass named MC classpath to mixin AP classpath
 				processorConfig.extendsFrom(
-								configs.getByName(Constants.Configurations.MINECRAFT_NAMED),
+								configs.getByName(Constants.Configurations.MINECRAFT_NAMED_RUNTIME),
```
nnym: Should this be changed to `MINECRAFT_NAMED_COMPILE` or left as it currently is?

src/main/java/net/fabricmc/loom/processors/MinecraftProcessedProvider.java
nnym: That is a problem. I'll have to hide the runtime JAR somehow. :thinking:

nnym:
Currently every Minecraft class is duplicated whenever an environment-specific processor is present. After some thought, my idea is
- to write changes common to both environments to the single JAR and,
- if environment-specific processors are present, to make additional compilation and runtime JARs, removing only the transformed classes from the single JAR and storing their original versions and their modified versions in separate JARs.

Doing so should allow both versions to be seen and debugged without causing every Minecraft class to be duplicated.

Granted, this would require a more advanced JAR processing API that allows environment-specific processors to transform classes individually and then specify whether they should be rewritten or not.

nnym:
> Currently every Minecraft class is duplicated whenever an environment-specific processor is present. After some thought, my idea is
> 
>     * to write changes common to both environments to the single JAR and,
> 
>     * if environment-specific processors are present, to make additional compilation and runtime JARs, removing only the transformed classes from the single JAR and storing their original versions and their modified versions in separate JARs.
> 
> 
> Doing so should allow both versions to be seen and debugged without causing every Minecraft class to be duplicated.
> 
> Granted, this would require a more advanced JAR processing API that allows environment-specific processors to transform classes individually and then specify whether they should be rewritten or not.

@modmuss50, is this solution acceptable? I don't know whether I should start work on it or not.

Issue #308, opened 2020-12-15
Cannot run genSources in subproject if root project does not have the fabric-loom plugin


solonovamax:
When running `genSources` in a subproject, you get the following error:
```
Error: Could not find or load main class net.fabricmc.loom.decompilers.fernflower.FabricForkedFFExecutor
Caused by: java.lang.ClassNotFoundException: net.fabricmc.loom.decompilers.fernflower.FabricForkedFFExecutor
```
This seems to be caused by the commit db1fe88, where for *some reason*, it was decided that it would be a good idea to rely on the root project having the `fabric-loom` gradle plugin.
If you'd like to reproduce this error, you can clone this repo [here](https://github.com/solonovamax/Terra/tree/agnostic) (make sure to check out the right branch), or try and recreate it in a clean environment ig.

If you want me to provide any other stack traces or whatever, feel free to ask.

Tagging #181, as that was the issue that prompted this commit.
(There were no other open issues for this, so this is not a duplicate.)

i509VCB:
~~Would running `:fabric:genSources` work? I do have some experience with this from the luckperms PR I have made~~

Edit: I misread the message lol, you said you ran it in subproject

Ideally this could be fixed by checking if loom is applied to root project and or any other projects. If loom is only applied to one project, then that should run fine with a small modification, but if it's two sub projects and not root then modmuss would need to figure out whats appropriate there.

Tigermouthbear:
@solonovamax can you make a pr with your fix? I'm getting the same error

modmuss50:
The fix was merged in the (currenly unstable) 0.6 branch: https://github.com/FabricMC/fabric-loom/commit/c96396fb2732e76ee460be622b8ba928ee95f24e

Issue #181, opened 2020-02-25
Loom Gradle task fails when not ran from root configuration


Martmists-GH:
[Log of `$ gradle :versions:1.16:gensources --debug`](https://transfer.sh/KDoZq/log.txt)

modmuss50:
Fixed: https://github.com/FabricMC/fabric-loom/commit/db1fe88fcc08538d5211be718ebe607d0488819f

PR #309, opened 2020-12-16
Fix issue for issue #308 (genSources cannot run on subproject if root doesn't have loom plugin.)

solonovamax:
loom now runs genSources on root project if the plugin exists in the root project, if it doesn't then it will build in the current project (the subproject)

Issue #181 also shouldn't appear again, although I haven't tested it for that. (I did test it on my project which was causing issues, so it probably™ works)


solonovamax:
????

I just tested it myself on gradle 4.10.2 with java 8 and it worked fine???

bruh what.

[here](https://hastebin.com/ihuyekomow.sql) is a hastebin of it successfully running the failed test. (it was `SimpleBuildFunctionalTest` that failed, so I used the command `./gradlew test --info --tests *SimpleBuildFunctionalTest` to run it, which only runs that test.)

I'm also like 99% sure that it didn't even run anything in the class that I modified... I don't see the `genSources` task anywhere in those logs, so I have absolutely no clue why this failed...

Edit: modmuss reran the tests and they worked fine.

gg

modmuss50 (approved):

PR #316, opened 2020-12-23
Inject stitch annotations as a Minecraft dependency

zml2008:
[Stitch](https://github.com/FabricMC/stitch/pull/36) | [Loader](https://github.com/FabricMC/fabric-loader/pull/351) | **Loom**

This is a (technically optional) companion to the Stitch PR to split environment annotations, to inject the stitch annotations as a compile dependency the same way jetbrains annotations are.

While loader could bring these in in most cases, having them injected makes the setup more fool-proof.

As with loader, this PR depends on the one in Stitch and won't build without a published version of stitch with changes applied.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
 	implementation ('com.google.guava:guava:28.0-jre')
 
 	// game handling utils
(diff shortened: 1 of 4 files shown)


modmuss50:
Stitch is not right for this, see loader PR for more info.

PR #318, opened 2020-12-25
Allow for customization of run config generation

FoxSamu:
It kinda bugged me that I wasn't able to properly add generated run configurations. We have some special run type in our project which runs a dedicated server to test whether it loads properly (and does not reference any client code it should not do) and right now we need a complicated setup to actually run that server. This PR should add the ability to configure the generated run configurations and `run<insert run config here>` tasks using a more clear setup

~~I haven't tested, idk how to test gradle plugins properly - either confirm that it works or tell me how to test this properly~~  
Edit: I did now

Run configurations look like this:
```groovy
minecraft {
    runs {
        client {
            // Client configs go here
        }
        server {
            // Server configs go here
        }

        anotherConfig {
            // Some other config goes here
            client true // Add correct OS params for client
        }
    }
}
```
By default `client` and `server` run configs are defined,
whether they are explicitly configured or not. Above should
generate these tasks:
- `runClient`
- `runServer`
- `runAnotherConfig`
And the above should generate three run configs:
- 'Minecraft Client'
- 'Minecraft Server'
- 'Minecraft AnotherConfig'

In addition: I updated GSON to 2.8.6 because 2.8.5 downloaded an empty artifact for some reason. If this should very necessarily changed back I'll do so.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ dependencies {
 	// libraries
 	implementation ('commons-io:commons-io:2.6')
 	implementation ('org.zeroturnaround:zt-zip:1.13')
-	implementation ('com.google.code.gson:gson:2.8.5')
+	implementation ('com.google.code.gson:gson:2.8.6')
 	implementation ('com.google.guava:guava:28.0-jre')
 
 	// game handling utils
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -24,31 +24,8 @@
 
 package net.fabricmc.loom;
 
-import java.io.File;
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
-import java.util.function.BiPredicate;
-import java.util.function.Function;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
-
-import javax.annotation.Nullable;
-
 import com.google.gson.JsonObject;
-import org.cadixdev.lorenz.MappingSet;
-import org.cadixdev.mercury.Mercury;
-import org.gradle.api.Project;
-import org.gradle.api.artifacts.Configuration;
-import org.gradle.api.artifacts.Dependency;
-import org.gradle.api.file.ConfigurableFileCollection;
-import org.gradle.api.plugins.BasePluginConvention;
-
+import groovy.lang.Closure;
 import net.fabricmc.loom.api.decompilers.LoomDecompiler;
 import net.fabricmc.loom.processors.JarProcessor;
 import net.fabricmc.loom.processors.JarProcessorManager;
@@ -57,6 +34,24 @@ import net.fabricmc.loom.providers.MinecraftMappedProvider;
 import net.fabricmc.loom.providers.MinecraftProvider;
 import net.fabricmc.loom.util.LoomDependencyManager;
 import net.fabricmc.loom.util.mappings.MojangMappingsDependency;
+import org.cadixdev.lorenz.MappingSet;
+import org.cadixdev.mercury.Mercury;
+import org.gradle.api.NamedDomainObjectContainer;
+import org.gradle.api.NamedDomainObjectFactory;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.file.ConfigurableFileCollection;
+import org.gradle.api.plugins.BasePluginConvention;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 public class LoomGradleExtension {
 	public String runDir = "run";
@@ -83,20 +78,21 @@ public class LoomGradleExtension {
 	private Mercury[] srcMercuryCache = new Mercury[2];
 	private Set<File> mixinMappings = Collections.synchronizedSet(new HashSet<>());
 
+	private NamedDomainObjectContainer<RunConfigSettings> runs;
+
 	/**
-	 * Loom will generate a new genSources task (with a new name, based off of {@link LoomDecompiler#name()})
-	 * that uses the specified decompiler instead.
+	 * Loom will generate a new genSources task (with a new name, based off of {@link LoomDecompiler#name()}) that uses
+	 * the specified decompiler instead.
 	 */
 	public void addDecompiler(LoomDecompiler decompiler) {
 		decompilers.add(decompiler);
 	}
 
 	/**
-	 * Add a transformation over the mapped mc jar.
-	 * Adding any jar processor will cause mapped mc jars to be stored per-project so that
-	 * different transformation can be applied in different projects.
-	 * This means remapping will need to be done individually per-project, which is slower when developing
-	 * more than one project using the same minecraft version.
+	 * Add a transformation over the mapped mc jar. Adding any jar processor will cause mapped mc jars to be stored
+	 * per-project so that different transformation can be applied in different projects. This means remapping will need
+	 * to be done individually per-project, which is slower when developing more than one project using the same
+	 * minecraft version.
 	 */
 	public void addJarProcessor(JarProcessor processor) {
 		jarProcessors.add(processor);
@@ -118,6 +114,7 @@ public class LoomGradleExtension {
 		this.project = project;
 		this.autoGenIDERuns = AbstractPlugin.isRootProject(project);
 		this.unmappedMods = project.files();
+		this.runs = project.container(RunConfigSettings.class, RunConfigSettings::new);
 	}
 
 	/**
@@ -135,8 +132,8 @@ public class LoomGradleExtension {
 	@Deprecated
 	public List<Path> getUnmappedMods() {
 		return unmappedMods.getFiles().stream()
-			.map(File::toPath)
-			.collect(Collectors.toList());
+						   .map(File::toPath)
+						   .collect(Collectors.toList());
 	}
 
 	public ConfigurableFileCollection getUnmappedModCollection() {
@@ -424,5 +421,113 @@ public class LoomGradleExtension {
 
 	public Set<File> getAllMixinMappings() {
 		return Collections.unmodifiableSet(mixinMappings);
+	}
+
+	public void runs(Closure<?> conf) {
+		runs.configure(conf);
+	}
+
+	public NamedDomainObjectContainer<RunConfigSettings> getRuns() {
+		return runs;
+	}
+
+	public static class RunConfigSettings {
+		private final List<String> vmArgs = new ArrayList<>();
+		private final List<String> programArgs = new ArrayList<>();
+		private String mode;
+		private String name;
+		private Boolean client;
+		private final String baseName;
+
+		public RunConfigSettings(String baseName) {
+			this.baseName = baseName;
+			setMode(baseName);
+		}
+
+		public String getBaseName() {
+			return baseName;
+		}
+
+		public List<String> getVmArgs() {
+			return vmArgs;
+		}
+
+		public List<String> getProgramArgs() {
+			return programArgs;
+		}
+
+		public String getMode() {
+			return mode;
+		}
+
+		public void setMode(String mode) {
+			this.mode = mode;
+		}
+
+		public String getName() {
+			return name;
+		}
+
+		public void setName(String name) {
+			this.name = name;
+		}
+
+		public boolean isClient() {
+			String m = mode != null ? mode : baseName;
+			return client != null ? client // Do not confuse users: detect client mode unless client mode is explicitly defined
+								  : m.toLowerCase().contains("client");
+		}
+
+		public void setClient(Boolean client) {
+			this.client = client;
+		}
+
+		public void mode(String mode) {
+			setMode(mode);
+		}
+
+		public void name(String name) {
+			setName(name);
+		}
+
+		public void client(boolean client) {
+			setClient(client);
+		}
+
+		public void server(boolean server) {
+			setClient(!server);
+		}
+
+		public void vmArg(String arg) {
+			vmArgs.add(arg);
+		}
+
+		public void vmArgs(String... args) {
+			vmArgs.addAll(Arrays.asList(args));
+		}
+
+		public void vmArgs(Collection<String> args) {
+			vmArgs.addAll(args);
+		}
+
+		public void property(String name, String value) {
+			vmArg("-D" + name + "=" + value);
+		}
+
+		public void properties(Map<String, String> props) {
+			props.forEach(this::property);
+		}
+
+		public void programArg(String arg) {
+			programArgs.add(arg);
+		}
+
+		public void programArgs(String... args) {
+			programArgs.addAll(Arrays.asList(args));
+		}
+
+		public void programArgs(Collection<String> args) {
+			programArgs.addAll(args);
+		}
 	}
 }
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -536,6 +536,10 @@ public class LoomGradleExtension {
 			vmArg("-D" + name + "=" + value);
 		}
 
+		public void property(String name) {
+			vmArg("-D" + name);
+		}
+
 		public void properties(Map<String, String> props) {
 			props.forEach(this::property);
 		}
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -27,10 +27,12 @@ package net.fabricmc.loom;
 import java.io.File;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiPredicate;
@@ -39,8 +41,11 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import com.google.gson.JsonObject;
+import groovy.lang.Closure;
 import org.cadixdev.lorenz.MappingSet;
 import org.cadixdev.mercury.Mercury;
+import org.gradle.api.Named;
+import org.gradle.api.NamedDomainObjectContainer;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.Dependency;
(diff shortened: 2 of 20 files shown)


FoxSamu:
Ok I got the testing strategy :)

i509VCB:
Could you change the PR to target `0.6-dev` since that is the current branch prs should target.

FoxSamu:
Ah yes I'll do

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -425,4 +432,113 @@ public synchronized File getNextMixinMappings() {
 	public Set<File> getAllMixinMappings() {
 		return Collections.unmodifiableSet(mixinMappings);
 	}
+
+	public void runs(Closure<?> conf) {
```
i509VCB, line 437: This should be `Action<?>` so mods using a kotlin dsl do not need to do any nasty `closureOf((T).() -> Unit)`

Also not sure if that should be a raw type here?

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -425,4 +432,113 @@ public synchronized File getNextMixinMappings() {
...
+
+		public void vmArgs(String... args) {
+			vmArgs.addAll(Arrays.asList(args));
+		}
+
+		public void vmArgs(Collection<String> args) {
+			vmArgs.addAll(args);
+		}
+
+		public void property(String name, String value) {
```
i509VCB: Would be nice if there was an overload of this which just passed the name so we can `-Dname`

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -135,8 +142,8 @@ public void addUnmappedMod(Path file) {
 	@Deprecated
 	public List<Path> getUnmappedMods() {
 		return unmappedMods.getFiles().stream()
-			.map(File::toPath)
-			.collect(Collectors.toList());
+				.map(File::toPath)
+				.collect(Collectors.toList());
```
i509VCB, line 147: unnecessary PR noise

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -425,4 +432,113 @@ public synchronized File getNextMixinMappings() {
...
+
+	public void runs(Closure<?> conf) {
+		runs.configure(conf);
+	}
+
+	public NamedDomainObjectContainer<RunConfigSettings> getRuns() {
+		return runs;
+	}
+
+	public static class RunConfigSettings implements Named {
```
i509VCB: Would it be possible to expose a way to select a specific module (sourceSet) to use in this. Main use case is a `Test Minecraft Client/Server` run config for fabric api.

i509VCB (changes requested):

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
FoxSamu: The thing here is that there is no direct way to do this with an `Action<?>`. This is just a redirect to the Gradle API that backs `NamedDomainObjectContainer`. I'll see if I can find a workaround.

FoxSamu: did it a bit differently, closure is still there but can now also say `run(name, action)` in addition to `runs { name {...} }`

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
FoxSamu: Done

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
FoxSamu: Not sure how I would do that. There isn't really anything configurable in the generated run configs when it comes to source sets... Gotta experiment with this...

FoxSamu:
Ok because I could not change the branch I PRed (only the base branch) it shows `dev/0.5` but it is updated to `dev/0.6`

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
FoxSamu: fixet it for as far as I can see

FoxSamu:
I have included `nogui` as configurable program parameter on the standard server config. Personally I prefer running with GUI so I can see the memory profile. If this idea get's used I  suggest disabling `nogui` by default but specifying it as a default configuration in the example mod.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
FoxSamu: Ok  I have a solution for this but I need to get the `Project` a `SourceSet` belongs to and that is apparently not easily done

build.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
 	// libraries
 	implementation ('commons-io:commons-io:2.6')
 	implementation ('org.zeroturnaround:zt-zip:1.13')
-	implementation ('com.google.code.gson:gson:2.8.5')
+	implementation ('com.google.code.gson:gson:2.8.6')
```
modmuss50, line 41: Ive had this being empty before, no idea what casues it, deleting the file seems to fix it.

Doing this should be fine, but I wouldnt be surprised if it breaks agian.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -428,4 +438,122 @@ public synchronized File getNextMixinMappings() {
...
+	// Workaround for ugly closures in Kotlin DSL buildscripts
+	public void run(String name, Action<RunConfigSettings> settingsAction) {
+		runs.create(name, settingsAction);
+	}
+
+	public NamedDomainObjectContainer<RunConfigSettings> getRuns() {
+		return runs;
+	}
+
+	public static class RunConfigSettings implements Named {
```
modmuss50: This class should go into its own file, this class is getting quite large these days.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -428,4 +438,122 @@ public synchronized File getNextMixinMappings() {
...
+		private final List<String> vmArgs = new ArrayList<>();
+		private final List<String> programArgs = new ArrayList<>();
+		private String mode;
+		private String name;
+		private Boolean client;
```
modmuss50: I think we should keep the doors open for more than just client or server.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -428,4 +438,122 @@ public synchronized File getNextMixinMappings() {
...
+	}
+
+	public NamedDomainObjectContainer<RunConfigSettings> getRuns() {
+		return runs;
+	}
+
+	public static class RunConfigSettings implements Named {
+		private final List<String> vmArgs = new ArrayList<>();
+		private final List<String> programArgs = new ArrayList<>();
+		private String mode;
```
modmuss50: Some basic comments might be nice on this, `mode` wasnt that clear what it does.

src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
```diff
@@ -50,11 +49,6 @@ public void apply(Project project) {
 			project.getLogger().lifecycle("Refresh dependencies is in use, loom will be significantly slower.");
 		}
 
-		// Apply default plugins
-		project.apply(ImmutableMap.of("plugin", "java"));
-		project.apply(ImmutableMap.of("plugin", "eclipse"));
-		project.apply(ImmutableMap.of("plugin", "idea"));
```
modmuss50: Where have these gone? Am I missing something?

src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
```diff
@@ -149,6 +149,7 @@ private static void populate(Project project, LoomGradleExtension extension, Run
 		}
 	}
 
+	@Deprecated // Replaced by runConfig(project, settings)
```
modmuss50: This isnt an API so feel free to remove it fully.

src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
```diff
@@ -157,29 +158,83 @@ public static RunConfig clientRunConfig(Project project) {
...
+
+		return runConfig;
+	}
+
+	// This can be removed at somepoint, its not ideal but its the best solution I could think of
+	@Deprecated // Whatever this is still going to mean, replaced it with more strict version below
 	public static boolean needsUpgrade(File file) throws IOException {
 		String contents = FileUtils.readFileToString(file, StandardCharsets.UTF_8);
 		return !(contents.contains("net.fabricmc.devlaunchinjector.Main"));
 	}
```
modmuss50: This can go now, its been long enough since this was added, it didnt really work anyway.

src/main/java/net/fabricmc/loom/task/LoomTasks.java
```diff
@@ -81,16 +81,21 @@ private static void registerIDETasks(TaskContainer tasks) {
...
-		tasks.register("runServer", RunServerTask.class, t -> {
-			t.setDescription("Starts a development version of the Minecraft server.");
-			t.setGroup("fabric");
+		project.afterEvaluate((p) -> {
+			for (LoomGradleExtension.RunConfigSettings config : extension.getRuns()) {
+				String configName = config.getName();
+				String taskName = "run" + configName.substring(0, 1).toUpperCase() + configName.substring(1);
+
+				tasks.register(taskName, RunGameTask.class, config).configure(t -> {
+					t.setDescription("Starts a development version of the Minecraft server.");
```
modmuss50: This description will be off for a few. Should prob change it to something a little more generic.

src/main/java/net/fabricmc/loom/task/RunClientTask.java
```diff
@@ -26,6 +26,7 @@
 
 import net.fabricmc.loom.configuration.ide.RunConfig;
 
+@Deprecated // Replaced by RunGameTasl
```
modmuss50: Typo.

This class should be Deprecated  👍

src/main/java/net/fabricmc/loom/task/RunServerTask.java
```diff
@@ -26,6 +26,7 @@
 
 import net.fabricmc.loom.configuration.ide.RunConfig;
 
+@Deprecated // Replaced by RunGameTasl
```
modmuss50: Again typo.

modmuss50 (changes requested):
Awesome start, this is something ive been meaning to do for a while 👍 

Just a few things that need changing. Once ready this will take a fair amount of testing.

src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
FoxSamu: whoops I removed them while resolving a merge conflict

src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
FoxSamu: Ok, `RunClientTask` and `RunServerTask` still use them but I'll redirect these tasks to `runConfig(...)`

src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
FoxSamu: Ok I'll remove the deprecated one. I'll keep `needsUpgrade(File, RunConfig)` below for now because it's still used and I'm not sure what to do with the usages. Should usages instead just assume an upgrade is always needed?

src/main/java/net/fabricmc/loom/task/LoomTasks.java
FoxSamu: Whoops that was a copy paste mistake, I'll change it

build.gradle
FoxSamu: Ok well I still had gson 2.8.6 cached so I updated it to that version. It didn't break anything so far

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
FoxSamu: Actually this only determines whether `-XstartOnFirstThread` should be applied and whether `KnotClient` should be used if the config isn't named `client` or `server` and no main class has been configured in any installer json (`fabric_installer.<x>.json` should be able to provide main classes for custom modes). But it's indeed a bit strange, so I thought to add a default main class property to the settings for when no main class is defined in any installer json, and just add the `-XstartOnFirstThread` on macos when configuring the default `client` and `server` setup

FoxSamu:
I concluded that adding program arguments is not compatible with launchwrapper. Launchwrapper uses tweaker classes to add program arguments manually, and adding program arguments to the run settings will result in launchwrapper getting these arguments instead of Minecraft itself. Not sure whether this is a problem, I suppose most modders just use Knot (which does seem to feed minecraft the program arguments directly).

src/main/java/net/fabricmc/loom/task/LoomTasks.java
FoxSamu: Ah good point, I forgot about that. Might just have been a copy paste mistake. It's fixed now

modmuss50:
I wouldnt worry too much about the LW support, I dont think it was really working too well before. As long as its not got worse then all should be fine. It can always be looked at on its own.

src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
```diff
@@ -149,35 +149,92 @@ private static void populate(Project project, LoomGradleExtension extension, Run
...
+
+		if (defaultMain == null) {
+			if (name.equals("client")) {
+				defaultMain = Constants.Knot.KNOT_CLIENT;
+			} else if (name.equals("server")) {
+				defaultMain = Constants.Knot.KNOT_SERVER;
+			} else {
+				// throw new IllegalArgumentException("Run configuration '" + name + "' must specify 'defaultMainClass'");
+
+				// Can go for above exception, or just assume user wants to start client (because that one is used most often)
```
FoxSamu, line 179: What the comment says: if no default main class is specified it can throw an exception (build fail) or just fall back on `KnotClient` (assuming client is most often run). What is better?

FoxSamu:
Resolved requested changes as much as I could. In addition I added two more things:
- Source sets can be configured in a run configuration now. This does not make sense on VSCode or Eclipse, but IntelliJ seems to care about them. Requested by @i509VCB.
- The run directory can now be properly configured for each run configuration. The run directory could be configured but it didn't do anything except in IntelliJ as far as I understood the code - I should have now made it such that it generates configs with the right run directory.

modmuss50 (approved):
Looks good 👍 

Thanks a lot for this, I will merge once tested.

modmuss50 (changes requested):
Hit a number of issues when trying to run this:

`RunConfig:224` triggers a NPE, looks like an easy fix by initialising `programArgs` and `vmArgs`.

After fixing that I get the following:

`The constructor for type RunGameTask should be annotated with @Inject.`

Both happens when running the `runClient` task.

Juuxel:
> `The constructor for type RunGameTask should be annotated with @Inject.`

Instead of that I'd do it like Gradle tasks usually do, and put the config as a field/property in the task class.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
Juuxel: You could also have `runs(Action<NamedDomainObjectContainer<RunConfigSettings>>)` to mirror Gradle's own object containers, which can be configured with `Action`s like that.

For example, `publications` from maven-publish supports it like this:

```kotlin
publishing {
    publications {
        create<MavenPublication>("first") { ... }
        create<MavenPublication>("second") { ... }
    }
}
```

FoxSamu: I thought about that but thought that wouldn't work but apparently it does

FoxSamu:
Thing is that `RunGameTask` extends `AbstractRunTask` which requires a `Function<Project, RunConfig>`. I must give that in the `RunGameTask` constructor so adding it as a property doesn't work in this case. But since `RunGameTask` is the only non-deprecated implementor of `AbstractRunTask` I can also just copy the logic of `AbstractRunTask` into `RunGameTask` and make `RunGameTask` not extend `AbstractRunTask` - I'll do that for the sake of custom `RunGameTask` usages in build scripts.

src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	/**
+	 * Removes the {@code nogui} argument for the server configuration. By default {@code nogui} is specified, this is
+	 * a convenient way to remove it if wanted.
+	 */
+	public void serverWithGui() {
+		while (programArgs.remove("nogui")) {
+			// N/A, the condition does the work
+		}
```
liach:
```suggestion
		programArgs.removeIf("nogui"::equals);
```
Removes all occurrences in one line

src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
```diff
@@ -0,0 +1,279 @@
...
+	public void name(String name) {
+		setConfigName(name);
+	}
+
+	public void defaultMainClass(String cls) {
+		setDefaultMainClass(cls);
+	}
+
+	public void runDir(String dir) {
+		setRunDir(dir);
```
liach, line 186: No point of such `set` overloads. Dunno how it looks in kotlin but in groovy it is difference between `runDir "run"` (this one) and `runDir = "run"` (setter), and I don't think it's worth creating an overload for an `=` sign

liach (commented):
I think that such run tasks can always be abstracted to a `JavaExec` from gradle. If possible I'd recommend letting loom create a javaexec from this extension/run spec.

src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
i509VCB: just `get/set` should suffice, kotlin will generate a `var` accessor in compiler and groovy will treat it like a property.

modmuss50:
@ShadewEnder do you plan on contiuning with this? or is it something I should pick up as its almost there?

FoxSamu:
Oh uh yes, I kind of paused this as I had to do university again (last weeks were a bit busy). I didn't forget it, although if you wanna continue this it's also fine to me.

modmuss50:
Ok, no worries. Education clearly comes first 👍 

Its 95% of the way there, so Ill finish it off. Thanks.

FoxSamu:
Ok perfect! Any ETA's for 0.6 already or should I just be patient?

modmuss50:
Created another PR based off this, <https://github.com/FabricMC/fabric-loom/pull/338> (I didnt seem to be able to push to this branch). Should be good to merge. I have made the API experimental for now so if things need to change they can.

modmuss50:
Pulled in via my other PR, thanks a lot for this 👍

FoxSamu:
you're welcome :)

PR #320, opened 2020-12-28
Jar Processor Stages

TheGlitch76:
Draft because this hasn't been extensively tested. I have confirmed that normal ("mapped") stage jar processors work as before, but the rest is mostly untested.
Also, the code is a bit of a mess.

Jar Processor stages allow a jar processor to pick if it wants to process an obf, intermediary, or mapped jar.

Justifications for adding the new stages:
- The **obf** stage is provided for things like using different modloaders over Loom. In this instance a jar processor can simply run the modloader's installer, revert the installer's output to obf, and then not have to duplicate any of looms' remapping work.
- The **intermediary** stage is for generic transformations (perhaps statically applied ASM transformations). It should be obvious why this is important--an author can write for the intermediary without having to do all the remapping logic themselves, or enforce a specific Yarn version.
- The **mapped** stage is what we already had.


TheGlitch76:
This seems to work fine but I still want to do more testing before a merge.
Reviews are welcome

PR #332, opened 2021-01-16
Gradle 7 support

modmuss50:
Gradle 7 is nearing release candidate with some major breaking changes.

I was bored, so decided to have a quite crack at fixing loom.

This clearly might break again before the final release. (Used the latest nightly).

When running on gradle 7 or newer `modCompile` will not added.

_Also fixes gradle 8 deprecation warnings_

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/RemappedConfigurationEntry.java
+++ b/src/main/java/net/fabricmc/loom/configuration/RemappedConfigurationEntry.java
@@ -25,6 +25,9 @@
 package net.fabricmc.loom.configuration;
 
 import org.gradle.api.artifacts.ConfigurationContainer;
+import org.gradle.api.plugins.JavaPlugin;
+
+import net.fabricmc.loom.util.gradle.GradleSupport;
 
 public class RemappedConfigurationEntry {
 	private final String sourceConfiguration;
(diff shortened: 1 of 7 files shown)


src/main/java/net/fabricmc/loom/configuration/RemappedConfigurationEntry.java
```diff
@@ -61,7 +64,7 @@ public String getRemappedConfiguration() {
 
 	public String getTargetConfiguration(ConfigurationContainer container) {
 		if (container.findByName(targetConfiguration) == null) {
-			return "compile";
+			return GradleSupport.IS_GRADLE_7_OR_NEWER ? JavaPlugin.IMPLEMENTATION_CONFIGURATION_NAME : "compile";
```
i509VCB: Here's a new one for the `Constants` file on the other end of the tenary operator

modmuss50: This code here is a little weird. I dont really like it overall. What it does is if the java-library plugin isnt present it will use `compile` or `impl` instead of `api` when doing `modApi`.

It might not be an awful idea to fix this. At the same time?

PR #333, opened 2021-01-22
Set classpath before compileJava task (#330)

### LINKED ISSUE #330: [0.6] runClient is completely broken
It appears that Loom 0.6 never recompiles mod classes with `runClient`. While this works fine with the IDEA run configs, it's a big problem for people using `runClient`.

Basic log inspection seems to show that on Loom 0.6, `runClient` doesn't ever cause `:compileJava` to run. However, doing `./gradlew compileJava runClient` still doesn't fix the issue.

Steps to reproduce:

1. Clone https://github.com/fabricmc/fabric-example-mod
2. Switch the Loom version to '0.6-SNAPSHOT' in build.gradle, **before running any gradle tasks**
3. Run `./gradlew runClient`
4. The fabric example mod is never loaded
5. Switch back to '0.5-SNAPSHOT'
6. Run `./gradlew runClient`
7. Note that the example mod is properly loaded
8. Switch back to 0.6-SNAPSHOT
9. Insert the following code into ExampleMod.java

```java
package net.fabricmc.example;

import net.fabricmc.api.ModInitializer;

public class ExampleMod implements ModInitializer {
	@Override
	public void onInitialize() {
		// This code runs as soon as Minecraft is in a mod-load-ready state.
		// However, some things (like resources) may still be uninitialized.
		// Proceed with mild caution.

		System.out.println("Hello Fabric world!");
		
		// This should make the game immediately exit
		System.exit(0);
	}
}
```

10. Launch the game
11. Note that the game launches successfully - this indicates that the added code is never loaded!
12. Switch back to '0.5-SNAPSHOT'
13. Launch the game
14. Note that the game abruptly exits, as expected

bytzo:
As of Loom 0.6.13, the `runClient` and `runServer` tasks do not compile or load the active mod classes. This is because the classpath of the project is set after the `compileJava` task, causing the mod classes to never be compiled. To fix this, I have moved setting the classpath to before the `compileJava` task, as done with versions 0.6.12 and earlier (see the change in 0.6.13 [here](https://github.com/FabricMC/fabric-loom/commit/792a64e2efecb850e4752a298a4e3d6f50f55a64#diff-128362915f1f370814294926952437205a6d7b36e7d08aa44bc55d727e782019)). This fixes #330.

Note that I have relatively little Java experience, so this fix could be wrong. However, it seems to have worked in my projects so far without issues.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
+++ b/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
@@ -45,6 +45,9 @@ public abstract class AbstractRunTask extends JavaExec {
 		super();
 		setGroup("fabric");
 		this.configProvider = config;
+
+		classpath(getProject().getConfigurations().getByName("runtimeClasspath"));
+		classpath(this.getProject().getExtensions().getByType(LoomGradleExtension.class).getUnmappedModCollection());
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


modmuss50:
Looks correct, I think I did this to try and save a few milisconds. Clearly didnt test it enough.

Thanks 👍

PR #335, opened 2021-01-23
Cache the checksum of asset index files.

shedaniel:
Since IDEA sync will "download" the assets again, I think it is good to make this faster. Without caching, this operation takes a second, this will take it down to an average of 10ms.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/assets/MinecraftAssetsProvider.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/assets/MinecraftAssetsProvider.java
@@ -46,6 +46,7 @@ import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 
 import net.fabricmc.loom.LoomGradleExtension;
+import net.fabricmc.loom.LoomGradlePlugin;
 import net.fabricmc.loom.configuration.providers.MinecraftProvider;
 import net.fabricmc.loom.configuration.providers.minecraft.MinecraftVersionInfo;
 import net.fabricmc.loom.util.Checksum;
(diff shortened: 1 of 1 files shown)


pluiedev (approved):
epic :intellithonk:

Juuxel (approved):

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/assets/MinecraftAssetsProvider.java
```diff
@@ -99,14 +117,24 @@ public static void provide(MinecraftProvider minecraftProvider, Project project)
...
+			String localFileChecksum = !file.exists() ? null : checksumInfos.computeIfAbsent(entry.getKey(), path -> {
+				try {
+					return Files.asByteSource(file).hash(Hashing.sha1()).toString();
+				} catch (IOException e) {
+					e.printStackTrace();
+					return null;
+				}
+			});
+
+			if (localFileChecksum == null || !localFileChecksum.equals(sha1)) {
```
modmuss50: This should respect respect --refresh-dependencies

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/assets/MinecraftAssetsProvider.java
```diff
@@ -135,13 +163,33 @@ public static void provide(MinecraftProvider minecraftProvider, Project project)
 							throw new RuntimeException("Failed to download: " + assetName, e);
 						}
 
+						try {
+							if (localFileChecksum == null) {
+								checksumInfos.put(entry.getKey(), Files.asByteSource(file).hash(Hashing.sha1()).toString());
```
modmuss50: You should be able to pull the hash from the AssetObject. Might need adding to that record.

modmuss50 (changes requested):
Looks good, nice spot on the performace issue. Just a few minor changes are needed.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/assets/MinecraftAssetsProvider.java
shedaniel: I don't think previously the asset downloading respect refresh dependencies, but I have added that now.

modmuss50 (approved):

PR #337, opened 2021-01-24
Use the given manifest version hash if possible

### LINKED ISSUE #334: Loom should use SHA1 hash for entrypoints
Currently, Loom decides to calculate a hash based on the URL which is very prone to breaking.
Quoting chocohead: `This is a sketchy check which is where it's getting the hash from by the way: https://github.com/FabricMC/fabric-loom/blob/dev/0.6/src/main/java/net/fabricmc/loom/configuration/providers/MinecraftProvider.java#L173:L179` and that the issue could be removed entirely with a much better SHA1 hash check from the manifest. This only affects custom manifests, but I think it should be easily fixable.

Chocohead:
Fixes #334


PhoenixVX:
👍

liach (approved):

modmuss50 (approved):
Yes, IDK what I was thinking when I wrote this.

PR #338, opened 2021-01-26
Improve run configs superseeds #318

modmuss50:
Basicly just #318 with a few final fixes.

I have made this new API experimental as I think its quite likely holes will be found (Mainly looking that the kotlin build script fans here). The aim will be to drop this marking during 0.6 once people have been able to have a play with it.

This is the basic example script I have been working, with. Tested via the CLI and in vscode/idea

```gradle
loom {
	runs {
		exampleConfig {
			client() // Adds the default client options

			vmArg "-XX:+ShowCodeDetailsInExceptionMessages"
			property "fabric.log.level", "debug"
		}
	}
}
```

As far as I can tell this is good to merge, and as its experimental we can break it as desired later down the line.

Many many thanks to @ShadewEnder for all of this its been on the loom todo list for a long time.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -24,31 +24,8 @@
 
 package net.fabricmc.loom;
 
-import java.io.File;
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
-import java.util.function.BiPredicate;
-import java.util.function.Function;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
-
-import javax.annotation.Nullable;
-
 import com.google.gson.JsonObject;
-import org.cadixdev.lorenz.MappingSet;
-import org.cadixdev.mercury.Mercury;
-import org.gradle.api.Project;
-import org.gradle.api.artifacts.Configuration;
-import org.gradle.api.artifacts.Dependency;
-import org.gradle.api.file.ConfigurableFileCollection;
-import org.gradle.api.plugins.BasePluginConvention;
-
+import groovy.lang.Closure;
 import net.fabricmc.loom.api.decompilers.LoomDecompiler;
 import net.fabricmc.loom.processors.JarProcessor;
 import net.fabricmc.loom.processors.JarProcessorManager;
@@ -57,6 +34,24 @@ import net.fabricmc.loom.providers.MinecraftMappedProvider;
 import net.fabricmc.loom.providers.MinecraftProvider;
 import net.fabricmc.loom.util.LoomDependencyManager;
 import net.fabricmc.loom.util.mappings.MojangMappingsDependency;
+import org.cadixdev.lorenz.MappingSet;
+import org.cadixdev.mercury.Mercury;
+import org.gradle.api.NamedDomainObjectContainer;
+import org.gradle.api.NamedDomainObjectFactory;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.file.ConfigurableFileCollection;
+import org.gradle.api.plugins.BasePluginConvention;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 public class LoomGradleExtension {
 	public String runDir = "run";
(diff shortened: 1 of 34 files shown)


src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -426,4 +432,20 @@ public synchronized File getNextMixinMappings() {
 	public List<LoomDecompiler> getDecompilers() {
 		return decompilers;
 	}
+
+	@ApiStatus.Experimental
+	public void runs(Closure<?> conf) {
```
Juuxel: I'd just use an `Action<NamedDomainObjectContainer<RunConfigSettings>>` here too. That should work nicely in both kt and groovy dsl.

src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
```diff
@@ -0,0 +1,283 @@
...
+		this.runDir = runDir;
+	}
+
+	public SourceSet getSource(Project proj) {
+		return source.apply(proj);
+	}
+
+	public void setSource(SourceSet source) {
+		this.source =
+				proj -> source;
```
Juuxel: Why is this on its own line?

Juuxel (changes requested):
Looks mostly good, just some small comments.

(Also, I might PR environmental variables at some point. I have them (and need them) in my own fork but no Eclipse support yet.)

src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
modmuss50: Checkstyle weirdness, updating checkstyle seems to have solved it.

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
modmuss50: Seems to work fine in groovy 👍

src/main/java/net/fabricmc/loom/task/RunClientTask.java
```diff
@@ -24,10 +24,17 @@
...
 
+import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.configuration.ide.RunConfig;
 
+@Deprecated // Replaced by RunGameTask
 public class RunClientTask extends AbstractRunTask {
 	public RunClientTask() {
-		super(RunConfig::clientRunConfig);
+		super(
+				project -> {
```
i509VCB: What the fuck is this mess lmao

src/main/java/net/fabricmc/loom/task/RunServerTask.java
```diff
@@ -24,10 +24,17 @@
...
 
+import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.configuration.ide.RunConfig;
 
+@Deprecated // Replaced by RunGameTask
 public class RunServerTask extends AbstractRunTask {
 	public RunServerTask() {
-		super(RunConfig::serverRunConfig);
+		super(
+				project -> {
```
i509VCB: Concern two electric boogloo

src/main/java/net/fabricmc/loom/util/Constants.java
```diff
@@ -126,4 +126,9 @@ private MixinArguments() {
 		private LaunchWrapper() {
 		}
 	}
+
+	public static final class Knot {
```
i509VCB, line 130: Add a private constructor here

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -58,6 +62,7 @@
 import net.fabricmc.loom.configuration.providers.minecraft.MinecraftMappedProvider;
 
 public class LoomGradleExtension {
+	@Deprecated // Configured in run configurations now
 	public String runDir = "run";
```
Juuxel: Does this still work? Generally `@Deprecated` stuff should work as usual, but I didn't see any compat code hooking this up to the new system. (If it doesn't work, it should be removed instead of confusing users)

src/main/java/net/fabricmc/loom/task/RunServerTask.java
ghost: That is very concerning...

Wouldn't it be best to not use a lambada?

<:concern:707016186281590855>

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
modmuss50: No, good point. I was just going to remove it. I don’t think it was commonly used anyway.

modmuss50: Seems to still be used to generate the directories, will full remove and fix.

modmuss50:
Right, gonna pull this in, its still experimental so please let any suggestions be known :)

FoxSamu:
image
It's working very well for me in the Gradle environment now that I first used it, although IntelliJ's inpsector seems to catch the `runs` field instead of the `runs` method when defining the runs block. I think it's because the field is not `private`. Generates this strange warning and prevents all autocompletion. I might have coded that myself in my PR, and not have realized that intellij would respond like this. I recommend changing the field name (and its getter/setter, but not the config method name) to `runConfigs` or something, to avoid confusion later on (now that it's still experimental that's probably still feasible)

modmuss50:
Yes, that sounds reasonable. I'll take a look later.

PR #343, opened 2021-02-02
Generate fabric.mod.json for included submodule jars

jpenilla:
When jij-ing subprojects, fabric.mod.json files are not generated like they are for other non-mod dependencies. This results in the jars being properly included but never being loaded.

This PR fixes that issue, although it's a bit ugly because  `ProjectDependency` and `ResolvedDependency` don't have a common parent.

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/NestedJars.java
+++ b/src/main/java/net/fabricmc/loom/build/NestedJars.java
@@ -255,7 +255,6 @@ public class NestedJars {
 	}
 
 	private static final class ProjectDependencyMetaExtractor implements DependencyMetaExtractor<ProjectDependency> {
-
 		@Override
 		public String group(ProjectDependency dependency) {
 			return dependency.getGroup();
@@ -273,7 +272,6 @@ public class NestedJars {
 	}
 
 	private static final class ResolvedDependencyMetaExtractor implements DependencyMetaExtractor<ResolvedDependency> {
-
 		@Override
 		public String group(ResolvedDependency dependency) {
 			return dependency.getModuleGroup();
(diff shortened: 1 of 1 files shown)


modmuss50:
What is your use for this? It seems easier (and better) to just make that sub project either a mod with better values than loom could come up with, or to shade it in?

jpenilla:
The use case for this is for the fabric platform of [cloud command framework](https://github.com/Incendo/cloud).
WIP PR: https://github.com/Incendo/cloud/pull/223
The structure of our project is that we have a core module, with several additional modules including various platform modules (Bukkit, Velocity, Sponge, etc).
Our `cloud-fabric` mod/module depends on core and other modules from the project, which are not mods.
Our current scripts jar in jar our submodule dependencies (`cloud-core`, `cloud-brigadier`, etc) as well as our one external dependency (geantyref), but fabric.mod.json files are only generated for the external dependency.
If we were to move `cloud-fabric` to it's own project, we could work around this issue without any changes to loom, however I don't see a reason not to add this feature to loom. The resulting jar/fabric.mod.jsons from the workaround (separate project) will be the same as with the changes from this PR.

zml2008:
hm, those modules could be shaded, but that makes the fabric distribution of cloud less flexible in the future -- if a module were added to the shaded distribution that had been previously include()d by other mods, there would be the potential for class file conflicts.

Shading is also not an option for geantyref -- multiple mods use it as a library, so for API interoperability it has to be JiJed unrelocated. `cloud` itself is a fairly young library, so I'd be in favor of being able to JiJ individual modules, just to keep more options open for the library down the road.

Less immediately relevant, but with any potential movement towards JPMS, shadowing will break module encapsulation, so I'd strongly prefer JiJ when possible to preserve signing and module metadata.

modmuss50 (approved):
Sounds resonable, changes look good.

Out of scope for this PR, but down the line id like to look at being able to provide custom mod metadata in place of the genertated stuff.

PR #346, opened 2021-02-09
Update mixin compile extensions + fixes

modmuss50:
Update mixin compile extensions
Fix mixin version parsing
Tweak some log output

I dont think the mixin version parsing was working too well, ive updated it to support all versions of mixin we have released. Im not sure if this should be back ported ot 0.5?

I have removed a whole handful of logging to cleanup the log, for loom's configuration I think it might be nice to add a progress bar (Similar to gen sources).


src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -326,18 +328,40 @@ public String getMixinJsonVersion() {
...
 
-			return dependency.getVersion();
+			return version;
 		}
 
 		return null;
 	}
 
+	public static String parseMixinVersion(String version) {
+		final Pattern pattern = Pattern.compile("(\\+mixin.)([0-9]+.[0-9]+)");
```
i509VCB: Make this pattern a static field so fabric api doesn't create 40 new patterns?

src/test/groovy/net/fabricmc/loom/MixinVersionTest.groovy
```diff
@@ -0,0 +1,21 @@
...
+
+		where:
+			input					| output
+			"0.7.11.10"				| "0.7"
+			"0.8.2+build.24"		| "0.8"
+			"0.9.1+mixin.0.8.2"		| "0.8"
+			"0.9.1+mixin.0.8.10"	| "0.8"
+			"1.12.3+mixin.0.8.2"	| "0.8"
+			"1.16.3+mixin.1.0.2"	| "1.0"
+			"1.16.3+mixin.1.10.0"	| "1.10"
```
i509VCB: Uh odd formatting?

i509VCB: Hmm that's weird, in viewing comments it's fine but diff view isn't.

src/test/groovy/net/fabricmc/loom/MixinVersionTest.groovy
```diff
@@ -0,0 +1,21 @@
...
+class MixinVersionTest extends Specification {
+
+	def "Mixin version parsing"() {
+		expect:
+			LoomGradleExtension.parseMixinVersion(input) == output
+
+		where:
+			input					| output
+			"0.7.11.10"				| "0.7"
+			"0.8.2+build.24"		| "0.8"
```
shedaniel: Shouldn't these return 0.8.2, I do see versions with min versions of 0.8.2

modmuss50: Thinking about it, it might just be best to remove this, the example mod sets it and im not even sure its actaully checked.

shedaniel: It is actually checked, but not that useful in the context of fabric as the loader determines the mixin version, mods should just declare the minimum loader version in their mod json.

modmuss50:
Removed the mixin min version. It should really be set by the mod, and not by loom's guess. Removes 100 lines of not so nice code as well.

PR #348, opened 2021-02-11
[WIP] Refactor meta handling + Asset/Native download/validation performance improvements.

modmuss50:
First pass on rewriting the Version Meta modal class, it was a mess before.
This also includes major performance improvements to download assets/natives.

The old MinecraftVersionInfo was a tottal mess of hacks upon hacks, this should be easier to maintain. I plan on adding some unit tests to ensure older mc versions are still supported.


modmuss50:
Tested on an intel mac and seems to work fine. Going to pull this there, please let me know if there are any issues as this is a rewrite to some loom's fundementals.

PR #349, opened 2021-02-12
RunConfigSettings.generateIDEConfig -> ideConfigGenerated + normal getter/setter

### LINKED ISSUE #347: Add conventional getter and setter to RunConfigSettings.generateIDEConfig for kt DSL
Without a conventional getter (with `is` for booleans) and setter it won't show up as a property in Kotlin DSL. (Groovy DSL also needs at least one of them for property format, but it can use the DSL-like method call syntax.)

Also, the name for the kotlin property would be `isGenerateIDEConfig` which is very ugly IMO, so I'd use `ideConfigGenerated` (=> `isIdeConfigGenerated`) or something similar instead.

Juuxel:
Fixes #347. This is a breaking change, but it shouldn't be too bad since it's for an experimental API.


modmuss50 (approved):
Looks good, thanks for this. 👍

PR #357, opened 2021-02-27
Fix custom sourceset not being correctly set for runconfig tasks

modmuss50:
A PR as 0.6 is technically stable now and I dont feel like testing or breaking it atm.


i509VCB (approved):

modmuss50:
Going to pull this into 0.7 so it can get a little bit of testing, I will prob backport this to 0.6 once im happy with it.

PR #361, opened 2021-03-05
Rewrite tests

modmuss50:
This PR rewrites the tests to make them easier to write, and easier for other people to use by example.

1) The project files (build.gradle and what not) are now stored as actaull files and not build up from a string.
2) The tests run in parallel on github actions, this should make it less painful in the future as more tests are added. (See: https://github.com/modmuss50/fabric-loom-1/actions/runs/625943230)
3) The test files now specify what gradle version to use, by default this is 6.8.3

I have written 3 new tests to start of with:

1) SimpleProjectTest - Builds the current example mod, and runs the 3 ide setup tasks.
2) ReproducibleBuildTest - Run the current example mod with the reproducible settings turned on. (Currently failing!/disabled)
3) MojangMappingsProjectTest - Builds a simple jar using the mojang mappings. 

Planned tests:

1) A subproject build - A lightweight fabric-api style thing.

Please feel free to suggest any extra tests, will prob get added down the line.


i509VCB:
Kotlin DSL tests would be nice to have, I can add some of those as needed.

modmuss50:
Ah yeah, I did think of that 👍

modmuss50:
I have added a super basic kotlin test. As well as a test using v1 mappings on 1.14.4 + gradle 4.9

PR #362, opened 2021-03-07
Remapping improvements.

modmuss50:
TODO:

- [x] Fix deprecation warnings on gradle 7
- [x] Cleanup the 2 code paths (shared caches or not)
- [x] Allow for custom batched remap jar jobs
- [x] Allow for using custom configurations for including jars
- [x] Allow for including FileCollections in jars. (Limited to only mods)
- [ ] ~~Allow for including files with custom mod metadata~~ Going to leave this till later

Some new API is exposed via this, its going to left as experimental as I do expect it to change freely.


modmuss50:
This should now be feature complete, I do want to bring this into my fabric PR (https://github.com/FabricMC/fabric/pull/1338) to ensure that it does cover everything.

modmuss50:
Assuming the tests pass this should now be ready.

PR #363, opened 2021-03-13
Try and fix reproducable builds on windows.

modmuss50:
And setup tests for it across windows, mac and linux.

Im a little stuck on this, tried a few things without much success.

### DIFF
--- a/.github/workflows/test-push.yml
+++ b/.github/workflows/test-push.yml
@@ -66,7 +66,32 @@ jobs:
           TEST_WARNING_MODE: fail
 
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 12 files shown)


modmuss50:
Im at a bit of a loss with this, gone a full ciricle to it only being broken on windows.

modmuss50:
Seems to just be the source hashes now. It seems like the test runner isnt always picking up changes locally, this seems to have been what has caused all of my confusion.

modmuss50:
Ayy! All checks passed, took a while. Spending hours on this while tired wasted a lot of time, but its there now.

PR #370, opened 2021-03-26
CRF linenumbers take 2

modmuss50:
Something is still wrong, not sure what though...

Inner classes seem to be handled wrong, might be the cause of this?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,7 +65,7 @@ dependencies {
 
 	// decompilers
 	implementation ('net.fabricmc:fabric-fernflower:1.3.0')
-	implementation ('org.benf:cfr:0.150')
+	implementation ('org.benf:cfr:0.151')
 
 	// source code remapping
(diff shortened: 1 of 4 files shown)


liach:
I remember that inner classes are dumped with dumpers with a custom type usage information. A bad overriding for one of those methods in cfr's code may be the problem. Also consider using fabric cfr, which supports adding javadoc to the decompiled code and naming lambdas.

modmuss50:
Yeah, we need to update fabric's cfr to include the linenumber changes first, getting it in here is easy enough but not worth doing if this issue cannot be fixed.

liach:
hmm, our cfr is up to date to at least jan 21. line number was aug 20 in upstream.

modmuss50:
Oh, so it is 👍

modmuss50:
Got a take 3 in the works, will open a new PR.

PR #373, opened 2021-03-28
Fix name comparision in GroovyXmlUtil

octylFractal:
Nodes can also have `groovy.xml.QName`s, which need to be compared using their `matches(Object)` method.

This fixes a bug where `fabric-loom` added an extra `dependencies` node in `MavenPublication#processEntry`, causing Gradle to fail the build.


octylFractal:
I have verified this fixes my problem locally.

modmuss50 (approved):
Looks good 👍

PR #379, opened 2021-04-02
Support Groovy 3's QName in GroovyXmlUtil

modmuss50:
Fixes a crash in gradle 7, also updated the tests to cover this.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/GroovyXmlUtil.java
+++ b/src/main/java/net/fabricmc/loom/util/GroovyXmlUtil.java
@@ -31,6 +31,8 @@ import java.util.stream.Stream;
 import groovy.util.Node;
 import groovy.xml.QName;
 
+import net.fabricmc.loom.util.gradle.GradleSupport;
+
 public final class GroovyXmlUtil {
 	private GroovyXmlUtil() { }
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/util/GroovyXmlUtil.java
```diff
@@ -63,9 +65,19 @@ private static boolean isSameName(Object nodeName, String givenName) {
 			return ((QName) nodeName).matches(givenName);
 		}
 
+		// New groovy 3 (gradle 7) class
+		if (GradleSupport.IS_GRADLE_7_OR_NEWER && nodeName.getClass().getName().equals("groovy.namespace.QName")) {
```
nnym, line 69: I think that you should be able to use `instanceof` here.
```suggestion
		if (GradleSupport.IS_GRADLE_7_OR_NEWER && nodeName instanceof groovy.namespace.QName) {
```

modmuss50: That will crash when being ran on groovy 2.

nnym: The short circuit should take care of it; no?

modmuss50: No, it wont as far as I know. It will try to load the class when that method is first called.

PR #394, opened 2021-05-03
Add classifier to "fake" mods' id if they have a classifier specified

### LINKED ISSUE #393: Include artifact classifier in generated `fabric.mod.json`
Currently, artifacts with different classifiers end up with the same generated ID. This causes the loader to only load one. I encountered this issue when trying to include native LWJGL libraries with certain classifiers and load them.

The relevant code:
https://github.com/FabricMC/fabric-loom/blob/12ac6cfcfe94fc3dd7029b18207b61c27effdea4/src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java#L194-L209

Since the classifier is typically empty, this wouldn't cause any backwards compatibility issues.

magneticflux-:
~~This needed a few changes to the `prepareForNesting` method to pass more about each file (`FileAndClassifier`)~~

Closes #393

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java
+++ b/src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java
@@ -192,7 +192,12 @@ public final class NestedDependencyProvider implements NestedJarProvider {
 
 		JsonObject jsonObject = new JsonObject();
 		jsonObject.addProperty("schemaVersion", 1);
-		jsonObject.addProperty("id", (metaExtractor.group(dependency) + "_" + metaExtractor.name(dependency)).replaceAll("\\.", "_").toLowerCase(Locale.ENGLISH));
+
+		jsonObject.addProperty("id",
+				(metaExtractor.group(dependency) + "_" + metaExtractor.name(dependency) + info.getClassifierSuffix())
+						.replaceAll("\\.", "_")
+						.toLowerCase(Locale.ENGLISH)
+		);
 		jsonObject.addProperty("version", metaExtractor.version(dependency));
 		jsonObject.addProperty("name", metaExtractor.name(dependency));
 
@@ -204,13 +209,21 @@ public final class NestedDependencyProvider implements NestedJarProvider {
 	}
 
 	private record DependencyInfo<D>(D dependency, DependencyMetaExtractor<D> metaExtractor, File file, @Nullable String classifier) {
-		public void validateInputs() {
+		void validateInputs() {
 			if (!file.exists()) {
 				throw new RuntimeException("Failed to include nested jars, as it could not be found @ " + file.getAbsolutePath());
 			}
 
 			if (file.isDirectory() || !file.getName().endsWith(".jar")) {
 				throw new RuntimeException("Failed to include nested jars, as file was not a jar: " + file.getAbsolutePath());
+			}
+		}
+
+		String getClassifierSuffix() {
+			if (classifier == null) {
+				return "";
+			} else {
+				return "_" + classifier;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


modmuss50:
Please target the 0.7 branch

magneticflux-:
@modmuss50 Should I target `0.8` now or is that only going to work with 1.17?

magneticflux-:
Looks like `AbstractArchiveTask.classifier` is deprecated and scheduled for removal in Gradle 8.0, but it was only introduced in Gradle 5.2. Can we bump the required Gradle version so we can use the new property?

modmuss50:
0.8 requires gradle 7. 

0.8 will work for older mc versions and is where this PR should target, should be easy enough to re-target

magneticflux-:
Looks like a spurious failure; that test passed on my local machine.

src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java
```diff
@@ -133,15 +129,16 @@ public static NestedDependencyProvider createNestedDependencyProviderFromConfigu
...
-				fileList.add(new DependencyInfo<>(dependency, new ResolvedDependencyMetaExtractor(), file));
-			}
+					.forEach((artifact) -> fileList.add(
+							new DependencyInfo<>(
+									dependency,
+									new ResolvedDependencyMetaExtractor(),
+									artifact.getFile(),
+									artifact.getClassifier()
+							))
+					);
```
haykam821: I would recommend using a normal for-loop here.

src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java
```diff
@@ -103,17 +103,13 @@ public static NestedDependencyProvider createNestedDependencyProviderFromConfigu
...
 
 				for (Task task : remapJarTasks.isEmpty() ? jarTasks : remapJarTasks) {
-					if (task instanceof RemapJarTask remapJarTask) {
-						File file = remapJarTask.getArchiveFile().get().getAsFile();
-						fileList.add(new DependencyInfo<>(projectDependency, new ProjectDependencyMetaExtractor(), file));
-					} else if (task instanceof AbstractArchiveTask abstractArchiveTask) {
+					if (task instanceof AbstractArchiveTask abstractArchiveTask) {
 						File file = abstractArchiveTask.getArchiveFile().get().getAsFile();
-						fileList.add(new DependencyInfo<>(projectDependency, new ProjectDependencyMetaExtractor(), file));
+						fileList.add(new DependencyInfo<>(projectDependency, new ProjectDependencyMetaExtractor(), file, abstractArchiveTask.getArchiveClassifier().getOrNull()));
```
haykam821: This line is quite long; could the archive classifier be placed on a separate line like the file is?

magneticflux-: I agree, I'll split that constructor into multiple lines.

src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java
magneticflux-: Yeah, that makes more sense. I used a stream since it previously did, but that's not necessary anymore.

src/main/java/net/fabricmc/loom/build/nesting/NestedDependencyProvider.java
```diff
@@ -103,17 +103,18 @@ public static NestedDependencyProvider createNestedDependencyProviderFromConfigu
...
-					} else if (task instanceof AbstractArchiveTask abstractArchiveTask) {
+					if (task instanceof AbstractArchiveTask abstractArchiveTask) {
 						File file = abstractArchiveTask.getArchiveFile().get().getAsFile();
-						fileList.add(new DependencyInfo<>(projectDependency, new ProjectDependencyMetaExtractor(), file));
+						fileList.add(new DependencyInfo<>(
+								projectDependency,
+								new ProjectDependencyMetaExtractor(),
+								file,
+								abstractArchiveTask.getArchiveClassifier().getOrNull()
+						));
```
haykam821: I don't think there's much of a point to keeping the `file` variable that helps with readability here if you are placing arguments on individual lines.

modmuss50 (approved):

PR #396, opened 2021-05-08
Field Migration

shedaniel:
Still having issues with source locations, would need to fix those first.
Fixes issues with remapping fields like KeyBinding.MAP, as the type of the field is changed with a forge patch.
Not sure about type migrations with methods however, but this PR will not cover those.


shedaniel:
oh wrong repo

PR #399, opened 2021-05-13
Add a java 8 bootstrap to provide better error messages.

modmuss50:
Add a boostrap subproject that is compiled against java 8, this can be used to validate gradle and java versions. We will be able to provide better error messages with steps to resolves either issue.


modmuss50:
This now seems to be working, need to apply the same module metadata hack to fernflower (it requires j11) to make it work all the way back to j8

modmuss50:
Going to close this as I think its going to be best for everyone to just build loom against j8 and use toolchains to allow gradle to download the correct version of java.

As much as I would like to use java 16 features in loom its clearly causing some people some grief trying to get it setup.

modmuss50:
Seems to be working now, got some nasty hacks in here to make it run on older java and espcially gradle.  Test coverage is somewhat manual (check the build logs) but I think this is fine for now as its not something that I expect will change much.

bootstrap/src/main/java/net/fabricmc/loom/bootstrap/LoomGradlePluginBootstrap.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * This bootstrap is compiled against a minimal gradle API and java 8, this allows us to show a nice error to users who run on unsupported configurations.
+ */
+@SuppressWarnings("unused")
+public class LoomGradlePluginBootstrap implements Plugin<PluginAware> {
+	private static final int MIN_SUPPORTED_MAJOR_GRADLE_VERSION = 7;
+	private static final int MIN_SUPPORTED_MAJOR_JAVA_VERSION = 16;
+
+	private static final String pluginClassName = "net.fabricmc.loom.LoomGradlePlugin";
```
Juuxel: `PLUGIN_CLASS_NAME` would be better since it's a constant

Issue #395, opened 2021-05-06
Incompatibility with dependencyResolutionManagement setting in gradle


Logiquo:
It seems that loom is incompatible with `dependencyResolutionManagement`. The problem is the same as the one described in this issue  https://github.com/gradle/gradle/issues/15754.

To summarize, the three options in `dependencyResolutionManagement` are
- `FAIL_ON_PROJECT_REPOS`: a repository declared in the project will trigger a build error.
- `PREFER_PROJECT`: ignore repository declared in the settings.
- `PREFER_SETTINGS`: ignore repository declared in the project.
based on documentation https://docs.gradle.org/current/javadoc/org/gradle/api/initialization/resolve/RepositoriesMode.html

Since loom declared its repository in the plugin, it will cause `FAIL_ON_PROJECT_REPOS` and `PREFER_SETTINGS` not to work. However, `PREFER_PROJECT` will cause the Gradle to ignore the repository declared in the settings, which can be problematic for large projects.

One solution might be to not declare a repo in the plugin, but I am not sure if it is possible.

liach:
The idea on moving away from hardcoded repos and declaring them in setting gradle does have advantages, like people can then use proxy maven for minecraft or fabric stuff. I personally support doing so, but there are other hardcoded components, like the launchermeta for minecraft.

On this setting thing, it's weird that gradle doesn't allow repositories to be declared in both places with this update.

Logiquo:
I think the solution for this problem should similar to https://github.com/SpongePowered/VanillaGradle/blob/7b86f75197363cf725c4351dd5a46de62e021c4b/src/main/java/org/spongepowered/gradle/vanilla/VanillaGradle.java
Make the loom both works for settings and project, and when the plugin applies on settings it add dependency and when the plugin applies to the project it does other stuff.

liach:
I suggest shipping an additional plugin that sets the repositories for the settings than making the same plugin work for both environments like sponge does. having the same plugin target two different things is quite confusing.

Logiquo:
Currently many repos (i.e. repos related to mappings) depend on the configuration in the projects. Without a local repo re-structure, it is impossible to make the loom work with dependencyResolutionManagement. Not matter we choose to use the same plugin or use another plugin for repos. I think it is necessary to remove the project-configuration-specific-name in the path of the repo.

Logiquo:
[...]

PR #400, opened 2021-05-14
add support for new dependencyResolutionManagement

Logiquo:


Logiquo:
This is related to #395.

modmuss50:
Had a quick look from my phone and looks good. 👍 can you add an integration test for this?

Let me know if you need help with that.

src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
```diff
@@ -40,37 +41,43 @@
...
-public class LoomGradlePlugin implements Plugin<Project> {
+public class LoomGradlePlugin implements Plugin<PluginAware> {
 	public static boolean refreshDeps;
 	public static final Gson GSON = new GsonBuilder().setPrettyPrinting().create();
 	public static final ObjectMapper OBJECT_MAPPER = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
 	@Override
-	public void apply(Project project) {
-		project.getLogger().lifecycle("Fabric Loom: " + LoomGradlePlugin.class.getPackage().getImplementationVersion());
+	public void apply(PluginAware target) {
```
liach, line 50: The diff can be simplified by retaining `apply(Project)` and adding a `apply(PluginAware)` at the same time. The override will move to the `apply(PluginAware)`, which will be:
```java
	@Override
	public void apply(PluginAware target) {
		target.getPlugins().apply(LoomRepositoryPlugin.class);
		if (target instanceof Project) {
			apply((Project) target);
		}
	}
```

src/main/java/net/fabricmc/loom/configuration/processors/MinecraftProcessedProvider.java
```diff
@@ -37,15 +39,20 @@
...
 
 public class MinecraftProcessedProvider extends MinecraftMappedProvider {
-	public static final String PROJECT_MAPPED_CLASSIFIER = "projectmapped";
+	public final String PROJECT_MAPPED_CLASSIFIER;
 
 	private File projectMappedJar;
 
 	private final JarProcessorManager jarProcessorManager;
 
+	private static String getProjectUUID(Project project) {
```
liach: Mind rename to `getProjectUuid` which is the fabric style?

src/main/java/net/fabricmc/loom/configuration/LoomDependencyManager.java
```diff
@@ -193,7 +194,9 @@ private static void handleInstallerJson(JsonObject jsonObject, Project project)
 
 			project.getLogger().debug("Loom adding " + name + " from installer JSON");
 
-			if (jsonElement.getAsJsonObject().has("url")) {
+			// If user choose to use dependencyResolutionManagement, then they should declare
+			// these repositories manually in the settings file.
+			if (jsonElement.getAsJsonObject().has("url") && !project.getGradle().getPlugins().hasPlugin(LoomRepositoryPlugin.class)) {
 				String url = jsonElement.getAsJsonObject().get("url").getAsString();
 				long count = project.getRepositories().stream().filter(artifactRepository -> artifactRepository instanceof MavenArtifactRepository)
```
liach, line 201: So here, when the count is 0, I think the best approach is to leave a warning or fail the build (maybe add a command-line option to control that) when the desired maven doesn't exist. Imo we should fail by default, but a system property for warning can allow people to use proxy mavens.

src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
```diff
@@ -40,37 +41,43 @@
...
+		target.getPlugins().apply(LoomRepositoryPlugin.class);
 
-		refreshDeps = project.getGradle().getStartParameter().isRefreshDependencies();
+		if (target instanceof Project) {
+			Project project = (Project) target;
```
Juuxel:
```suggestion
		if (target instanceof Project project) {
```

src/main/java/net/fabricmc/loom/configuration/processors/MinecraftProcessedProvider.java
```diff
@@ -37,15 +39,20 @@
 import net.fabricmc.loom.util.Constants;
 
 public class MinecraftProcessedProvider extends MinecraftMappedProvider {
-	public static final String PROJECT_MAPPED_CLASSIFIER = "projectmapped";
+	public final String PROJECT_MAPPED_CLASSIFIER;
```
Juuxel: This should also follow the naming conventions (`projectMappedClassifier`)

src/main/java/net/fabricmc/loom/configuration/MavenConfiguration.java
```diff
@@ -26,32 +26,7 @@
 
 import org.gradle.api.Project;
 
-import net.fabricmc.loom.LoomGradleExtension;
-
 public class MavenConfiguration {
 	public static void setup(Project project) {
```
Juuxel: I think this can just be removed now - all it did was set up those repos.

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+
+		if (target instanceof Settings) {
+			Settings settings = (Settings) target;
+
+			repositories = settings.getDependencyResolutionManagement().getRepositories();
+
+			// leave a marker so projects don't try to override these
+			settings.getGradle().getPluginManager().apply(LoomRepositoryPlugin.class);
+		} else if (target instanceof Project) {
+			Project project = (Project) target;
```
Juuxel:
```suggestion
		} else if (target instanceof Project project) {
```

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+import org.gradle.api.invocation.Gradle;
+import org.gradle.api.plugins.PluginAware;
+
+public class LoomRepositoryPlugin implements Plugin<PluginAware> {
+	@Override
+	public void apply(PluginAware target) {
+		RepositoryHandler repositories = null;
+
+		if (target instanceof Settings) {
+			Settings settings = (Settings) target;
```
Juuxel:
```suggestion
		if (target instanceof Settings settings) {
```

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+		}
+	}
+
+	File getUserCache() {
+		File gradleUserHomeDir = null;
+
+		if (target instanceof Settings) {
+			gradleUserHomeDir = ((Settings) target).getGradle().getGradleUserHomeDir();
+		} else if (target instanceof Project) {
+			gradleUserHomeDir = ((Project) target).getGradle().getGradleUserHomeDir();
```
Juuxel:
```suggestion
		if (target instanceof Settings settings) {
			gradleUserHomeDir = settings.getGradle().getGradleUserHomeDir();
		} else if (target instanceof Project project) {
			gradleUserHomeDir = project.getGradle().getGradleUserHomeDir();
```

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+		return userCache;
+	}
+
+	public File getRootPersistentCache() {
+		File rootDir = null;
+
+		if (target instanceof Settings) {
+			rootDir = ((Settings) target).getRootDir();
+		} else if (target instanceof Project) {
+			rootDir = ((Project) target).getRootDir();
```
Juuxel:
```suggestion
		if (target instanceof Settings settings) {
			rootDir = settings.getRootDir();
		} else if (target instanceof Project project) {
			rootDir = project.getRootDir();
```

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+		return persistentCache;
+	}
+
+	public File getRootBuildCache() {
+		File rootDir = null;
+
+		if (target instanceof Settings) {
+			rootDir = ((Settings) target).getRootDir();
+		} else if (target instanceof Project) {
+			rootDir = ((Project) target).getRootDir();
```
Juuxel:
```suggestion
		if (target instanceof Settings settings) {
			rootDir = settings.getRootDir();
		} else if (target instanceof Project project) {
			rootDir = project.getRootDir();
```

Juuxel (changes requested):

Logiquo:
> Had a quick look from my phone and looks good. 👍 can you add an integration test for this?
> 
> Let me know if you need help with that.

Hi, I did not have any experience in writing integration tests. If you can give me some examples/tutorials then I will be happy to write it.

liach:
@LogicFan seems what you said in https://github.com/FabricMC/fabric-loom/issues/395#issuecomment-841046582 still needs to be addressed?

Logiquo:
> @LogicFan seems what you said in [#395 (comment)](https://github.com/FabricMC/fabric-loom/issues/395#issuecomment-841046582) still needs to be addressed?

I agree with your solution, however, I am still testing this part of the code to make sure the behaviour is expected. So a commit will be up for today or tomorrow after I am confident that my code is correct.

liach:
Also on unit tests, I think you will make a backbone project at https://github.com/FabricMC/fabric-loom/tree/dbc1f4cba7ad6c9bbce0aea46e682897a61c53dd/src/test/resources/projects and add a driver code file at https://github.com/FabricMC/fabric-loom/tree/dbc1f4cba7ad6c9bbce0aea46e682897a61c53dd/src/test/groovy/net/fabricmc/loom/test/integration

src/main/java/net/fabricmc/loom/configuration/LoomDependencyManager.java
Logiquo: The `settings.getDependencyResolutionManagement().getRepositories()` and `project.getRepositories()` are different object in gradle. Since we do not add any repo into `project.getRepositories()` if `dependencyResolutionManagement` used, the `project.getRepositories()` will always be empty. Furthermore, there is currently no way of retrieving `settings` in project lifecycle, this means we do not know if a repo is added or not. Thus, the `count` will always be zero. 

This limitation will cause your idea hard to implement, and hence I prefer the existing code (I really don't have any other solution for this).

Logiquo:
The test is added. Waiting for approval.

src/test/resources/projects/dependencyResolutionManagement/projmap/src/main/resources/modid.accesswidener
```diff
@@ -0,0 +1,2 @@
+accessWidener   v1   named
+extendable   class   FluidState
```
modmuss50: This isnt a valid accesswidener, I would remove it as I dont see how its realted to this test.

modmuss50: Oh, I see it forces a project mapped jar, still not a valid class name

Logiquo: Sorry, I do not have much experience with how the accesswidener works. This test is to force a build to use a project-mapped jar. Could you change this to a valid accesswidener? Thanks.

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -0,0 +1,200 @@
...
+			repo.setUrl(cache.getRootPersistentCache());
+			repo.patternLayout(layout -> {
+				layout.artifact("[revision]/[artifact]-[revision](.[ext])");
+			});
+			repo.metadataSources(IvyArtifactRepository.MetadataSources::artifact);
+		});
+	}
+}
+
+final class Cache {
```
modmuss50, line 100: Might be out of scope for this PR, but it does duplicate some stuff in LoomGradleExtension, moving this out into its own class and removing the duplicates might not be a bad idea. 

At somepoint I want to give LGE a good cleanup so im not too fussed about doing this now.

modmuss50 (approved):

src/test/resources/projects/dependencyResolutionManagement/projmap/src/main/resources/modid.accesswidener
modmuss50:
```suggestion
extendable   class   net/minecraft/fluid/FluidState
```

Logiquo:
Hi, may I ask what is the ETA to merge this PR? Is there anything more I need to do? This is my first PR in GitHub so I do not have much experience.

liach (approved):

modmuss50:
Should be ready to merge, I just want to test it locally first.

src/main/java/net/fabricmc/loom/configuration/processors/MinecraftProcessedProvider.java
```diff
@@ -37,15 +39,20 @@
...
 public class MinecraftProcessedProvider extends MinecraftMappedProvider {
-	public static final String PROJECT_MAPPED_CLASSIFIER = "projectmapped";
+	public final String projectMappedClassifier;
 
 	private File projectMappedJar;
 
 	private final JarProcessorManager jarProcessorManager;
 
+	private static String getProjectUuid(Project project) {
+		return UUID.nameUUIDFromBytes(project.getProjectDir().toString().getBytes(StandardCharsets.UTF_8)).toString();
```
modmuss50: Im not sure if a whole UUID is needed? would just the project name (formatted in a classifier safe format) be a better solution?

Logiquo: In theory, anything that unique for the project should be fine. UUID is just for naming collision avoidance. I don't know if the project name would be safe enough for this. Can people define two sub-projects with the same name?

liach: I think if you use full subproject names, you are fine. You can define like `:sub1:duplicate` and `:sub2:duplicate` subprojects within one project I think, but their full project names differ.

Logiquo: I see, I will change it to the fully-qualified-name. It will be more human-readable, and hopefully shorter. A commit will be up for this tomorrow.

Logiquo: A commit is up. My biggest concern is what if there is invalid char (i.e. cannot exists in the path) inside the project name. Any solution for this? Should we write a transformer that replaces all the invalid char?

I cannot find a good resource for what is a valid name for the Gradle project, seems and char except `:` are legal.

PR #413, opened 2021-06-08
Basic layered mappings with @ParchmentMC support

modmuss50:
See [parchmentmc.org](https://parchmentmc.org/announcement/release/2021/06/08/announcement-parchment-initial-release.html) for more info about this project. TLDR; Mojang mapping params and javadoc. Constant unpicking is not supported. This can be used like so:

```groovy
repositories {
	maven {
		name = "ldtteam"
		url = "https://ldtteam.jfrog.io/artifactory/parchmentmc-snapshots/"
	}
}

dependencies {
	minecraft "com.mojang:minecraft:1.16.5"
	mappings loom.layered() {
		officalMojangMappings()
		parchment("org.parchmentmc.data:parchment-1.16.5:20210608-SNAPSHOT@zip")
	}


	...
}
```

No default maven server with parchment mappings is included, this is not something we plan to re-host, I imagine the forge/sponge guys will also have it on their maven.

This PR also lays some of the basic groundwork for layered mappings, by rewriting the mojang mapping handling to use our new [https://github.com/FabricMC/mapping-io/](mapping-io) library.

This is by no means going to lead to using over yarn by default, I still strongly believe yarn's benefits out weigh even this. But it will make using mojmaps much nicer for those that choose to do so.


modmuss50:
This PR now includes the start of layered mappings, not tested at all right now, but should be very unit testable.

src/test/groovy/net/fabricmc/loom/test/util/ProjectTestTrait.groovy
```diff
@@ -29,7 +29,7 @@ import org.gradle.testkit.runner.GradleRunner
 
 trait ProjectTestTrait {
 	final static String DEFAULT_GRADLE = "7.0.1"
-	final static String PRE_RELEASE_GRADLE = "7.2-20210527220045+0000"
+	final static String PRE_RELEASE_GRADLE = "7.2-20210612220215+0000"
```
modmuss50: Looks like this has deprecated some stuff we are using, I will invistage this separately.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/mojmap/MojangMappingLayer.java
```diff
@@ -0,0 +1,102 @@
...
+	@Override
+	public void visit(MappingVisitor mappingVisitor) throws IOException {
+		var clientMappings = new File(workingDir(), "client.txt");
+		var serverMappings = new File(workingDir(), "server.txt");
+
+		download(clientMappings, serverMappings);
+
+		printMappingsLicense(clientMappings.toPath());
+
+		// Make named the source namespace
```
Player3324: doc is backwards

src/main/java/net/fabricmc/loom/configuration/providers/mappings/intermediary/IntermediaryMappingLayer.java
```diff
@@ -0,0 +1,64 @@
...
+	public MappingNamespace getSourceNamespace() {
+		return MappingNamespace.OFFICIAL;
+	}
+
+	@Override
+	public void visit(MappingVisitor mappingVisitor) throws IOException {
+		// Populate named with intermediary
+		MappingNsCompleter nsCompleter = new MappingNsCompleter(mappingVisitor, Collections.singletonMap(MappingNamespace.NAMED.stringValue(), MappingNamespace.INTERMEDIARY.stringValue()));
+
+		// TODO can we save creating a tree just to add a namespace?
```
Player3324: https://github.com/FabricMC/mapping-io/commit/a54ad14a4c24226a8b523ca3d53ce9309e182273 offers a clean alternative

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentPrefixStripingMappingVisitor.java
```diff
@@ -0,0 +1,50 @@
...
+import net.fabricmc.mappingio.adapter.ForwardingMappingVisitor;
+
+public final class ParchmentPrefixStripingMappingVisitor extends ForwardingMappingVisitor {
+	protected ParchmentPrefixStripingMappingVisitor(MappingVisitor next) {
+		super(next);
+	}
+
+	@Override
+	public boolean visitMethodArg(int argPosition, int lvIndex, String srcName) {
+		return super.visitMethodArg(argPosition, lvIndex, stripMethodArg(srcName));
```
Player3324, line 39: Are you sure the srcName is appropriate here? If the mappings are not backwards it's either null or snowman. Assuming the dst name is with the prefix, overriding visitDstName and checking the targetKind for METHOD_ARG would be the appropriate take.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentTreeV1.java
```diff
@@ -0,0 +1,165 @@
...
+		String version,
+		@Nullable
+		List<Class> classes,
+		@Nullable
+		List<Package> packages
+) {
+	public void visit(MappingVisitor visitor, String srcNamespace) {
+		while (true) {
+			if (visitor.visitHeader()) {
+				visitor.visitNamespaces(srcNamespace, Collections.emptyList());
```
Player3324, line 45: dst ns shouldn't be empty, named maybe? or the generic MappingUtil.NS_TARGET_FALLBACK?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentTreeV1.java
```diff
@@ -0,0 +1,165 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import net.fabricmc.mappingio.MappedElementKind;
+import net.fabricmc.mappingio.MappingVisitor;
+
+public record ParchmentTreeV1(
+		String version,
+		@Nullable
+		List<Class> classes,
```
Player3324, line 38: A potential optimization is making this visit the underlying visitor on the fly from JsonReader. This would need a temporary MemoryMappingTree if flags contains multiple_passes (see e.g. the enigma reader)

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentTreeV1.java
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	public record Parameter(
+			int index,
+			String name,
+			@Nullable
+			String javadoc
+	) {
+		public void visit(MappingVisitor visitor) {
+			if (visitor.visitMethodArg(index, index, name)) {
```
Player3324, line 149: assuming name is the dst name, this would have to pass null in visitMethodArg's `srcName` and instead call visitDstName between visitMethodArg and visitElementContent.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentTreeV1.java
Player3324: actually it is only mojmap, so ignore this and some related comments

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentTreeV1.java
modmuss50: Going to keep with the current solution, it should be fairly short lived and not that large in the context of gradle.

modmuss50:
Going to pull this in now, the whole layered mappings stuff is far from done, and will be worked on in smaller follow up PRs. The current layered mappings and parchment API is unstable so I may well break it, that does mean if you have ideas or suggestions that now is the time to make them known.

liach:
Hmm, I found shedaniel has a similar project at https://github.com/architectury/crane. Will loom add dedicated support, or at least test cases, for it?

shedaniel:
Ideally it would just apply like a normal tiny layer.

modmuss50:
Yes, once I get around to finishing the mapping-io layer this should just work.

PR #414, opened 2021-06-11
Fix a bug during genSources

### LINKED ISSUE #369: "File already in use" errors in some cases
I have not observed this issue myself, but if you search for `gradlew stop` in the Fabric discord you can see many instances of people encountering this error. If there is a way for Loom to handle this better or provide nicer diagnostics, that would be very nice.

frqnny:
Fixes #369 

Basically, it would break on Windows systems because '/' is a Unix thing, so this field should work on both Windows and Unix.


frqnny:
Closing since this didn't fix it. I also learned that I probably shouldn't try messing with gradle plugins because I have no idea what I'd be doing.

PR #417, opened 2021-06-13
Add AbstractFernFlowerDecompiler.configureJavaExec

Juuxel:
For example, this can be used to add files the decompiler classpath. (Specifically, I'd use this to make a configurable decompiler where it downloads the wanted version based on the project configuration.)


modmuss50:
I dont think im going to merge this, I want to rework how the fernflower decompiler works to use a gradle configuration for the classpath and pull fernflower from maven instead of trying to relaunch. For now just duplicate this class as its going to break either way.

Juuxel:
Yeah, I already went with [duplication](https://github.com/Juuxel/LoomQuiltflower/blob/master/src/main/java/juuxel/loomquiltflower/impl/loom/AbstractFernFlowerDecompiler.java) since I want to support a wide range of Loom versions in my plugin anyway.

PR #421, opened 2021-06-23
[0.7] Backport explicit Eclipse dependency versions

halotroop2288:
This is needed for Legacy Fabric. Eclipse libraries were causing builds to fail on JDK 1.8.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-all.zip
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,12 @@ repositories {
 	mavenCentral()
 }
 
+configurations.all {
+	resolutionStrategy {
+		failOnNonReproducibleResolution()
+	}
+}
+
 dependencies {
 	implementation gradleApi()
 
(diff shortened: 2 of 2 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-all.zip
```
modmuss50, line 6: No! We cannot just bump gradle versions like this for old versions

modmuss50 (changes requested):
Have you tested this? Especially on older gradle versions like 4.9?

gradle/wrapper/gradle-wrapper.properties
halotroop2288: Please suggest how to fix this in version 4.9, then:
```
> Could not find method failOnNonReproducibleResolution() for arguments [] on object of type org.gradle.api.internal.artifacts.ivyservice.resolutionstrategy.DefaultResolutionStrategy.
```
Without this block, it fails to find the Eclipse dependencies:
https://github.com/FabricMC/fabric-loom/pull/421/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R36-R40

```
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not find org.eclipse.platform:org.eclipse.compare.core:[[3.6.1000]].
```

modmuss50: uhh, this is what I was worried about. We cannot just bump the minimum version of gradle like that. 

In your settings.gradle file please make sure that the fabric maven is the top one as we have applied a server side "fix" on our maven.

halotroop2288: > uhh, this is what I was worried about. We cannot just bump the minimum version of gradle like that.

For what reason? Some misguided attempt at support for older non-LTS versions of Java?

> In your settings.gradle file please make sure that the fabric maven is the top one as we have applied a server side "fix" on our maven.

For what? Legacy Fabric API? Or Loom? In either case, enforcing a load order for repositories is kind of concerning. And pushing "fixes" directly to Maven instead of automating them is extra concerning.

modmuss50: Loom 0.7 and bellow support gradle 4.9 we cannot just break this with a minor patch.

Your project, this should never been needed but its now forced becuase eclipse didnt use hard coded deps. We had to get a fix out that sovled the issue for most people.

Sadly I dont have control over what eclipse does to their deps this should have never happened in the first place.

halotroop2288: > Loom 0.7 and below support Gradle 4.9 we cannot just break this with a minor patch.

OK that's fine. I still think we should find a workaround that works on 4.9 so it can be automated. I'll leave this open for discussion, still using your suggestion for now.

modmuss50: Yes, id like to test just paching the pom as in theory that should work.

modmuss50:
Replaced with: https://github.com/FabricMC/fabric-loom/pull/429

PR #422, opened 2021-06-24
Inline compatibility declaration

Ab20zipYT:
There is absolutely no need of writing two lines to declare the source and target Java compatibility.
Also, it is better to use the JavaVersion class to get the final constants of Java versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,8 +11,7 @@ plugins {
 	id "org.cadixdev.licenser" version "0.5.0"
 }
 
-sourceCompatibility = 16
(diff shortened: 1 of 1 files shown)


modmuss50:
While this PR isnt wrong looking at your account history and been warned by others I am going to close all of your PRs. My assumption is you want to get a PR in so we dont need to approve workflow runs.

The example mod is the main place where things such as this should be "perfect" but im not too bothered about the libs.

Sorry if you are just trying to be helpful, if so this possibly isnt the right way to go about it.

PR #423, opened 2021-06-26
Better Mixin AP configuration

Logiquo:
This is still WIP. ~The purpose of this PR is to help multi-project build which may involve other projects containing mixin code~.

I think it also makes sense to let the AP application be configurable (i.e. apply selected projects, not all projects). Suggestions are needed.

After discussion with modmuss50, the goal is to have the ability to specify ref-map for each `sourceSet` similar to what MixinGradle does.

- [X] [Tested] Ability to change the ref-map name based on each source set.
- [X] [Tested] Ability to customize which source set applies mixin annotation processor.
- [X] [Tested] Ability to do auto-ref-map for `*.mixin.json`
- [ ] ~[NotTested] [Experimental] Ability to cross-project.~ (too hacky) 

The syntax will be the following
```Gradle
mixin {
        // If the user do not add any source set, then the default source sets are all non-test source sets inside current project (same as before) with refmap name being loom.refmapName. 
	add(sourceSets["main"], "refmap0001.json") // add source set main to mixin ap, with output refmap name being refmap0001.json
	add(sourceSets["mixin"]) // add source set mixin to mixin ap, with output refmap name being loom.refmapName
	add(sourceSets["mixin1"], "refmap0003.json") {
                // You can optionally exclude or include any mixin json to let loom automatically put the refmap entry. The default is any `*.mixins.json` inside resource directory.
		it.exclude("m1_2.mixins.json")
	}
}
```


Logiquo:
Currently, if we use multiple source sets and shadow them at the end, then this would happen. 
image

There are two solutions to this. 
1. Using a classifier for different source sets.
2. Writing a custom transformer to merge JSON. 

@modmuss50 Which one do you prefer?

Juuxel:
I'd say the different source sets would work well. It already matches what Loom is doing with subprojects AFAIK. However, the mixin AP shouldn't be applied to `allprojects` - you can have subprojects in the hierarchy that have nothing to do with Loom or mixins.

Logiquo:
The idea is to introduce a new syntax that allows programmers to specify which source sets apply mixin AP and its corresponding refMapName.

It will be similar to
```
mixin {
  add(sourceSets.main, "abc_refmap.json")    // the refmap generated for this source set will name "abd_refmap.json"
  add(anotherSourceSets)                              // the refmap genertaed for this soruce set will use name from loom.refmapName
}
```
And this should also automatically add refmap to its corresponding `*.mixin.json` based on its file location (similar to what existing automatically adding refmap to `*.mixin,json`, which allow different name)

If the programmer does not configure `mixin` extension, then the current behaviour should preserved. (Namely, all non-test source sets under current project)

Logiquo:
Hi, I think this PR is ready to review. Even if the cross-project part does not work, I can disable it in one line and the other parts of the function won't be affected. @modmuss50 @sfPlayer1

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
```diff
@@ -0,0 +1,274 @@
...
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.util.PatternSet;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A gradle extension to configure mixin annotation processor.
+ */
+public class MixinAnnotationProcessorExtension {
```
modmuss50: Any reason this is its own extension, couldnt it just be part of looms? so:

```
loom {
   mixin {
   }
}
```

Idk why someone would use mixin gradle with loom but I dont think we need to have this clash?

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
```diff
@@ -0,0 +1,274 @@
...
+	public static final String MIXIN_INFORMATION_CONTAINER = "mixin";
+	private boolean isDefault;
+
+	private final Project project;
+	private final String loomId;
+	private boolean isCrossProject;
```
modmuss50: I dont think cross project is ever going to work and is going to be a massive pain to maintian, id just remove it.

Logiquo: I agree, I test it locally and it is super-hacky. I will remove it in the next commit.

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
Logiquo: There is no technical difficulty with this. However, I think making them separate will be cleaner for writing build scripts. And also touching LoomGradleExtension is a pain for me. 

About the clashing, how about I re-name it to `loom-mixin`. Or if you really want I can make it inside LGE. What's your idea?

modmuss50: It should still be its own class, but you can just have a `mixin(Action<MixinAnnotationProcessorExtension>)` method in LGE.

LGE needs a good cleanup, there is a lot of stuff in there that doesnt belong there, but thats out of scope here.

Logiquo: Then I think I will rename `mixin` to `loomMixin` and add `mixin(Action<MixinAnnotationProcessorExtension>)` into LGE, so people can choose whatever they like. How do you think?

Chocohead: I don't see why this is in need of a separate extension in the first place, when would it not be possible to access the `loom` extension but it would `loomMixin`?

Logiquo: To be honest, I just don't know how to register an `extension` without a name in gradle. Is this even something possible? The main point here is that I don't want to put new code into LGE since it is hard to work with.

Logiquo: @Chocohead any suggestion?

modmuss50: My PR (https://github.com/FabricMC/fabric-loom/pull/431) to refactor LGE should help make this easier, it really shouldnt be its own extension.

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
```diff
@@ -0,0 +1,236 @@
...
+	 * A information container stores necessary information
+	 * for configure mixin annotation processor. It stores
+	 * in [SourceSet].ext.mixin.
+	 */
+	public static final class MixinInformationContainer {
+		private final SourceSet sourceSet;
+		private final String refmapName;
+		private final PatternSet mixinJsonPattern;
+
+		private Stream<String> mixinJsonNames;
```
modmuss50: Why a Stream and not something more conventional like a set or list?

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
```diff
@@ -0,0 +1,236 @@
...
+						container.setMixinJsonNames(mixinJsonNames);
+						return true;
+					}
+
+					return false;
+				});
+	}
+
+	@NotNull
+	public Stream<Configuration> getApConfigurations(Project project0, Function<String, String> getApConfigNameFunc) {
```
modmuss50: Why pass in project0 when you already have project? I guess this is left over from the days of this being across project.

src/test/resources/projects/mixinApAutoRefmap/build.gradle
```diff
@@ -0,0 +1,139 @@
+plugins {
```
modmuss50, line 1: Amazing test 👍

src/main/java/net/fabricmc/loom/build/mixin/KaptApInvoker.java
```diff
@@ -62,18 +64,26 @@ public KaptApInvoker(Project project) {
 		kaptExtension.setIncludeCompileClasspath(false);
 	}
 
+	private static Map<SourceSet, JavaCompile> getInvokerTasks(Project project) {
+		MixinAnnotationProcessorExtension mixin = project.getExtensions().getByType(MixinAnnotationProcessorExtension.class);
+		return mixin.getInvokerTasks(project, AnnotationProcessorInvoker.JAVA)
```
modmuss50: Question: Should this be java or kotlin?

modmuss50 (changes requested):

src/main/java/net/fabricmc/loom/build/mixin/KaptApInvoker.java
Logiquo: The old code uses Java. I basically copy & paste it here.

modmuss50: Thats fine, I should have checked, was just questioning it.

Logiquo:
I am wondering if this is still useful as I am in favour of https://github.com/FabricMC/tiny-remapper/pull/63. @modmuss50 What do you think?

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
Logiquo: fixed

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
Logiquo: Because it was from `Stream` (see code below)
```Java
Stream<String> mixinJsonNames = sourceSet.getResources().matching(pattern).getFiles().stream().map(File::getName);
```
and will be used as `Stream`
```Java
mixinJsonNames.map(
```
It is unnecessary to collect it then stream it.

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
Logiquo: done.

Logiquo:
@modmuss50 Do you mind having a look?

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
```diff
@@ -0,0 +1,236 @@
...
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.util.PatternSet;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A gradle extension to configure mixin annotation processor.
+ */
```
modmuss50: The public api of this class should be defined in an interface, shouldnt be too much of a pain to do.

modmuss50 (approved):
Im happy to go with this, its done so there is no point in not using it. Marking it as experimental might be a good idea so we can iterate on it later if needed.

Just 1 minor comment, and needs merging.

src/main/java/net/fabricmc/loom/MixinAnnotationProcessorExtension.java
Logiquo: Done. Waiting for the test result to see if I break anything.

Logiquo:
I don't know how to mark the API as experimental. Do I need to specify some annotation? Could you help me with that?

modmuss50:
Put `@ApiStatus.Experimental` from `org.jetbrains.annotations.` on the interface, and the method in LoomGradleExtensionAPI

Logiquo:
done.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -82,6 +83,7 @@ default Dependency officialMojangMappings() {
 
 	NamedDomainObjectContainer<RunConfigSettings> getRunConfigs();
 
+	@ApiStatus.Experimental
 	void mixin(Action<MixinApExtension> action);
```
Juuxel: Shouldn't this accept the API?

Logiquo: True. I forget this part when I do the refactor.

Logiquo: updated.

src/main/java/net/fabricmc/loom/api/MixinApExtensionAPI.java
```diff
@@ -0,0 +1,52 @@
...
+import org.gradle.api.tasks.util.PatternSet;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Experimental
+public interface MixinApExtensionAPI {
+	/**
+	 * Apply Mixin AP to sourceSet.
+	 * @param sourceSet the sourceSet that applies Mixin AP
+	 * @param refmapName the output ref-map name
+	 */
```
Juuxel, line 42: It might be good to add these javadoc comments to all methods, and document the `Action<PatternSet>`, but not a blocking issue.

src/test/resources/projects/mixinApSimple/build.gradle
```diff
@@ -0,0 +1,119 @@
...
+}
+
+		assemble {
+	dependsOn tasks["mixinJar0"]
+	dependsOn tasks["mixinJar1"]
+}
+
+// configure the maven publication
+		publishing {
+	publications {
```
Juuxel:
```suggestion
assemble {
	dependsOn tasks["mixinJar0"]
	dependsOn tasks["mixinJar1"]
}

// configure the maven publication
publishing {
	publications {
```

src/main/java/net/fabricmc/loom/extension/MixinApExtension.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * A gradle extension to configure mixin annotation processor.
+ */
+@ApiStatus.Experimental
+public interface MixinApExtension extends MixinApExtensionAPI {
+	String MIXIN_INFORMATION_CONTAINER = "mixin";
+
+	/**
+	 * A information container stores necessary information
```
Juuxel:
```suggestion
	 * An information container stores necessary information
```

src/main/java/net/fabricmc/loom/extension/MixinApExtension.java
```diff
@@ -0,0 +1,126 @@
...
+	 * for configure mixin annotation processor. It stores
```
Juuxel:
```suggestion
	 * for configuring the mixin annotation processor. It's stored
```

Juuxel (approved):

src/main/java/net/fabricmc/loom/api/MixinApExtensionAPI.java
Logiquo: done.

modmuss50:
Great, thanks for this. 👍

PR #428, opened 2021-07-05
Add versions used to compile/build against to jar manifest

modmuss50:
TODO tests and cleanup.


src/main/java/net/fabricmc/loom/configuration/JarManifestConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.java.archives.Attributes;
+import org.gradle.jvm.tasks.Jar;
+import org.gradle.util.GradleVersion;
+
+import net.fabricmc.loom.LoomGradleExtension;
+import net.fabricmc.loom.LoomGradlePlugin;
+import net.fabricmc.loom.util.Constants;
+
+public final record JarManifestConfiguration(Project project, Jar jar) {
```
liach:
```suggestion
public record JarManifestConfiguration(Project project, Jar jar) {
```
records are automatically final.

Player3324:
the mapping namespace and maybe associated version could be interesting as well

src/main/java/net/fabricmc/loom/configuration/JarManifestConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+	public void configure() {
+		// Done in doFirst to ensure that everyone else has had time to modify the manifest
+		// and to ensure that the compile classpath has already been resolved.
+		jar.doFirst(configureManifestAction());
```
Juuxel: Doesn't doing this lazily in `doFirst` completely prevent users from customising this? I don't think you should force all jars from the `jar` task to have these attributes necessarily.

modmuss50: I wasnt too sure about this, might move it till after remapping. Will be needed anyway if we are going to include the namespace.

src/main/java/net/fabricmc/loom/configuration/JarManifestConfiguration.java
modmuss50: I will see if I can add a checkstyle rule for this. https://checkstyle.sourceforge.io/config_modifier.html#RedundantModifier might be a good idea to have.

modmuss50:
I have now moved this to after remapping.

```
Manifest-Version: 1.0
Fabric-Loom-Version: 0.0.0+unknown
Fabric-Gradle-Version: 7.0.1
Fabric-Mixin-Compile-Extensions-Version: 0.4.4
Fabric-Tiny-Remapper-Version: 0.3.2
Fabric-Loader-Version: 0.11.2
Fabric-Mixin-Version: 0.9.2+mixin.0.8.2
Fabric-Mapping-Namespace: intermediary
```

Is an example of what will be added to the manifest.

modmuss50:
This will mean we can no longer be certain that builds are reproduable across loom patch versions, I think its fine to say if you want that you should depend on a specific loom build 👍

Player3324 (approved):
LGTM, thanks for tackling this!

Player3324 (changes requested):
Actually I need a way to know whether it was fabric mixin or not (maven group?)

Player3324:
Likewise the version of the mappings, kind of plays into layered mapping territory?

modmuss50:
Sure I can add the mixin group, the actual minecraft version would be useful as well.

The mapping version might be less useful and would just be a hash when using layered mappings. I can add it but it doesn’t seem terribly useful.

Player3324 (approved):

PR #430, opened 2021-07-08
Fix local file mod dependencies

### LINKED ISSUE #56: Using mod* configurations with a flatDir repository does not deobfuscate
When `modCompile` is used to depend on a JAR from a `flatDir` repository, the dependency is not deobfuscated from intermediary names.

For example, if I remove the `-dev` classifier of a JAR dependency in my [build.gradle.kts](https://github.com/Juuxel/ThisAndThat/blob/3d691b5c1d0719456f2665523cb24b876a9d606d/build.gradle.kts#L68), this happens:

idea64_2019-02-04_21-43-15

### LINKED ISSUE #250: modRuntime doesn't work with local files
Clone of #101, title and description copied from there.

Not much else to say. No errors or warnings logged, any lines like the following are seemingly just ignored:

```groovy
modRuntime files("file/path/here.jar")
modRuntime fileTree(dir: "dir/path/here", include: "*.jar")
```

With a bit of finegling, CurseForge's automatic Maven repo is a decent substitute when trying to add local dev mods (and probably better in the long run anyways), but this is definitely a slight annoyance and may not be avoidable for other applications. `modCompile` does work with `files`, but not with `fileTree`, and obviously doesn't have the benefit of `modRuntime`'s runtime-only properties.

### LINKED ISSUE #403: modImplementation cannot accept local files
A lib I was using had problems with its maven, so, naturally, I downloaded the jar and added it with a fileTree for all jars in the libs folder. But, it silently failed to add it to my workspace
Assuming it was just fileTree, I then included it manually with `files`, but that, too, failed.
I ended up needing to use regular `implementation` to get it onto the classpath! This is stupid.
Using Loom 0.7.30, minecraft 1.16.5
Likely related to #391
Please fix when possible

Juuxel:
Fixes #56, fixes #250 and fixes #403.

- `flatDir` dependencies are fixed by replacing their empty group and version strings with `unspecified`
- `files`/`fileTree` dependencies (`FileCollectionDependency` instances) are fixed by adding custom logic as they aren't caught by `sourceConfig.getResolvedConfiguration().getResolvedArtifacts()`.
  - They don't go through resolution at all, and this is also why `DependencyProcessor` has custom logic for `SelfResolvingDependency`s like them. (I didn't use `SelfResolvingDependency` here because project dependencies are also instances of it, and they are already caught earlier since they are resolved properly. This would duplicate them.)


src/main/java/net/fabricmc/loom/build/ModCompileRemapper.java
```diff
@@ -57,6 +61,19 @@
 
 @SuppressWarnings("UnstableApiUsage")
 public class ModCompileRemapper {
+	// These are placeholders that are used when the actual group/version is missing (null or empty).
+	// This happens when the dependency is a FileCollectionDependency or from a flatDir repository.
+	private static final String MISSING_GROUP = "unspecified";
+	private static final String MISSING_VERSION = "unspecified";
```
modmuss50: Should it use the file hash, what will happen now if you replace the input with the same file? it needs to remap any changes.

src/test/groovy/net/fabricmc/loom/test/util/ProjectTestTrait.groovy
```diff
@@ -59,7 +60,7 @@ trait ProjectTestTrait {
 			}
 
 			tempFile.parentFile.mkdirs()
-			tempFile << file.text
+			Files.copy(file, tempFile)
```
modmuss50: The groovy way of doing this would be `tempFile.bytes = file.bytes`

src/main/java/net/fabricmc/loom/util/Checksum.java
```diff
@@ -60,4 +61,13 @@ public static boolean equals(File file, String checksum) {
...
 	}
+
+	public static String murmur3(File file) {
+		try {
+			HashCode hash = Files.asByteSource(file).hash(Hashing.murmur3_32());
+			return hash.toString();
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to get file hash of " + file, e);
+		}
+	}
```
modmuss50: Whats the reasoning behind murmur? I would just use something more common md5, sha1 and substring it, chances of a collison would be rare enough.

modmuss50 (commented):
Few very minor points, over all looks great, thanks for adding comprehensive tests 👍

src/main/java/net/fabricmc/loom/util/Checksum.java
Juuxel: I got it from [a StackOverflow answer](https://stackoverflow.com/a/38678199), and it was already available in Guava and didn't need any substringing. I can change it to a more common hash, though, if you want to.

src/test/groovy/net/fabricmc/loom/test/util/ProjectTestTrait.groovy
Juuxel: Sure, I'll change it. (Though if the files were a bit larger, copying would probably require less memory since it streams the content gradually.)

modmuss50: Thats true, its fine either way. Just not a massive fan using libs that I 1/2 plan on removing at somepoint.

src/main/java/net/fabricmc/loom/util/Checksum.java
Juuxel: Switched it to substringing a SHA256 hash string to ~~16~~ 12 characters.

modmuss50: Brilliant, that should be way more than enough 👍

modmuss50 (approved):

PR #431, opened 2021-07-10
Refactor LoomGradleExtension

modmuss50:
LoomGradleExtension had become a massive blob of public API and internal logic this refactor splits those two out a little.


Juuxel:
I feel like the amount of interfaces is a bit excessive - does every feature need to be its own interface? I'd split it to `LoomGradleExtensionApi` and `LoomGradleExtensionImpl`, then the base interface/class inheriting from both.

modmuss50:
I am a little unsure about the lage number of interfaces. My thinking behind it was it might make unit tests a bit easier to write in the future? However this might be wishful thinking as a lot of other things would need moving out into their own interfaces.

I think I will move them all into one as even then its still not that hard to mock out just part of it if we do get to better unit tests.

modmuss50:
Ive moved it all into 2 interfaces now, one for public facing API's and one for internal stuff. I think there is more to be done but now there is a clear picture of what is public API.

modmuss50:
I plan on merging this in, in its current state. There is still more to be done but I think this is a good step forward, and makes the public API much more clear.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -0,0 +1,81 @@
...
+
+	Dependency layered(Action<LayeredMappingSpecBuilder> action);
+
+	String getRefmapName();
+
+	void setRefmapName(String refmapName);
+
+	boolean isRemapMod();
+
+	void setRemapMod(boolean remapMod);
```
Juuxel, line 78: Note that moving from a plain `remapMod` field to a getter/setter pair like this is a breaking change for Kotlin buildscripts (the generated property name is now `isRemapMod`), but it probably doesn't hurt:
- most people don't use this
- most people don't use kts buildscripts
- 0.9 is unstable in the first place

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * This is the public api available exposed to build scripts.
+ */
+public interface LoomGradleExtensionAPI {
+	File getAccessWidener();
+
+	void setAccessWidener(Object file);
+
+	void setShareCaches(boolean shareCaches);
```
Juuxel, line 48: Maybe add a getter for this to expose it as a property to .kts buildscripts (Kotlin properties must have a getter)?

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -0,0 +1,81 @@
...
+import net.fabricmc.loom.configuration.processors.JarProcessor;
+import net.fabricmc.loom.configuration.providers.mappings.LayeredMappingSpecBuilder;
+
+/**
+ * This is the public api available exposed to build scripts.
+ */
+public interface LoomGradleExtensionAPI {
+	File getAccessWidener();
+
+	void setAccessWidener(Object file);
```
Juuxel, line 46: I'm not 100% sure off the top of my head whether Kotlin recognises this as a property setter for `accessWidener`, so that's probably worth checking out later. (If not, .kts scripts will have to call setAccessWidener manually or Loom will have to add a File overload, nothing major)

Juuxel (approved):

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
modmuss50: I think this is fine, mostly becuase I dont know of any case where this is even used.

I expect there might be stuff that is missed from this, we can easily evaulate it and add it back if needed.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
modmuss50: I think there used to be a file overload that was depreated, this might explain why. Easy enough to add back if needed.

PR #439, opened 2021-07-17
Allow developer set the site where to download game resource.

Krysztal112233:
Hi.
In China or elsewhere, it's a pain for developers to download game resource files. This Pull Request adds a property for developers to modify where loom should go to download game resource files.

But I'm not sure it makes sense to configure the download address in gradle.properties, maybe it needs to be configured elsewhere?

The pull request can now be configured to download the address.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
@@ -39,6 +39,7 @@ import net.fabricmc.loom.decompilers.DecompilerConfiguration;
 import net.fabricmc.loom.task.LoomTasks;
 
 public class LoomGradlePlugin implements Plugin<Project> {
+	public static Project project;
 	public static boolean refreshDeps;
 	public static final Gson GSON = new GsonBuilder().setPrettyPrinting().create();
 
@@ -47,6 +48,7 @@ public class LoomGradlePlugin implements Plugin<Project> {
 		project.getLogger().lifecycle("Fabric Loom: " + LoomGradlePlugin.class.getPackage().getImplementationVersion());
 
 		refreshDeps = project.getGradle().getStartParameter().isRefreshDependencies();
+		LoomGradlePlugin.project = project;
 
 		if (refreshDeps) {
 			MappingsCache.INSTANCE.invalidate();
--- a/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
@@ -30,6 +30,7 @@ import com.google.gson.GsonBuilder;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 
+import net.fabricmc.loom.configuration.MirrorConfiguration;
 import net.fabricmc.loom.configuration.CompileConfiguration;
 import net.fabricmc.loom.configuration.FabricApiExtension;
 import net.fabricmc.loom.configuration.MavenPublication;
@@ -39,7 +40,6 @@ import net.fabricmc.loom.decompilers.DecompilerConfiguration;
 import net.fabricmc.loom.task.LoomTasks;
 
 public class LoomGradlePlugin implements Plugin<Project> {
-	public static Project project;
 	public static boolean refreshDeps;
 	public static final Gson GSON = new GsonBuilder().setPrettyPrinting().create();
 
@@ -48,7 +48,6 @@ public class LoomGradlePlugin implements Plugin<Project> {
 		project.getLogger().lifecycle("Fabric Loom: " + LoomGradlePlugin.class.getPackage().getImplementationVersion());
 
 		refreshDeps = project.getGradle().getStartParameter().isRefreshDependencies();
-		LoomGradlePlugin.project = project;
 
 		if (refreshDeps) {
 			MappingsCache.INSTANCE.invalidate();
@@ -65,6 +64,7 @@ public class LoomGradlePlugin implements Plugin<Project> {
 		project.getExtensions().add("loom", project.getExtensions().getByName("minecraft"));
 		project.getExtensions().create("fabricApi", FabricApiExtension.class, project);
 
+		MirrorConfiguration.setup(project);
 		CompileConfiguration.setupConfigurations(project);
 		IdeConfiguration.setup(project);
 		CompileConfiguration.configureCompile(project);
--- a/src/main/java/net/fabricmc/loom/configuration/MirrorConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/MirrorConfiguration.java
@@ -1,0 +1,27 @@
+package net.fabricmc.loom.configuration;
+
+import net.fabricmc.loom.util.Constants;
+import net.fabricmc.loom.util.Mirrors;
+
+import org.gradle.api.Project;
+
+public class MirrorConfiguration {
+	public static void setup(Project project){
+		if (!(project.hasProperty("loom_libraries_base")
+				|| project.hasProperty("loom_resources_base")
+		        || project.hasProperty("version_manifests"))){
+			return;
+		}
+
+		String LIBRARIES_BASE_MIRRORS = project.hasProperty("loom_libraries_base")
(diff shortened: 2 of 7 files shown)


src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
```diff
@@ -39,6 +39,7 @@
 import net.fabricmc.loom.task.LoomTasks;
 
 public class LoomGradlePlugin implements Plugin<Project> {
+	public static Project project;
```
modmuss50: No, this should not be static.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftVersionMeta.java
```diff
@@ -208,6 +209,10 @@ public Artifact getArtifact() {
 	}
 
 	public final class Artifact extends Downloadable {
+		@Override
+		public String getUrl() {
+			return Constants.LIBRARIES_BASE + super.getPath();
+		}
```
modmuss50: No, this is not good, the url from the version meta should be whats used.

src/main/java/net/fabricmc/loom/util/Constants.java
```diff
@@ -30,12 +30,15 @@
...
 import net.fabricmc.loom.configuration.RemappedConfigurationEntry;
 import net.fabricmc.loom.util.gradle.GradleSupport;
 
 public class Constants {
-	public static final String LIBRARIES_BASE = "https://libraries.minecraft.net/";
-	public static final String RESOURCES_BASE = "http://resources.download.minecraft.net/";
+	public static String LIBRARIES_BASE = LoomGradlePlugin.project.getProperties().get("libraries_base") == null ?
+			"https://libraries.minecraft.net/" : String.valueOf(LoomGradlePlugin.project.getProperties().get("LIBRARIES_BASE"));
+	public static String RESOURCES_BASE = LoomGradlePlugin.project.getProperties().get("resources_base") == null ?
+			"http://resources.download.minecraft.net/" : String.valueOf(LoomGradlePlugin.project.getProperties().get("RESOURCES_BASE"));
```
modmuss50: Leave these constants alone, and use the replacement where they are used.

src/main/java/net/fabricmc/loom/util/Constants.java
```diff
@@ -30,12 +30,15 @@
...
 import org.objectweb.asm.Opcodes;
 
+import net.fabricmc.loom.LoomGradlePlugin;
 import net.fabricmc.loom.configuration.RemappedConfigurationEntry;
 import net.fabricmc.loom.util.gradle.GradleSupport;
 
 public class Constants {
-	public static final String LIBRARIES_BASE = "https://libraries.minecraft.net/";
-	public static final String RESOURCES_BASE = "http://resources.download.minecraft.net/";
+	public static String LIBRARIES_BASE = LoomGradlePlugin.project.getProperties().get("libraries_base") == null ?
```
modmuss50: should be `loom_libraries_base`

modmuss50 (changes requested):
Beyond the changes not being correct im a little uneasy about doing this as I belive China has its own version of minecraft for a reason, thus you should be modding that.

Im not sure if I will merge this in as I dont want to be going agasint the wishes of mojang or local laws.

drHuangMHT:
> Beyond the changes not being correct im a little uneasy about doing this as I belive China has its own version of minecraft for a reason, thus you should be modding that.
> 
> Im not sure if I will merge this in as I dont want to be going agasint the wishes of mojang or local laws.

Hi there, I'm also a Chinese fabric mod developer. In fact, due to some network restrictions implemented by China authorities, accessing web services hosted outside China is really slow, sometimes even completely blocked. But when developing fabric mods you often need to fetch resource files. That slow speed really makes developing a disaster. So making file source changeable is really important for many Chinese developers. I would really appreciate it if that is made possible. If you are interested in Minecraft China, please let me know and I will try to explain it to you.

3TUSK:
> China has its own version of minecraft for a reason, thus you should be modding that.

Clarification. Technical-wise, they *are* the Java/Bedrock Editions, (IIRC) with runtime mods for moderation purposes. There isn't a different "Chinese version" of the game *software* to mod with. This is also why there are modders who can write Forge-based mods and put their work(s) on both CurseForge and NetEase. I am not sure about NetEase's Fabric support tho. 

Not to mention that Chinese modded minecraft scene has existed long before Mojang/Microsoft's partnership with NetEase.

ZekerZhayard:
Maybe @liach knows many stories about how hard to configure a dev env in China.

src/main/java/net/fabricmc/loom/configuration/MirrorConfiguration.java
```diff
@@ -0,0 +1,27 @@
+package net.fabricmc.loom.configuration;
+
+import net.fabricmc.loom.util.Constants;
+import net.fabricmc.loom.util.Mirrors;
+
+import org.gradle.api.Project;
+
+public class MirrorConfiguration {
```
liach, line 32: Should we make this as an extension or just a property? For users in China, if they do need to connect to say alibaba's mirror server, they currently specify it in `build.gradle`'s `repositories` block than passing in some properties.

If this is an extension named like `loomMirror`, you can still use it like:
```gradle
loomMirror {
  libraries_base = project.loom_libraries_base // a string custom base
  resources_base = project.loom_resources_base // another string custom base
  version_manifests = uri(project.version_manifests) // parsing the manifests into uri
}
```
and obtain the same behavior

Logiquo: I think it makes more sense to put it in the property file or setting file instead of the build file. This thing does not change the building behavior, and hence should not be part of the build file.

liach: Yeah, this is fine. I don't think there is any other concerns with using project property.

src/main/java/net/fabricmc/loom/util/Mirrors.java
```diff
@@ -0,0 +1,19 @@
...
+	public static String VERSION_MANIFESTS;
+
+	public static void changeMirror(String lib,String res,String manifest){
+		LIBRARIES_BASE_MIRRORS = lib;
+		RESOURCES_BASE_MIRRORS = res;
+		VERSION_MANIFESTS = manifest;
+	}
+
+	public static boolean hasMirror(){
+		return ! (LIBRARIES_BASE_MIRRORS.isEmpty()
```
liach: Won't this crash with an NPE since the fields are `null` (`changeMirror` not called) when none of the properties are given?

I recommend creating 3 methods `getLibrariesBase` `getResourcesBase` `getVersionManifests` than using the fields directly.

modmuss50:
Can you please target the 0.9 branch.

Krysztal112233:
> 
> 
> Can you please target the 0.9 branch.

Okay, we'll finish it in the next few days on 0.9 branch.

liach:
In fact, for 0.9, the code would change quite a bit, as the repository declaration has been moved earlier per gradle recommendations. You would have to patch https://github.com/FabricMC/fabric-loom/blob/dev/0.9/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java instead

Krysztal112233:
> 
> 
> In fact, for 0.9, the code would change quite a bit, as the repository declaration has been moved earlier per gradle recommendations. You would have to patch https://github.com/FabricMC/fabric-loom/blob/dev/0.9/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java instead

Yes, we are aware of the problem. But we observe that many Chinese developers are still using 0.7 version of loom, so can this feature be added in 0.7 as well?

Logiquo:
I think modmuss50 decides 0.7 will be LTS, so this should be possible? But it really depends on the opinion of modmuss50.

modmuss50:
Let’s make sure to get it into 0.9 first then it can be backported as required.

Otherwise there is a chance it will only ever be in 0.7. 0.9 is currently the unstable testing version so it’s a great place to target without causing people issues.

modmuss50:
Sorry I do not plan to backport this PR, there should be little reason why you cannot use the latest loom version.

PR #441, opened 2021-07-19
Use tiny-remapper to remap mixin annotation.

Logiquo:
~~This is pending https://github.com/FabricMC/tiny-remapper/pull/63 and I know the current implementation is far from correct. Need some input on how to correctly do it.~~

~~Pending release of tiny-remapper of 0.5.0~~

start working on it now.


Logiquo:
Hi @modmuss50. Since the mixin extension is merged into tiny-remapper (though still pending release to maven), I want to discuss how this would work with loom. There are two problems need to discuss.

1. where do you want to put the option to enable/disable it. What I am currently thinking is to have
```
loom {
  useLegacyMixinAp = true; // do you want default to be true or false?
  mixinAp { // was "mixin" in the current version. Throw exception if "useLegacyMixinAp" is false.
  }
}
```
do you have any input on this?

2. the `sponge-mixin-0.9.4+mixin.0.8.2.jar` (maybe also other versions, haven't look at them yet) ships with `javax.annotation.processing.Processor`, which cause java compiler to try to start the mixin Ap regardless we set it or not. And even worse, since the related class file is not there, it will throw an exception. Is it possible to update these mixin jars without `javax.annotation.processing.Processor` service file?

modmuss50:
Would be nice to enable this for everyone, but we will want to start rolling it out slowly. Loom 0.9 is basically done at this point so we could target 0.10 for this with it enabled for everyone.

2. If I understand what you are saying, not adding mixin to the `annotationProcessor` config is still causing it to try and load mixin when compling. That does sound really annoying. I will have to take a look myself at this as ideally we can contiune to support the mixin AP for a little bit. Worse case could loom add a dummy mixin AP that just doesnt crash? Idk without looking into it.

Logiquo:
> If I understand what you are saying, not adding mixin to the annotationProcessor config is still causing it to try and load mixin when compling.

Yes, that's what I mean. With a closer look at the loom, I think this causes the problem.

https://github.com/FabricMC/fabric-loom/blob/dev/0.9/src/main/java/net/fabricmc/loom/configuration/LoomDependencyManager.java#L193-L205

What's the reason we want to put these dependencies in the ap config? I guess we should be able to delete it just fine, as the `fabric-mixin-extension` will pull mixin ap anyway.

modmuss50:
Thats used to add deps (mainly asm I think) onto the classpath for mce/mixin. That wont be needed if mixin AP's is not being used 👍 will be good to get rid of that.

Logiquo:
Just curious, shouldn't the `processor` version of mixin have all those dependencies available? Why does the `fabric-mixin-extension` pulls the `mixin` and add these dependencies manually? I think it can pull the `mixin-processor` directly and avoid adding these dependencies.

Logiquo:
This one is ready to review.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -153,10 +155,10 @@ default void setRemapMod(boolean remapMod) {
 
 	NamedDomainObjectContainer<RunConfigSettings> getRunConfigs();
 
-	@ApiStatus.Experimental
-	void mixin(Action<MixinApExtensionAPI> action);
+	void mixinAp(Action<MixinApExtensionAPI> action);
```
modmuss50: Why the rename?

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -67,6 +67,8 @@ default void setAccessWidener(File file) {
 
 	Property<Boolean> getShareRemapCaches();
 
+	Property<Boolean> getUseLegacyMixinAp();
```
modmuss50: I would put this in the MixinApExtensionAPI, possibly rename `MixinApExtensionAPI` to `MixinExtensionAPI` makes more sense to me to put all mixin related options under that.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -133,25 +133,27 @@ public static void configureCompile(Project p) {
...
-
-			if (project.getPluginManager().hasPlugin("scala")) {
-				project.getLogger().info("Configuring compiler arguments for Scala");
-				new ScalaApInvoker(project).configureMixin();
-			}
-
-			if (project.getPluginManager().hasPlugin("org.jetbrains.kotlin.kapt")) {
-				project.getLogger().info("Configuring compiler arguments for Kapt plugin");
-				new KaptApInvoker(project).configureMixin();
+			if (extension.getUseLegacyMixinAp().get()) {
```
modmuss50:
```
if (!extension.getUseLegacyMixinAp().get()) {
   return
}
```

Would be a little neater, this is fine tho.

modmuss50 (changes requested):

modmuss50:
Please also target the exp/0.10 branch. We can possiblly look at backporting this disabled by default as it should be quite safe.

Logiquo:
fixed all review items.

modmuss50:
Where are we at with this? Does this depend on https://github.com/FabricMC/tiny-remapper/pull/75 and loader changes? Or is that a seprate thing?

modmuss50 (approved):

Logiquo:
This one is good to go. FabricMC/tiny-remapper#75 is a separate thing and won't affect this (It's to solve https://github.com/FabricMC/fabric/commit/2e359e933a5a3e188c90c14f286d578f2aad290c and I'm still working on it).

PR #442, opened 2021-07-21
Allow developer set the site where to download game resource.(0.9 branch)

hanbings:
We have roughly completed the mirror configuration file support for the 0.9 branch, and we will complete the 0.7 branch later

There are still some doubts about the 0.9 branch I want to ask: notice that there are some strings in the LoomRepositoryPlugin of the 0.9 branch that do not quote Constants, for example [LoomRepositoryPlugin.java#L76](https://github.com/FabricMC/fabric-loom/blob/dev/0.9/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java#L76) 
Do I need to change them to use MirrorUtil like in other code?


src/main/java/net/fabricmc/loom/util/MirrorUtil.java
```diff
@@ -0,0 +1,70 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.util;
+
+public class MirrorUtil {
```
shedaniel, line 29: This should be configured per project imo

modmuss50:
The urls around https://github.com/FabricMC/fabric-loom/blob/dev/0.9/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java#L76 should be in the constants, and should be overrideable as well.

README.md
```diff
@@ -1,32 +1,39 @@
...
+    * Change the loom version to `0.6.local`. For example `id 'fabric-loom' version '0.6.local'`
 4. Create a Gradle run configuration:
-   * Set the Gradle project path to the project you have just configured above
-   * Set some tasks to run, such as `clean build` you can change these to suit your needs.
-   * Add the run configuration you created earlier to the "Before Launch" section to rebuild loom each time you debug
+    * Set the Gradle project path to the project you have just configured above
+    * Set some tasks to run, such as `clean build` you can change these to suit your needs.
+    * Add the run configuration you created earlier to the "Before Launch" section to rebuild loom each time you debug
 5. You should now be able to run the configuration in debug mode, with working breakpoints.
+
```
Juuxel: I don't think these README reformatting changes are really useful, are you planning to change the actual content of it?

Krysztal112233: Ah I was careless with the formatting here, sorry about that.

I'll roll it back.

hanbings:
The commit of incorrect readme changes has been cleaned up. Since Loom RepositoryPlugin is called earlier in [LoomGradlePlugin.java#L57](https://github.com/FabricMCCN/fabric-loom-cn/blob/dev/0.9/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java#L57) than [LoomGradlePlugin.java#L85](https://github.com/FabricMCCN/fabric-loom-cn/blob/dev/0.9/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java#L85) using MirrorUtil to obtain resource links in LoomRepositoryPlugin will cause NPE. How can I change their calling sequence?

modmuss50:
I would keep it very simple, just have a class with a few static methods similar to this:

```java
public static String getVersionManifests(Project project) {
	if (project.hasProperty("loom_version_manifests")) {
		return String.valueOf(project.property("loom_version_manifests"));
	}

	return Constants.VERSION_MANIFESTS;
}
```

However this wont work for PluginAware. Using a java poperty might be required as I cannot see a way to get the project pops with just a PluginAware (Am I missing something?) Java props can still be set by the settings.gradle file. If you do that it would be easier to populate the existing constant values as well.

shedaniel:
Any reason to go with properties instead of using values in the extension?

modmuss50:
This should be globally and not be per project. It’s also needed before the extension is ready.

Logiquo:
> Dynamic Properties
> In addition to the properties of this interface, the Settings object makes some additional read-only properties available to the > settings script. This includes properties from the following sources:
> 
> Defined in the gradle.properties file located in the settings directory of the build.
> Defined the gradle.properties file located in the user's .gradle directory.
> Provided on the command-line using the -P option.

Based on the documentation of Gradle, it seems possible to access data from the property.

hanbings:
> I would keep it very simple, just have a class with a few static methods similar to this:
> 
> ```java
> public static String getVersionManifests(Project project) {
> 	if (project.hasProperty("loom_version_manifests")) {
> 		return String.valueOf(project.property("loom_version_manifests"));
> 	}
> 
> 	return Constants.VERSION_MANIFESTS;
> }
> ```
> 
> However this wont work for PluginAware. Using a java poperty might be required as I cannot see a way to get the project pops with just a PluginAware (Am I missing something?) Java props can still be set by the settings.gradle file. If you do that it would be easier to populate the existing constant values as well.

I'm sorry to reply to you after such a long time. The newly submitted code has deleted the MirrorConfigcuation class, and now it only uses a MirrorUtil class to complete the work. The MirrorConfigcaution class was written on the 0.7 branch, because some places where Constants constant url was called were not passed the Project instance, which made it impossible to use the above methods to obtain the configuration, and the class that used the url in the 0.9 branch has been correctly passed the Project instance. I'm sorry we have these heavy code from the old version to the new version directly

Logiquo:
I've experimented locally, it's entirely possible to read the value of `gradle.properties` from `gradle.settings`.

Just do 
```Java
this.getSettings().getExtensions().getExtraProperties().get​(String name)
```

Please notice you need to use `has` method to check first, otherwise `ExtraPropertiesExtension.UnknownPropertyException` might be thrown.
```

**Comment by hanbings** [comment]
> I would keep it very simple, just have a class with a few static methods similar to this:
> 
> ```java
> public static String getVersionManifests(Project project) {
> 	if (project.hasProperty("loom_version_manifests")) {
> 		return String.valueOf(project.property("loom_version_manifests"));
> 	}
> 
> 	return Constants.VERSION_MANIFESTS;
> }
> ```
> 
> However this wont work for PluginAware. Using a java poperty might be required as I cannot see a way to get the project pops with just a PluginAware (Am I missing something?) Java props can still be set by the settings.gradle file. If you do that it would be easier to populate the existing constant values as well.

When the LoomRepositoryPlugin gets the Setting, we can use settings.getGradle().getRootProject() to get the Project instance.
Correct the code format and test, it can work

File: src/main/java/net/fabricmc/loom/util/MirrorUtil.java (hunk 1 above)
Comment (Logiquo) [reply src/main/java/net/fabricmc/loom/util/MirrorUtil.java:29]: I can hardly believe someone will want to have a different mirror for each sub-project.

**Review by Logiquo — APPROVED** [review]

File: src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java [hunk 3]
```diff
@@ -69,11 +76,11 @@ public void apply(PluginAware target) {
 		});
 		repositories.maven(repo -> {
 			repo.setName("Fabric");
-			repo.setUrl("https://maven.fabricmc.net/");
+			repo.setUrl(MirrorUtil.getFabricRepository(finalAware));
```
Logiquo:
```suggestion
			repo.setUrl(MirrorUtil.getFabricRepository((ExtensionAware) target);
```
This should be simpler.

hanbings: Yes, thanks for the reminder, the new commit has completed the changes.

Krysztal112233:
I think this PR is ready. Mind having a look ? @modmuss50

modmuss50 (approved):
Looks fine, could be cleaned up a little but dont worry about that.

I dont think there is a nice way to write an intergration test for this, so as long as the existing ones pass I dont see any reason not to go with this.

Krysztal112233:
We had some problems with our actions and the test timed out. It's not a problem with our code, it's a timeout caused by it queuing for more than 24 hours.
Also we noticed that development has now started on the 0.10 branch, do we need to move to the 0.10 branch?
@modmuss50

Logiquo:
Any update for this PR? Seems long enough to bump it again.

PR #443, opened 2021-07-21
Fix removing of jar processors not marking the file as invalid.

modmuss50:
I might just store a list of the jar processors used instead of breaking the api. Feedback is welcome.


modmuss50:
https://github.com/FabricMC/fabric-loom/pull/444 is a better fix.

PR #444, opened 2021-07-22
Use jar processor ids to fix #432

### LINKED ISSUE #432: Removing jar processors does not regenerate the project mapped jar
Loom only checks if the current jar processors are up to date, and not if the jar processor list is now shorter.

Juuxel:


src/main/java/net/fabricmc/loom/configuration/processors/JarProcessorManager.java
```diff
@@ -47,13 +65,56 @@ public boolean isInvalid(File file) {
...
+		String jarProcessorIds = jarProcessors.stream()
+				.map(JarProcessor::getId)
+				.sorted()
+				.collect(Collectors.joining(";"));
+
+		try {
+			return CharSource.wrap(jarProcessorIds)
+					.asByteSource(StandardCharsets.UTF_8)
+					.hash(Hashing.sha256())
+					.toString().substring(0, 12);
```
Juuxel: Hmm, might as well use the full 64-character hash here as a few extra bytes don't matter in the manifest (and there's no need to substring then)

Juuxel:
```suggestion
					.toString();
```

modmuss50 (approved):
Doesnt really need hashing, but this is fine 👍

PR #445, opened 2021-07-24
Use Property in LoomGradleExtension & Move task groups to constants

shedaniel:
Replacing the getters and setters is quite annoying since we have to choose a different name for the methods.

Using properties allow us to finalize and freeze values that should not be changed after evaluation (see finalizeValue)

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -107,8 +107,6 @@ public interface LoomGradleExtensionAPI {
 		return getGameJarProcessors().get();
 	}
 
-	@Deprecated(forRemoval = true)
-	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
 	default void addJarProcessor(JarProcessor processor) {
 		getDeprecationHelper().replaceWithInLoom0_11("jarProcessors", "gameJarProcessors");
 		getGameJarProcessors().add(processor);
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -32,7 +32,6 @@ import org.gradle.api.NamedDomainObjectContainer;
 import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.RegularFileProperty;
-import org.gradle.api.model.ReplacedBy;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
 import org.jetbrains.annotations.ApiStatus;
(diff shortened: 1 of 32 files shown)


modmuss50:
Awesome 👍 This was the next thing I had planned for this. Will take a propper look later.

Wonder if it would be worth to show a message when using a deprecated method similar to gradles? Idk if gradle has an api for that that we could re-use.

Do any of the tests need updating?

shedaniel:
Gradle's deprecation setup is unfortunately internal: https://github.com/gradle/gradle/blob/master/subprojects/logging/src/main/java/org/gradle/internal/deprecation/DeprecationLogger.java

The tests should all be fine.

modmuss50:
Should be easy enough to impliment some basic sort of warning.

Gradle does Expose the warning mode via: `Project.getGradle().getStartParameter().getWarningMode()`. Fail and None are just the two i'd worry about. Should be easy enough to log or crash with that.  Idk how long we want to keep it deprecated? Remove in 0.11?

src/main/java/net/fabricmc/loom/configuration/accesswidener/AccessWidenerJarProcessor.java
```diff
@@ -77,14 +77,15 @@ public String getId() {
 	@Override
 	public void setup() {
 		LoomGradleExtension loomGradleExtension = LoomGradleExtension.get(project);
+		File awPath = loomGradleExtension.getAccessWidenerPath().getAsFile().get();
```
Juuxel: Haven't tested this, but `get().getAsFile()` might be a bit more efficient since afaik there's no intermediate provider instance created then.

src/test/resources/projects/dependencyResolutionManagement/projmap/build.gradle
```diff
@@ -69,7 +69,7 @@ jar {
 }
 
 minecraft {
-	accessWidener = file("src/main/resources/modid.accesswidener")
+	accessWidener.set file("src/main/resources/modid.accesswidener")
```
Juuxel: Fyi, `=` works with properties in the groovy dsl. Also, shouldn't this be `accessWidenerPath`?

src/main/java/net/fabricmc/loom/util/DeprecationHelper.java
```diff
@@ -0,0 +1,86 @@
...
+	class ProjectBased implements DeprecationHelper {
+		private final Project project;
+		private final AtomicBoolean usingDeprecatedApi = new AtomicBoolean(false);
+
+		public ProjectBased(Project project) {
+			this.project = project;
+
+			project.getGradle().buildFinished(buildResult -> {
+				if (usingDeprecatedApi.get()) {
+					project.getLogger().lifecycle("Deprecated Loom APIs were used in this build, making it incompatible with future versions of loom. "
```
Juuxel:
```suggestion
					project.getLogger().lifecycle("Deprecated Loom APIs were used in this build, making it incompatible with future versions of Loom. "
```

modmuss50:
Change looks good, seems a few of the intergration tests are failing now. (they are all ran with warn-mode to set to fail)

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -31,36 +31,90 @@
...
+	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
+	default List<LoomDecompiler> getDecompilers() {
+		getDeprecationHelper().replaceWithInLoom0_11("decompilers", "loomDecompilers");
+		return getLoomDecompilers().get();
+	}
 
-	List<JarProcessor> getJarProcessors();
+	@Deprecated(forRemoval = true)
+	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
+	default void addDecompiler(LoomDecompiler decompiler) {
```
Juuxel, line 97: This and `shareCaches` are probably fine to keep on their own since they're utility methods and not direct getters or setters.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -31,36 +31,90 @@
...
+	}
 
+	@Deprecated(forRemoval = true)
+	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
 	default void shareCaches() {
 		setShareCaches(true);
 	}
 
-	List<LoomDecompiler> getDecompilers();
+	ListProperty<LoomDecompiler> getLoomDecompilers();
```
Juuxel: Maybe `gameDecompilers` or `minecraftDecompilers`? `loom { loomDecompilers.add(...) }` is a bit awkward

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -31,36 +31,90 @@
...
 
 /**
  * This is the public api available exposed to build scripts.
  */
 public interface LoomGradleExtensionAPI {
-	File getAccessWidener();
+	@ApiStatus.Internal
+	DeprecationHelper getDeprecationHelper();
 
+	RegularFileProperty getAccessWidenerPath();
```
Juuxel, line 52: Re-introducing `accessWidener` as an alias for this in the future would be nice, I think. (though not necessary)

Juuxel (commented):
A few more comments, great PR though (`Property` for tasks next? :smile:)

src/test/resources/projects/mixinApAutoRefmap/build.gradle
```diff
@@ -68,7 +68,7 @@ java {
 }
 
 loom {
-	refmapName = "refmap0000.json"
+	mixin.defaultRefmapName = "refmap0000.json"
 
 	mixin {
```
Juuxel:
```suggestion
	mixin {
		defaultRefmapName = "refmap0000.json"
```

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -31,36 +31,88 @@
...
+	@Deprecated(forRemoval = true)
+	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
+	default List<JarProcessor> getJarProcessors() {
+		getDeprecationHelper().replaceWithInLoom0_11("jarProcessors", "gameJarProcessors");
+		return getGameJarProcessors().get();
+	}
+
+	@Deprecated(forRemoval = true)
+	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
+	default void addJarProcessor(JarProcessor processor) {
```
Juuxel, line 110: The comment about utility methods also applies here (sorry, didn't notice this before) but it's really just a matter of opinion whether these are kept.

Juuxel (approved):

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -31,36 +31,86 @@
...
-	void addJarProcessor(JarProcessor processor);
+	@Deprecated(forRemoval = true)
+	@ApiStatus.ScheduledForRemoval(inVersion = "0.11")
+	default List<JarProcessor> getJarProcessors() {
+		getDeprecationHelper().replaceWithInLoom0_11("jarProcessors", "gameJarProcessors");
+		return getGameJarProcessors().get();
+	}
+
+	default void addJarProcessor(JarProcessor processor) {
+		getDeprecationHelper().replaceWithInLoom0_11("jarProcessors", "gameJarProcessors");
```
Juuxel:
```suggestion
```

Juuxel (changes requested):

modmuss50 (approved):

PR #446, opened 2021-07-24
Make SourceRemapper threaded

modmuss50:
WIP: needs testing, and need to see if its actually faster.


modmuss50:
When testing on my MBP this now takes 4.1s vs 9.8s, not a massive improvment but better than nothing. Just need to fix the logging and then this should be good.

PR #447, opened 2021-07-26
Fix loom.accessWidener being a read-only property in Kotlin

Juuxel:
The setter with an `Object` parameter isn't considered the setter of the `accessWidener` property in Kotlin (understandably), so this PR changes the parameter type to `File`.

### DIFF
--- a/src/test/groovy/net/fabricmc/loom/test/integration/KotlinTest.groovy
+++ b/src/test/groovy/net/fabricmc/loom/test/integration/KotlinTest.groovy
@@ -31,7 +31,6 @@ import spock.lang.Unroll
 
 import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
 
-@IgnoreIf({ jvm.java16Compatible }) // Fails on J16 > due to https://youtrack.jetbrains.com/issue/KT-45566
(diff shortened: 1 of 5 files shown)


src/test/groovy/net/fabricmc/loom/test/integration/KotlinTest.groovy
```diff
@@ -31,7 +31,6 @@ import spock.lang.Unroll
 
 import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
 
-@IgnoreIf({ jvm.java16Compatible }) // Fails on J16 > due to https://youtrack.jetbrains.com/issue/KT-45566
```
modmuss50, line 34: Also need to remove this:

https://github.com/FabricMC/fabric-loom/blob/dev/0.9/.github/workflows/test-push.yml#L48-L50

and

https://github.com/FabricMC/fabric-loom/blob/dev/0.9/.github/workflows/test-push.yml#L77-L79

Juuxel: Done.

modmuss50 (approved):

modmuss50:
Thanks

PR #451, opened 2021-07-27
fix #448

### LINKED ISSUE #448: Automatic refmaps aren't added to mixin jsons if the name is not *.mixins.json
Before #423, mixin jsons were scanned by their contents (whether they contained `mixins`, `client` or `server`). This allowed them to have pretty much any name that doesn't match the most popular `*.mixins.json`. Some Fabric mods, like [Adorn](https://github.com/Juuxel/Adorn/blob/1.16/src/main/resources/mixins.adorn.json) or [Litematica](https://github.com/maruohon/litematica/blob/master/src/main/resources/mixins.litematica.json) use prefix notation for their mixins instead. (For example, due to the MCDev IDEA plugin not supporting `*.mixins.json` in the past)

If a mixin JSON still uses this naming or any other naming pattern that does not end in `.mixins.json`, the `refmap` property isn't automatically added to the mixin json, since `*.mixins.json` is hardcoded in Loom:
https://github.com/FabricMC/fabric-loom/blob/b558ee1a4604a0fbf0eba8edab401f2a831e4fb6/src/main/java/net/fabricmc/loom/extension/MixinApExtensionImpl.java#L62

There are three possibly fixes to this:
1. Reading the mixin json paths by the contents. This is cursed and not really efficient, either. (It reads all top-level json files in the jar)
2. Reading the mixin json paths from a fabric.mod.json (possibly in addition to a pattern). This is the most reliable way to catch the mixin jsons.
3. Switching to a more sophisticated pattern, like in modern MCDev (minecraft-dev/MinecraftDev#691)

Logiquo:
Initial attempt. Add a new test case for this. Waiting for test results from the workflow.


src/main/java/net/fabricmc/loom/extension/MixinApExtensionImpl.java
```diff
@@ -79,7 +78,7 @@ private String getDefaultMixinRefmapName() {
 
 	@Override
 	protected PatternSet add0(SourceSet sourceSet, Provider<String> refmapName) {
-		PatternSet pattern = new PatternSet().setIncludes(Collections.singletonList("*.mixins.json"));
+		PatternSet pattern = new PatternSet().setIncludes(Collections.singletonList("*.json"));		// TODO: should I use `*.json` or `*`?
```
Logiquo: Need input for this @modmuss50 @Juuxel

Juuxel: `*.json` is fine imo, the files can't be anything else.

src/main/java/net/fabricmc/loom/build/MixinRefmapHelper.java
```diff
@@ -43,29 +51,72 @@
...
+		return StreamSupport.stream(fabricModJson.getAsJsonArray("mixins").spliterator(), false)
+				.map(e -> {
+					if (e instanceof JsonPrimitive str) {
+						return str.getAsString();
+					} else if (e instanceof JsonObject obj) {
+						return obj.get("config").getAsString();
+					} else {
+						throw new RuntimeException("Incorrect fabric.mod.json format");
+					}
+				}).collect(Collectors.toList());
```
Juuxel:
```suggestion
				}).collect(Collectors.toSet());
```
Might as well use a set here since duplicates don't matter (though I don't know how loader behaves with them)

Juuxel (approved):

Logiquo:
fixed.

modmuss50 (approved):
Great 👍 thanks.

PR #454, opened 2021-07-28
Add a tiny mappings layer

Juuxel:
- [x] Basic functionality
- [ ] Integration tests
- [ ] Is the naming good? (Do we want to expose the format in the API, or should it have some more general name?)

I also updated mapping-io to 0.1.8 since earlier versions corrupted data when loading Yarn.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingLayer.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingLayer.java
@@ -31,9 +31,10 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.util.List;
-import java.util.Locale;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
+
+import org.jetbrains.annotations.Nullable;
 
 import net.fabricmc.loom.configuration.providers.mappings.MappingLayer;
 import net.fabricmc.loom.configuration.providers.mappings.MappingNamespace;
@@ -42,18 +43,16 @@ import net.fabricmc.mappingio.MappingVisitor;
 import net.fabricmc.mappingio.adapter.MappingSourceNsSwitch;
 import net.fabricmc.mappingio.format.Tiny2Reader;
 
-public record TinyMappingLayer(File mappingFile) implements MappingLayer {
-	private static final String MAPPING_PATH = "mappings/mappings.tiny";
-
+public record TinyMappingLayer(File mappingFile, @Nullable String mappingPath) implements MappingLayer {
 	@Override
 	public void visit(MappingVisitor mappingVisitor) throws IOException {
-		if (mappingFile.getName().toLowerCase(Locale.ROOT).endsWith(".tiny")) { // bare tiny file
+		if (mappingPath == null) { // bare file
 			try (Reader reader = new FileReader(mappingFile)) {
 				read(mappingVisitor, reader);
 			}
 		} else { // assume a zip/jar
 			try (ZipFile zip = new ZipFile(mappingFile)) {
-				ZipEntry entry = zip.getEntry(MAPPING_PATH);
+				ZipEntry entry = zip.getEntry(mappingPath);
 
 				if (entry == null) {
 					throw new IOException("Could not find mappings inside " + mappingFile);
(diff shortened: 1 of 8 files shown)


src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingsSpec.java
```diff
@@ -0,0 +1,35 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.configuration.providers.mappings.tiny;
+
+import net.fabricmc.loom.configuration.providers.mappings.MappingContext;
+import net.fabricmc.loom.configuration.providers.mappings.MappingsSpec;
+
+public record TinyMappingsSpec(String dependencyNotation) implements MappingsSpec<TinyMappingLayer> {
```
modmuss50: Im not a fan of this being a string, imo it should be some sort of file resolver that can handle maven deps and local files. With local files it's hashcode should be calulated using the file's hash so it will regen when the local file changes.

I did start looking into this but got a bit burnt out for what ever reason: https://github.com/FabricMC/fabric-loom/compare/dev/0.9...modmuss50:mappingiolayer

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingLayer.java
```diff
@@ -0,0 +1,84 @@
...
+
+import net.fabricmc.loom.configuration.providers.mappings.MappingLayer;
+import net.fabricmc.loom.configuration.providers.mappings.MappingNamespace;
+import net.fabricmc.loom.configuration.providers.mappings.intermediary.IntermediaryMappingLayer;
+import net.fabricmc.mappingio.MappingVisitor;
+import net.fabricmc.mappingio.adapter.MappingSourceNsSwitch;
+import net.fabricmc.mappingio.format.Tiny2Reader;
+
+public record TinyMappingLayer(File mappingFile) implements MappingLayer {
+	private static final String MAPPING_PATH = "mappings/mappings.tiny";
```
modmuss50: Ideally we would want to be able to change this from the spec, idk if its a big priority

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingLayer.java
```diff
@@ -0,0 +1,84 @@
...
+					read(mappingVisitor, reader);
+				}
+			}
+		}
+	}
+
+	private void read(MappingVisitor visitor, Reader reader) throws IOException {
+		// we might have official -> intermediary, named here - let's reorder just to be safe
+		MappingSourceNsSwitch nsSwitch = new MappingSourceNsSwitch(visitor, getSourceNamespace().stringValue());
+		Tiny2Reader.read(reader, nsSwitch);
```
modmuss50: How hard would it to become a general MappingLayer that supports all formats supported by mapping-io?

modmuss50 (commented):
This is a great start, I did look at adding a general mapping layer here: https://github.com/FabricMC/fabric-loom/compare/dev/0.9...modmuss50:mappingiolayer but got a bit burnt out by trying to do too much.

Having something simple like this that covers most cases might be a better first step than working towards covering all possible use cases.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingLayer.java
Player3324: There is actually `MappingReader.read` which will auto-detect the format if null. Still needs a way for the user to funnel the name spaces through though, as they may be MappingUtil.NS_SOURCE_FALLBACK/NS_TARGET_FALLBACK or some non-fabric conventions.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingsSpec.java
Juuxel: I wanted to use `Dependency`, but resolving that would tie the mapping context to a Gradle `Project` (so no mocking?)

Also, would it be possible to use a truncated sha256 hash of the resulting mapping file? That feels more reliable than java hash codes.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/TinyMappingLayer.java
Juuxel: Added in a builder class.

PR #490, opened 2021-09-10
Expose layered mappings as an API

modmuss50:
Very experimental, but cleans stuff up a bit, and makes it possible to add your own MappingsSpec without stepping too far over the api/impl line. (Still not perfect but a lot better).

There is room for better docs but thats for another time.


modmuss50:
Going to merge this, as usual there is more to do, but this is another good step forward. Still experimental so prone to changes.

PR #483, opened 2021-08-31
Make parchment layer test also test the parameter after being reordered

shedaniel:
The reorder method is taken straight from LayeredMappingsDependency: https://github.com/FabricMC/fabric-loom/blob/3ded0964c4282060ca9f50ef9db908d83b30e443/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsDependency.java#L78-L80

This PR should fail tests.

PR #491, opened 2021-09-11
Fix ParchmentMappingLayerTest test

shedaniel:
Read more about this issue in #483 

This should fix tests.

Parameter names were dropped due to their missing source names, this is considered a bug in mapping-io and was fixed.

This PR updates mapping-io, and reverted a mess up by me in #483 with the removing prefix option.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,7 +83,7 @@ dependencies {
 
 	implementation 'net.fabricmc:access-widener:1.1.0'
 	// wait for player to publish a version
-	implementation 'net.fabricmc:mapping-io:0.2.9999'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -82,7 +82,8 @@ dependencies {
 	implementation ('net.fabricmc:tiny-mappings-parser:0.3.0+build.17')
 
 	implementation 'net.fabricmc:access-widener:1.1.0'
-	implementation 'net.fabricmc:mapping-io:0.2.0'
+	// wait for player to publish a version
```
modmuss50: Remove please :)

shedaniel: Removed

Juuxel (approved):

PR #495, opened 2021-09-14
Migrate fully to mapping-io + Add comments about transitive access widners to generated sources

modmuss50:
Needs tests and a better java doc comment:

Screenshot 2021-09-14 at 23 57 39

Also removes TMP and moves everything to mapping-io


Player3324 (approved):

src/main/java/net/fabricmc/loom/task/MigrateMappingsTask.java
```diff
@@ -149,20 +148,21 @@ private File loadMappings() {
...
 			Files.copy(fileSystem.getPath("mappings/mappings.tiny"), temp, StandardCopyOption.REPLACE_EXISTING);
 		}
 
-		try (BufferedReader reader = Files.newBufferedReader(temp)) {
-			return TinyMappingFactory.loadWithDetection(reader);
-		}
+		MemoryMappingTree mappingTree = new MemoryMappingTree();
+		MappingReader.read(temp, mappingTree);
+
+		return mappingTree;
```
Juuxel: (A bit off-topic for this PR) Couldn't this read the mappings from the jar directly since that uses `Path`s too?

modmuss50: Done

PR #496, opened 2021-09-16
Interface Injection

shartte:


Juuxel (approved):

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -0,0 +1,273 @@
...
+		@Override
+		public void visit(int version, int access, String name, String signature, String superName, String[] interfaces) {
+			Set<String> modifiedInterfaces = new LinkedHashSet<>(interfaces.length + injectedInterfaces.size());
+			Collections.addAll(modifiedInterfaces, interfaces);
+
+			for (InjectedInterface injectedInterface : injectedInterfaces) {
+				modifiedInterfaces.add(injectedInterface.ifaceName());
+			}
+
+			super.visit(version, access, name, signature, superName, modifiedInterfaces.toArray(new String[0]));
```
Ueaj-Kerman, line 267: You are not fixing the signature here, if the signature is null, u can ignore it more or less, but if it's not then u will need to fix the signature, this can be done just by appending the injected interface's descriptor to the end of the signature. If this is not done, the interface may not appear in source if the class has any kind of generics (type arguments or parameterized supers)

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -0,0 +1,284 @@
...
+		} catch (IOException e) {
+			return true;
+		}
+
+		if (hash == null) {
+			return true;
+		}
+
+		return !Arrays.equals(inputHash, hash);
+	}
```
modmuss50, line 165: You shouldnt need to do this, see JarProcessor#getId and TransitiveAccessWidenerJarProcessor.

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
shartte: Alright, I tried my best at amending the signature. Does this seem sane to you?

modmuss50 (approved):

PR #502, opened 2021-09-22
Support a wider range of parameters in FileSpec

Juuxel:
Now supports
- all `Provider`s recursively (incl. `Property`)
- groovy template string literal `GString` and other `CharSequence`s (this crashed before if you had sth like `parchment("group:name:$parchmentVersion")`)
- `Path` and gradle `FileSystemLocation` (the values of `RegularFileProperty`)

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileSpec.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileSpec.java
@@ -73,9 +73,9 @@ public interface FileSpec {
 		} else if (o instanceof File f) {
 			return createFromFile(f);
 		} else if (o instanceof Path p) {
-			return createFromFile(p.toFile());
+			return createFromFile(p);
 		} else if (o instanceof FileSystemLocation l) {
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):

src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileSpec.java
```diff
@@ -42,17 +44,35 @@
  */
 @ApiStatus.Experimental
 public interface FileSpec {
+	/**
+	 * Creates a file spec.
+	 *
+	 * <p>The parameter will be evaluated like this:
+	 * <ul>
+	 * <li>{@link File}, {@link Path} and {@link org.gradle.api.file.FileSystemLocation} will be resolved as local files</li>
```
Juuxel:
```suggestion
	 * <li>{@link File}, {@link Path} and {@link FileSystemLocation} will be resolved as local files</li>
```
It's already imported, so FQN not necessary

modmuss50 (approved):

PR #504, opened 2021-09-22
Rewrire Jar and Source remapping tasks.

modmuss50:
Brings it upto date with modern gradle features, should fix a lot of oddities along the way as well as making it much more maintainable.

See:

https://docs.gradle.org/current/userguide/build_services.html#operation_listener
https://docs.gradle.org/current/userguide/lazy_configuration.html


src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+
+public class RemapTaskConfiguration {
+	private static final String REMAP_JAR_TASK_NAME = "remapJar";
+
+	public static void setupRemap(Project project) {
+		TaskContainer tasks = project.getTasks();
+
+		// Configure the default jar task
+		tasks.named(JavaPlugin.JAR_TASK_NAME, AbstractArchiveTask.class).configure(task -> {
+			task.getArchiveClassifier().set("dev");
```
ghost: `task.getArchiveClassifier().convention("dev");` if it works.

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+
+			task.getTinyRemapperBuildService().convention(getOrCreatetinyRemapperBuildServiceProvider(project,
+					task.getSourceNamespace(),
+					task.getTargetNamespace(),
+					classpath
+			));
+		});
+	}
+
+	private static Provider<TinyRemapperBuildService> getOrCreatetinyRemapperBuildServiceProvider(final Project project, final Provider<String> from, final Provider<String> to, final FileCollection classpath) {
```
Juuxel: Minor nitpick: `Tiny` should be capitalised

Juuxel (changes requested):
Generally really nice (I love the parallel support, proper properties and dropping afterEvaluate if possible), but does it still not support the `Jar` api? To me, it looks like it just still has the api and blissfully ignores it.

This can probably be solved by letting the regular logic generate the jar and remapping that in place (input file = `from(zipTree(otherFile))`)

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
modmuss50: Fixed.

modmuss50:
This should now be feature complete, it seems to be starting to work as expected. Got a few more tests I want to do before merging. I expect there will be issues especially around some edge cases that we dont test for, but those can always be fixed later.

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+				project.getLogger().info("{} task was not found, not remapping sources", sourcesJarTaskName);
+				return;
+			}
+
+			if (!(sourcesTask instanceof Jar sourcesJarTask)) {
+				project.getLogger().info("{} task is not a Jar task, not remapping sources", sourcesJarTaskName);
+				return;
+			}
+
+			sourcesJarTask.getArchiveClassifier().set("dev-sources");
```
Juuxel:
```suggestion
			sourcesJarTask.getArchiveClassifier().convention("dev-sources");
```

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+			// Setup the input file and the nested deps
+			task.getInputFile().convention(jarTask.getArchiveFile());
+			Configuration includeConfiguration = project.getConfigurations().getByName(Constants.Configurations.INCLUDE);
+			task.getNestedJars().from(new IncludedJarFactory(project).getNestedJars(includeConfiguration));
+		});
+
+		// Configure the default jar task
+		tasks.named(JavaPlugin.JAR_TASK_NAME, AbstractArchiveTask.class).configure(task -> {
+			task.getArchiveClassifier().convention("dev");
+			task.finalizedBy(remapJarTaskProvider);
```
Juuxel: Hmm, so it's not possible to run `jar` without `remapJar` anymore?

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+
+import net.fabricmc.loom.LoomGradleExtension;
+import net.fabricmc.loom.build.nesting.IncludedJarFactory;
+import net.fabricmc.loom.util.Constants;
+
+public class RemapTaskConfiguration {
+	public static final String REMAP_JAR_TASK_NAME = "remapJar";
+	public static final String REMAP_SOURCES_JAR_TASK_NAME = "remapSourcesJar";
+
+	// TODO respect extension.getSetupRemappedVariants in here
```
Juuxel: This should be fairly simple, just `if` gates around the parts in this class that deal with `api/runtime/sourcesElements`

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -27,276 +27,255 @@
...
+		submitWork(RemapAction.class, params -> {
+			if (getAddNestedDependencies().get()) {
+				params.getNestedJars().plus(getNestedJars());
+			}
 
-		if (singleRemap) {
-			jarRemapper.remap();
-		}
-	}
+			params.getRemapAccessWidener().set(extension.getAccessWidenerPath().isPresent());
```
Juuxel: IMO this should rather read it from the fabric.mod.json than check the extension. This approach has some issues with multiproject jars where
- project A has an AW
- project B has a shadowed copy of project A's `namedElements`, so the AW would still be `named`
- project B declares A's AW in its fabric.mod.json but it's not remapped because project B doesn't declare it

(one example of these project setups is Adorn, where I have to unnecessarily apply the AW to "project B" and create a second MC jar when only "project A" needs it)

src/main/java/net/fabricmc/loom/task/service/TinyRemapperService.java
```diff
@@ -0,0 +1,128 @@
...
+
+		Property<Boolean> getUseMixinExtension();
+	}
+
+	public static synchronized Provider<TinyRemapperService> create(Project project, String from, String to) {
+		final LoomGradleExtension extension = LoomGradleExtension.get(project);
+
+		return project.getGradle().getSharedServices().registerIfAbsent(extension.getMappingsProvider().getBuildServiceName("remapJarService", from, to), TinyRemapperService.class, spec -> {
+			spec.parameters(params -> {
+				params.getClasspath().plus(project.getConfigurations().getByName(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME));
```
Juuxel: Shouldn't the classpath rather come from `remapJar`'s configuration and not be hardcoded? Also, this makes it hard to declare some custom `RemapJarTask`s since the classpath is locked to always have these files.

Juuxel (changes requested):

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
modmuss50: I have made it always remap the AW if one is present in the jar, I dont see a usecase to remap the classes and not the AW.

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -69,32 +67,37 @@ public static void setupRemap(Project project) {
...
 		});
 
 		// Configure the default jar task
 		tasks.named(JavaPlugin.JAR_TASK_NAME, AbstractArchiveTask.class).configure(task -> {
 			task.getArchiveClassifier().convention("dev");
-			task.finalizedBy(remapJarTaskProvider);
 			task.getDestinationDirectory().set(new File(project.getBuildDir(), "devlibs"));
 		});
 
+		tasks.named("assemble").configure(task -> task.dependsOn(remapJarTaskProvider));
```
Juuxel: FYI there's a constant in `BasePlugin` for this

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -0,0 +1,136 @@
...
+			task.dependsOn(sourcesJarTask);
+			task.getInputFile().convention(sourcesJarTask.getArchiveFile());
+
+			project.getArtifacts().add(JavaPlugin.SOURCES_ELEMENTS_CONFIGURATION_NAME, task);
+
+			if (extension.getSetupRemappedVariants().get()) {
+				// Remove the dev sources artifact
+				Configuration configuration = project.getConfigurations().getByName(JavaPlugin.SOURCES_ELEMENTS_CONFIGURATION_NAME);
+				configuration.getArtifacts().removeIf(a -> a.getFile().equals(sourcesJarTask.getArchiveFile().get().getAsFile()));
+			}
```
Juuxel:
```suggestion
			if (extension.getSetupRemappedVariants().get()) {
				project.getArtifacts().add(JavaPlugin.SOURCES_ELEMENTS_CONFIGURATION_NAME, task);

				// Remove the dev sources artifact
				Configuration configuration = project.getConfigurations().getByName(JavaPlugin.SOURCES_ELEMENTS_CONFIGURATION_NAME);
				configuration.getArtifacts().removeIf(a -> a.getFile().equals(sourcesJarTask.getArchiveFile().get().getAsFile()));
			}
```

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
modmuss50: Ah thanks, I didnt know where it was hiding :D

Juuxel: Also it might be nice to hook `remapSourcesJar` to this as well

modmuss50:
Going to merge this, all of the tests are now passing (ignoring the random failiures). There may be some more to be done to this later, but this should be a great improvement. 👍

PR #508, opened 2021-09-28
Add -Dloom.refresh=true option

shedaniel:


src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
```diff
@@ -64,7 +64,7 @@ public void apply(PluginAware target) {
 	public void apply(Project project) {
 		project.getLogger().lifecycle("Fabric Loom: " + LOOM_VERSION);
 
-		refreshDeps = project.getGradle().getStartParameter().isRefreshDependencies();
+		refreshDeps = project.getGradle().getStartParameter().isRefreshDependencies() || Objects.equals("true", System.getProperty("loom.refresh"));
```
modmuss50: Minor nit pick:

```suggestion
		refreshDeps = project.getGradle().getStartParameter().isRefreshDependencies() || Boolean.getBoolean("loom.refresh");
```

Looks good otherwise 👍

shedaniel: thanks, didn't know that existed :O

modmuss50 (approved):

PR #511, opened 2021-09-30
Rewrite CFR & FernFlower decompiler interface.

### LINKED ISSUE #355: Could not find or load main class net.fabricmc.loom.decompilers.fernflower.FabricForkedFFExecutor
This is a reopening of #288

https://github.com/warjort/fabric-example-mod

The above fork modifies the example mod to use subprojects and the latest loom from maven. 
There is no loom in the root project.
Both "gensources" and "example:gensources" fail to launch the jvm with that classpath error.

modmuss50:
CFR Linemapping and javadocs are working as expected.
Fernflower has been cleaned up.

The genSources task now uses gradle's WorkerExecutor to fork, should work a lot better when loom is not applied to the root project.

https://openjdk.java.net/jeps/380 has been used to create an IPC channel between the forked decompiler JVM and the main JVM to handle logging.


modmuss50:
Now works on windows 7 for the 2 of you that still use it ;)

modmuss50:
Going to merge this, please let me know if (when) anything breaks. Trying out CFR would be great.

PR #513, opened 2021-10-03
Use NIO instead of ZipUtil

shedaniel:
ZipUtil is slow, doing `transformEntries` actually copies unrelated ZipEntry, which is why it is a better idea to replace it in favor of NIO.

Example: JarProcessorManager.process#manifestTransformed comes down from 1.5s to 100ms

The only issue with NIO is the single-state FS, which may be problematic in the future, but currently this is fine (and we use NIO sometimes in loom anyways)


src/main/java/net/fabricmc/loom/util/NIOZipUtils.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.stream.Stream;
+
+import com.google.common.base.MoreObjects;
+import org.jetbrains.annotations.Nullable;
+
+import net.fabricmc.loom.LoomGradlePlugin;
+import net.fabricmc.stitch.util.Pair;
+import net.fabricmc.stitch.util.StitchUtil;
+
+public class NIOZipUtils {
```
modmuss50: I will do a full review later, but this looks really good 👍 

I would rename this class to just `ZipUtils` now.

liach (approved):
Removing another library is always a good thing

shedaniel:
one test failed and only the windows side 👀👀

Juuxel:
The windows error might even be unrelated to the changes:
> ```
> [19:57:18] [Server thread/INFO]: Saving chunks for level 'ServerLevel[world]'/minecraft:overworld
> #
> # There is insufficient memory for the Java Runtime Environment to continue.
> # Native memory allocation (malloc) failed to allocate 1799536 bytes for Chunk::new
> # An error report file with more information is saved as:
> # C:\Users\RUNNER~1\AppData\Local\Temp\groovy-generated-tmpdir-13963685673897551665\server\hs_err_pid3768.log
> #
> # Compiler replay data is saved as:
> # C:\Users\RUNNER~1\AppData\Local\Temp\groovy-generated-tmpdir-13963685673897551665\server\replay_pid3768.log
> )
> 
> 	at net.fabricmc.loom.test.integration.FabricAPITest.build and run (gradle #version)(FabricAPITest.groovy:65)
> ```

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,254 @@
...
+
+			return Files.exists(fsPath);
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to check file from zip", e);
+		}
+	}
+
+	public static void unpackAll(Path zip, Path output) {
+		try (StitchUtil.FileSystemDelegate fs = StitchUtil.getJarFileSystem(new File(zip.toFile().getAbsolutePath()), false)) {
+			for (Path fsPath : (Iterable<Path>) Files.walk(fs.get().getPath("/"))::iterator) {
```
Juuxel: `Files.walk` should be closed in a try block. According to [the JDK javadoc about `Files.walk`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#walk(java.nio.file.Path,int,java.nio.file.FileVisitOption...)):

> This method must be used within a try-with-resources statement or similar control structure to ensure that the stream's open directories are closed promptly after the stream's operations have completed.

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,254 @@
...
+		try {
+			Files.deleteIfExists(zip);
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to add file to zip", e);
+		}
+
+		if (!Files.isDirectory(from)) throw new IllegalArgumentException(from + " is not a directory!");
+
+		try (StitchUtil.FileSystemDelegate fs = StitchUtil.getJarFileSystem(zip.toFile(), true)) {
+			for (Path fromPath : (Iterable<Path>) Files.walk(from)::iterator) {
```
Juuxel: Same here.

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,254 @@
...
+
+	public static boolean transformString(Path zip, Collection<Pair<String, UnsafeUnaryOperator<String>>> transforms) {
+		return transformString(zip, transforms.stream());
+	}
+
+	public static boolean transformString(Path zip, Stream<Pair<String, UnsafeUnaryOperator<String>>> transforms) {
+		return transformString(zip, transforms
+				.collect(Collectors.groupingBy(Pair::getLeft,
+						Collectors.mapping(Pair::getRight,
+								Collectors.reducing(null, MoreObjects::firstNonNull)))));
```
Juuxel: Hmm, I wonder if this is a bit inefficient with the nested collectors...

Juuxel (changes requested):

src/main/java/net/fabricmc/loom/util/ZipUtils.java
shedaniel: Shouldn't be too bad, if it is I can replace it with non-stream code

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -251,4 +249,16 @@ public static boolean transform(Path zip, Map<String, UnsafeUnaryOperator<byte[]
...
 		R apply(T arg) throws IOException;
 	}
+
+	private static <T> Map<String, UnsafeUnaryOperator<T>> collectTransformersStream(Stream<Pair<String, UnsafeUnaryOperator<T>>> transforms) {
+		Map<String, UnsafeUnaryOperator<T>> map = new HashMap<>();
+		Iterator<Pair<String, UnsafeUnaryOperator<T>>> iterator = transforms.iterator();
+
+		while (iterator.hasNext()) {
+			Pair<String, UnsafeUnaryOperator<T>> next = iterator.next();
+			map.put(next.getLeft(), next.getRight());
```
liach: This has an inconsistent duplicate handling mechanism from the previous version, where this uses the latest of duplicates while the previous uses the earliest of duplicates. Should we just add a warning if there is any duplicate?

shedaniel: hmm I think it should be better to chain the transformers here

P.S. it doesn't have to be consistent with the previous code, these are all written by me

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,264 @@
...
+import java.util.Iterator;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import org.jetbrains.annotations.Nullable;
+
+import net.fabricmc.loom.LoomGradlePlugin;
+import net.fabricmc.stitch.util.Pair;
+import net.fabricmc.stitch.util.StitchUtil;
```
modmuss50: Dont use stitch for this, long term goal is to move away from it, so we dont need any more code based on it.

modmuss50 (changes requested):

src/main/java/net/fabricmc/loom/util/ZipUtils.java
shedaniel: Do you want me to essentially copy that class as an util into loom?

modmuss50: Yes, thats fine. 👍

shedaniel: Done

src/main/java/net/fabricmc/loom/util/FileSystemUtil.java
```diff
@@ -0,0 +1,87 @@
...
+		public FileSystem get() {
+			return fs;
+		}
+	}
+
+	private FileSystemUtil() {
+	}
+
+	private static final Map<String, String> jfsArgsCreate = new HashMap<>();
+	private static final Map<String, String> jfsArgsEmpty = new HashMap<>();
```
liach: Since we are on new java, just use `Map.of("create", "true")` and `Map.of()`, which is also safer for immutability.

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,262 @@
...
+import net.fabricmc.stitch.util.Pair;
+
+public class ZipUtils {
+	public static boolean contains(Path zip, String path) {
+		try (FileSystemUtil.Delegate fs = FileSystemUtil.getJarFileSystem(zip, false)) {
+			Path fsPath = fs.get().getPath(path);
+
+			return Files.exists(fsPath);
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to check file from zip", e);
```
modmuss50, line 56: I think these should all throw the IOException, using UncheckedIOException is just going to lead to lazy error handling.

I think doing this might be out of scope of this PR however, and can be done at a later date. 👍

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,262 @@
...
+				Path dstPathParent = dstPath.getParent();
+				if (dstPathParent != null) Files.createDirectories(dstPathParent);
+				Files.copy(fsPath, dstPath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.COPY_ATTRIBUTES);
+			}
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to unpack file from zip", e);
+		}
+	}
+
+	public static byte @Nullable [] unpack(Path zip, String path) {
```
modmuss50:
```suggestion
	public static @Nullable byte[] unpack(Path zip, String path) {
```

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,262 @@
...
+		try {
+			return unpackStrict(zip, path);
+		} catch (UncheckedIOException e) {
+			if (e.getCause() instanceof NoSuchFileException) {
+				return null;
+			}
```
modmuss50: Throw FileNotFound? Again possibly something for later, but I think worth adding a TODO for the very least.

Just seen unpackStrict 👍

src/main/java/net/fabricmc/loom/util/ZipUtils.java
```diff
@@ -0,0 +1,262 @@
...
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to unpack file from zip", e);
+		}
+	}
+
+	public static void pack(Path from, Path zip) {
+		try {
+			Files.deleteIfExists(zip);
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to add file to zip", e);
```
modmuss50:
```suggestion
			throw new UncheckedIOException("Failed delete existing zip file", e);
```

modmuss50 (commented):
Overall looks good, I think this is a good case of not getting something perfect the first time. Its quite clealy a NIO port of the old zip utils lib, this is a great steip forward.

Beyond this there are some ovious improvements I think will be worth making down the line, I am happy to pull this in in its current state.

Thanks 👍

src/main/java/net/fabricmc/loom/util/ZipUtils.java
shedaniel: @Nullable can't apply to booleans, the @Nullable in front of the [] means nullable array, @Nullable in front of booleans means the element type is nullable

shedaniel:
I think I will just remove UncheckedIOExceptions to make resolve most of the reviews issues

src/main/java/net/fabricmc/loom/util/ZipUtils.java
liach: yeah this is how type annotations work. It is ugly indeed!

modmuss50: Ah, nice always learning something, thanks.

modmuss50:
Pushed quite a few changes, before (pre this PR) a few places would have failed silently, this is a bit more offensive.

modmuss50 (approved):
Ive made requested changes, and added some unit tests (And fixed a bug by doing so). Im happy to merge this unless anyone has an objection.

Thanks a lot for doing the bulk of the work. 👍

PR #518, opened 2021-10-15
Add AccessWidenerValidator

modmuss50:
This ensures that all lines in a AW are valid, and relate to a class/method/field in the mapped mc jar.

I havent added a way to skip it as I am unsure on a usecase for this, if anyone has a good usecase I can easily add something to allow skipping the validation.


src/main/java/net/fabricmc/loom/configuration/LoomDependencyManager.java
```diff
@@ -174,6 +175,15 @@ public void handleDependencies(Project project) {
...
 		}
+
+		// TODO a way to disable this?
+		if (extension.getAccessWidenerPath().isPresent()) {
+			AccessWidenerValidator.validate(
+					extension.getAccessWidenerPath().get().getAsFile().toPath(),
+					extension.getMinecraftMappedProvider().getMappedJar().toPath(),
+					project.getLogger()
+			);
+		}
```
Juuxel: IMO this should be a task that `check` automatically depends on instead of being hardcoded during project setup (although the project setup does use the AW file too).

Juuxel (changes requested):

src/main/java/net/fabricmc/loom/configuration/LoomDependencyManager.java
modmuss50: Thats a good idea, I didnt think of that. Would make it easy to disable it as well if wanted.

I will do that 👍

modmuss50: Moved to task 👍

src/test/groovy/net/fabricmc/loom/test/integration/AccessWidenerTest.groovy
```diff
@@ -65,4 +65,24 @@ class AccessWidenerTest extends Specification implements GradleProjectTestTrait
...
+
+	@Unroll
+	def "invalid (#awLine)"() {
+		setup:
+			def gradle = gradleProject(project: "accesswidener", version: version)
+			new File(gradle.projectDir, "src/main/resources/modid.accesswidener").append(awLine)
+			def errorPrefix = "Failed to validate access-widener file modid.accesswidener on line 10: java.lang.RuntimeException: "
+
+		when:
+			def result = gradle.run(task: "check", expectFailuire: true)
```
Juuxel:
```suggestion
			def result = gradle.run(task: "check", expectFailure: true)
```

src/test/groovy/net/fabricmc/loom/test/util/GradleProjectTestTrait.groovy
```diff
@@ -156,7 +156,7 @@ trait GradleProjectTestTrait {
 
             runner.withArguments(args as String[])
 
-            return runner.build()
+            return options.expectFailuire ? runner.buildAndFail() : runner.build()
```
Juuxel:
```suggestion
            return options.expectFailure ? runner.buildAndFail() : runner.build()
```

src/main/java/net/fabricmc/loom/task/LoomTasks.java
```diff
@@ -57,6 +57,13 @@ public static void registerTasks(Project project) {
 		tasks.register("downloadAssets", DownloadAssetsTask.class, t -> t.setDescription("Downloads required assets for Fabric."));
 		tasks.register("remapSourcesJar", RemapSourcesJarTask.class, t -> t.setDescription("Remaps the project sources jar to intermediary names."));
 
+		tasks.getByName("check").dependsOn(
+				tasks.register("validateAccessWidener", ValidateAccessWidenerTask.class, t -> {
+					t.setDescription("Validate that all the rules in the access widener against the minecraft jar");
```
Juuxel:
```suggestion
					t.setDescription("Validate all the rules in the access widener against the Minecraft jar");
```

Juuxel (changes requested):

PR #520, opened 2021-10-15
Only add loader + mc deps to AP classpath

modmuss50:
Should fix locked file issues.


modmuss50:
Fixed in Mixin via: https://github.com/FabricMC/Mixin/commit/87f48a945c107539ff2ace92bef65a220b616812

PR #523, opened 2021-10-22
fix launch on old versions

thecatcore:


modmuss50:
What minecraft versions is this for? Please also target the dev/0.10 branch. Backports can be sorted later.

thecatcore:
It's for 1.7.10 and older. The target is 0.7 because it's the last version of loom to support both custom intermediary url and Java 8 which is the recommended version for old versions.

halotroop2288 (approved):

shedaniel:
You should be adding the arguments via the runs block in the loom extensions, however I am not sure if 0.7 have those, you'd better off with forking Loom though. Also you can compile to Java 8 with latest Loom, doesn't matter

modmuss50:
I agree with @shedaniel, 0.7 can do that, and you can use 0.10 to build and run for java 8.

Ideally we should be using the args from the launcher meta for each specific version, but for now I would just add it to your build.gradle.

PR #525, opened 2021-10-25
Add dev config with default jar task on it, to aid with depending on …

modmuss50:
…sub projects.

Dpeneding on a sub project is now just `project(path: ":name", configuration: "namedElements")`.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -57,7 +57,10 @@ public final class CompileConfiguration {
 		extension.createLazyConfiguration(Constants.Configurations.MINECRAFT).configure(configuration -> configuration.setTransitive(false));
 		extension.createLazyConfiguration(Constants.Configurations.INCLUDE).configure(configuration -> configuration.setTransitive(false)); // Dont get transitive deps
 		extension.createLazyConfiguration(Constants.Configurations.MAPPING_CONSTANTS);
-		extension.createLazyConfiguration(Constants.Configurations.DEV);
+		extension.createLazyConfiguration(Constants.Configurations.NAMED_ELEMENTS).configure(configuration -> {
+			configuration.setCanBeConsumed(true);
+			configuration.setCanBeResolved(false);
+		});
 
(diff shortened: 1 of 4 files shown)


shedaniel (approved):

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -57,6 +57,7 @@ public static void setupConfigurations(Project project) {
 		extension.createLazyConfiguration(Constants.Configurations.MINECRAFT).configure(configuration -> configuration.setTransitive(false));
 		extension.createLazyConfiguration(Constants.Configurations.INCLUDE).configure(configuration -> configuration.setTransitive(false)); // Dont get transitive deps
 		extension.createLazyConfiguration(Constants.Configurations.MAPPING_CONSTANTS);
+		extension.createLazyConfiguration(Constants.Configurations.DEV);
```
Juuxel: Should be declared `canBeResolved = false; canBeConsumed = true`

Juuxel (changes requested):

Juuxel:
`dev` is a bit generic of a name, so I wonder if it'll conflict a lot with existing configurations declared by users? You could always go with `devElements` to mirror `apiElements`/`runtimeElements`. But a good change nonetheless :rocket:

modmuss50:
I was unsure on it conflicting, `named`, `namedOutputs` or `namedElements` might make more sense?

Juuxel:
`namedOutputs` and `namedElements` would both be good imo. "outputs" is clear to the user, but "elements" does follow the gradle conventions, which might be good as well.

Juuxel (approved):

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -147,6 +151,9 @@ public static void configureCompile(Project p) {
 			}
 		});
 
+		// Add the "dev" jar to the "dev" configuration
```
Juuxel:
```suggestion
		// Add the "dev" jar to the "namedElements" configuration
```

PR #527, opened 2021-10-26
Make CFR the default decompiler

modmuss50:
- Make CFR the default decompiler
- Expose decompiler options


src/main/java/net/fabricmc/loom/task/GenerateSourcesTask.java
```diff
@@ -98,6 +103,7 @@ public GenerateSourcesTask(LoomDecompiler decompiler) {
 
 		getOutputs().upToDateWhen((o) -> false);
 		getMaxMemory().convention(4096L).finalizeValueOnRead();
+		getOptions().convention(Collections.emptyMap()).finalizeValueOnRead();
```
Juuxel: Iirc `empty()` suffices as the convention here

modmuss50: Yep, I was unsure if this was the case, just added this to be sure. 👍

PR #528, opened 2021-10-31
Allow to set custom intermediary url again

thecatcore:


src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionApiImpl.java
```diff
@@ -85,6 +86,7 @@ protected LoomGradleExtensionApiImpl(Project project, LoomFiles directories) {
 		this.transitiveAccessWideners = project.getObjects().property(Boolean.class)
 				.convention(true);
 		this.transitiveAccessWideners.finalizeValueOnRead();
+		this.intermediary = project.getObjects().property(String.class);
```
ghost: Please use conventions to set the default value rather than getOrDefault when getting the URL.

Will do a better check when home as it's a bit hard on mobile

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -211,4 +211,10 @@ default String getCustomManifest() {
 	 * @return the property controlling the transitive access wideners
 	 */
 	Property<Boolean> getEnableTransitiveAccessWideners();
+
+	Property<String> getIntermediaryUrl();
+
+	default void setIntermediaryUrl(String url) {
+		getIntermediaryUrl().set(url);
+	}
```
modmuss50: This provides no benefit. You can just do `intermediaryUr = "https://example.com"` on the popertry.

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionApiImpl.java
```diff
@@ -173,6 +175,10 @@ public String getModVersion() {
 
 	protected abstract LoomFiles getFiles();
 
+	public Property<String> getIntermediaryUrl() {
```
modmuss50, line 180: Missing `@Override`

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -99,8 +99,9 @@ default boolean ideSync() {
 	}
 
 	default String getIntermediaryUrl(String minecraftVersion) {
-		// TODO reimplement a way to change this, was never really supported api anyway
-		return String.format("https://maven.fabricmc.net/net/fabricmc/intermediary/%1$s/intermediary-%1$s-v2.jar", minecraftVersion);
+		return String.format(this.getIntermediaryUrl()
+						.getOrElse("https://maven.fabricmc.net/net/fabricmc/intermediary/%1$s/intermediary-%1$s-v2.jar"),
+				minecraftVersion);
```
modmuss50: Should the poperty set this string as the convenion?

modmuss50 (changes requested):

Juuxel:
I feel like the format should be documented as well. It's not obvious from the property (which itself needs a short explaining JD comment IMO) that there are specific format parameters for it.

modmuss50 (approved):

ghost (approved):

PR #531, opened 2021-11-07
Add a test for leaking runtime scope dependencies on compile classpath

Juuxel:
This makes MavenProjectTest fail because dependencies declared with a "runtime" scope are still put on the compile classpath when you use a `mod*` config on the compile classpath. I don't really know at the moment how to fix this, but this test is a good starting point IMO.


modmuss50:
Test looks good, I think the fix for this should wait for 0.11

Juuxel:
I believe this was fixed by #807.

PR #533, opened 2021-11-11
Fix `namedElements` not extending from api

modmuss50:
Fix artifact not waiting for remapAllJars with shared caches

+ Intergration tests for both


Juuxel:
`apiElements` might be better here since it's all the things intended to publish at compile time. On the other hand, there's also the `runtimeElements` for runtime stuff. (I think you can also do something to separate compile time and runtime elements for outgoing configurations, but haven't looked into that.)

Now that I've written this I realise that those might include intermediary mods, so that's no good...

Juuxel (approved):

modmuss50:
@Juuxel mind taking another quick look at `ModCompileRemapper.java` I had to do this to get named `modApi` mods onto the namedElements config. It seems to work in my test and dont think it will have any odd side affects.

Juuxel:
Seems all correct (and pretty clever too). I might take a look at some point whether it's possible to do the compile time/runtime split for dependencies for this config, but that's not really important. (Although forwarding required runtime libs would be nice I guess, but not very urgent)

PR #536, opened 2021-11-18
Add support for mixin files outside of root in MixinRefmapHelper

ghost:
This PR adds the ability to add the `refmap` attribute in the mixin json files that are not located at the root of the jar file/in the resources folder. This is not a major issue as modders can add it manually (if they still want to put the files in a subdirectory), but can be convenient to have, since these mixin files would still load fine in the game.


thecatcore (approved):

oliviathevampire (approved):

modmuss50:
I think it would be worth adding/expanding the intergration tests to cover this. Partly to make sure it works and most importantally to ensure that it doesnt get broken at a later date.

I think it might fit nicely into this exisitng test: 

https://github.com/FabricMC/fabric-loom/blob/dev/0.10/src/test/groovy/net/fabricmc/loom/test/integration/MixinApAutoRefmapTest.groovy

https://github.com/FabricMC/fabric-loom/tree/dev/0.10/src/test/resources/projects/mixinApAutoRefmap

Let me know if you need any help with this. 👍

modmuss50:
Also looks like you have a number of checkstyle issues.

src/main/java/net/fabricmc/loom/build/MixinRefmapHelper.java
```diff
@@ -94,6 +104,15 @@ public static boolean addRefmapName(Project project, Path outputPath) {
 		}
 	}
 
+	private static String getRootPath(SourceSet sourceSet) {
+		File root = Iterables.getFirst(sourceSet.getResources().getSrcDirs(), null);
```
Juuxel: Hmm, it could also be nice to search for the mixin file in all resource directories of the source set (this `Iterables.getFirst` can cause unexpected errors when splitting source sets into many directories)

ghost:
> I think it would be worth adding/expanding the intergration tests to cover this. [...] Let me know if you need any help with this. +1

I think I understand how to do the integration test, but I'm not entirely sure.

> Also looks like you have a number of checkstyle issues.

I just noticed that, and I'm working on fixing those, but I don't really understand what the `missing blank line before block at same indentation level` mean.

Juuxel:
> I just noticed that, and I'm working on fixing those, but I don't really understand what the missing blank line before block at same indentation level mean.

You need to have empty lines before and after `if` (and other) blocks.

```java
abc();

if (x) {
    ...
}

ghi();
```

ghost:
> it could also be nice to search for the mixin file in all resource directories of the source set

That should be easy to fix.

> You need to have empty lines before and after if (and other) blocks.

Thank you for telling me.

src/main/java/net/fabricmc/loom/build/MixinRefmapHelper.java
```diff
@@ -94,6 +119,26 @@ public static boolean addRefmapName(Project project, Path outputPath) {
 		}
 	}
 
+	private static String[] getRootPaths(SourceSet sourceSet) {
+		Set<Fire> dirs = ;
```
modmuss50: This doesnt compile.

ghost: Oops, I guess I forgot to save before pushing :/

src/main/java/net/fabricmc/loom/build/MixinRefmapHelper.java
```diff
@@ -67,11 +67,33 @@ public static boolean addRefmapName(Project project, Path outputPath) {
...
+						.map(root -> {
+							String rootPath = root.getAbsolutePath();
+
+							if (rootPath.charAt(rootPath.length() - 1) != File.separatorChar) {
+								rootPath += File.separatorChar;
+							}
+
+							return rootPath;
+						})
+						.toArray();
```
Juuxel:
```suggestion
				String[] rootPaths = sourceSet.getResources().getSrcDirs().stream()
						.map(root -> {
							String rootPath = root.getAbsolutePath();

							if (rootPath.charAt(rootPath.length() - 1) != File.separatorChar) {
								rootPath += File.separatorChar;
							}

							return rootPath;
						})
						.toArray(String[]::new);
```
I'm not sure if the cast from `Object[]` is even safe, but this is nicer anyway

ghost: I didn't know it could be done this way. At least I learned something today :)

ghost:
Sorry for the delay on this, but I just figured out that, while Windows uses backslashes for path separators ([I even thought about that at first](https://github.com/FabricMC/fabric-loom/blob/e8d0b4fa413b89dcd0a15ce0de667baffd0e1174/src/main/java/net/fabricmc/loom/build/MixinRefmapHelper.java#L74-L76)), Java (or the ZipUtils class) doesn't seem to. Hopefully, it will succeed this time.

oliviathevampire:
Could you guys please re-run the workflows?

thecatcore (approved):

PR #538, opened 2021-11-20
Add an option (enabled by default) to map synthetic field and method names from the official mojang mappings.

modmuss50:
This is enabled by default to match the existing behaviour. This default should be changed in 0.11.

This option can be changed like so:
```groovy
dependencies {
	minecraft "com.mojang:minecraft:1.16.5"
	mappings loom.layered {
		officialMojangMappings {
			nameSyntheticMembers = false
		}
	}
}
```

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/LayeredMappingSpecBuilder.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/LayeredMappingSpecBuilder.java
@@ -24,7 +24,10 @@
 
 package net.fabricmc.loom.api.mappings.layered.spec;
 
+import groovy.lang.Closure;
+import groovy.lang.DelegatesTo;
 import org.gradle.api.Action;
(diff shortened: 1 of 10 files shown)


Juuxel:
Hmm, could it be possible to have an associated getter too? Otherwise this won't have property syntax in the Kotlin DSL (kt needs at least a getter for that, a plain setter won't do).

This feels awkward:
```kotlin
officialMojangMappings {
  setNameSyntheticMembers(false)
}
```

modmuss50:
Yes, thats fine, kotlin always has to be awkward 👍

IMS212:
The setting itself works by changing it in loom’s defaults, but you cannot set it with a build script it seems.

modmuss50:
Yeah, I will add an intergration test and try and figure out why its not working.

modmuss50:
Ive added an ingtergration test for this, not quite sure why its not working.

modmuss50:
Seems to be working now 👍 Will merge.

PR #542, opened 2021-11-25
Fix `include` generating modid's that are too long for loader

NebelNidas:


modmuss50 (approved):

modmuss50:
This was rewritten in 0.11, I will include the change and commit it shortly. Thanks for the PR.

PR #547, opened 2021-11-30
Bump jackson-databind to 2.13.0

jpenilla:
Fixes issues when running alongside other gradle plugins using a newer version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,7 +63,7 @@ dependencies {
 	// libraries
 	implementation ('commons-io:commons-io:2.8.0')
 	implementation ('com.google.code.gson:gson:2.8.8')
-	implementation ('com.fasterxml.jackson.core:jackson-databind:2.12.5')
+	implementation ('com.fasterxml.jackson.core:jackson-databind:2.13.0')
 	implementation ('com.google.guava:guava:30.1.1-jre')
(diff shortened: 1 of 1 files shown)


modmuss50:
What kind of issues are you seeing? If other plugins depend on a newer version that should be pulled in. Also curious to know incase I see anyone having issues.

jpenilla:
> If other plugins depend on a newer version that should be pulled in

This is my understanding as well, but it seems like there is some weirdness that can happen with subprojects applying different plugins.

In my case, I apply loom as a settings plugin, in order to use `repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)`, which means even though plugins I apply to some subprojects depend on jackson 2.13.0, I end up with multiple versions on the classpath (the older one I found to be from loom after a build scan).

modmuss50:
This isnt really ideal, as going by that theory doing this will break other plugins that use this older version.

jpenilla:
I agree that it's not ideal, although I still think it's better to break by being more up to date than by being out of date. If this isn't something you want to do I'm fine with manually depending on jackson in my build scripts.

modmuss50:
Im going to close this, this is not a good fix to the issue you are describing. Either way I updated all in the libs fo 0.11

PR #549, opened 2021-12-02
Validate remapped mods also have a remapped access widener.

modmuss50:
Validate remapped mods also have a remapped access widener as older versions of loom generate an invalid aw file when it fails to remap. Should fix  the common case of people having runtime errors due to AW's in the wrong namespace.

Cleans up the code a little, more cleanup is needed around here at a later date.


modmuss50:
Tested seems to work. Let me know if you have any issues with it 👍

PR #554, opened 2021-12-08
Support ARM natives, rewrite native handling and misc cleanup.

modmuss50:
- Much improved and simplified natives extract task
- Use idea ext plugin to run post sync tasks (download assets, run configs, natives)
- First class support for ARM machines (Windows, linux and mac)
- Put server libs on their own classpath. (With versions that use the bundled server jar.)
- Includes a small mod patch mc to fix issues on MacOS


src/main/java/net/fabricmc/loom/configuration/ide/intelij/IntelijConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.configuration.ide.intelij;
```
Juuxel: It has two Ls (intellij), although just idea might be both easier to type and easier to read

modmuss50: Yes, I think idea might be better, my thinking behind this was to try and make sure there is no confusion with the old `idea` task, but I dont think thats an issue.

PR #555, opened 2021-12-11
Add remapJar etc. task dependencies to assemble instead of build

Juuxel:
Gradle docs:

> Plugins and build authors should attach tasks that produce distributions and other consumable artifacts to this lifecycle task.

(it's one of the tasks that `build` depends on)


modmuss50 (approved):

modmuss50:
This was done along with my remap jar rewrite in 0.11. I dont think its really needed in 0.10 so I think its ok to close this?

Thanks anyway.

PR #561, opened 2021-12-20
Split the minecraft jar into client only and common jars.

modmuss50:
This is something that has been on my radar for a while, even more since it got much easier with changes in 1.18 and 1.17, it has been confirmed that Mojang develop the game with split common / client sources.

This PR adds a `fabric.loom.experimental.splitMcJars` option that when enabled will split the normally merged into a `clientonly` and a `common` jar. The common jar is very similar to the existing server jar, however the client only jar only contains all of the classes and data that are only present in the official client jar.

Doing this breaks some funimental design decisions that have been in loom, loader and mods. This PR is a first step to help add support for doing this in fabric's projects. The next step beyond this PR will be to add the option to split mod source code in a similar fasion to provide compile time sided safety.

This PR also contains a number of refactors with the goal of contiuning to cleanup and modernise loom's codebase, there is still more that I want to do, but will handle that at a later date due to the growing size of this PR.


modmuss50:
This can now setup a dev env with split jars:



Im unsure as to why the fonts aren't loading, hopefully it doesnt turn out to be a blocking issue.

src/test/groovy/net/fabricmc/loom/test/unit/MinecraftJarSplitterTest.groovy
```diff
@@ -0,0 +1,69 @@
...
+
+import net.fabricmc.loom.configuration.providers.BundleMetadata
+import net.fabricmc.loom.configuration.providers.minecraft.MinecraftJarSplitter
+import spock.lang.Specification
+
+class MinecraftJarSplitterTest extends Specification {
+    public static final String CLIENT_JAR_URL = "https://launcher.mojang.com/v1/objects/7e46fb47609401970e2818989fa584fd467cd036/client.jar"
+    public static final String SERVER_BUNDLE_JAR_URL = "https://launcher.mojang.com/v1/objects/125e5adf40c659fd3bce3e66e67a16bb49ecc1b9/server.jar"
+
+    public static final File mcJarDir = new File("build/testjars/")
```
Juuxel: Should this be a temp dir or in one? `build` in the jvm run location sounds unsafe

modmuss50: Good catch, it was easy to find when first testing it 👍

PR #564, opened 2021-12-29
fix yarn mc version check for wildcards

isXander:
when using a wildcard for yarn build version, this check fails


modmuss50:
My longer term plan is to use: https://github.com/FabricMC/yarn/pull/2948#pullrequestreview-841404921

This might work for now tho.

modmuss50:
Closing this as it was superseeded by the 0.11 changes.

PR #567, opened 2022-01-01
IntelliJ Code-Style Config

shartte:
I can see that you attempted to add this to .editorconfig. I don't know if this works for anyone? If it does, it certainly does absolutly nothing for me and routline fails to apply the correct order.

The files in this PR however, will be picked up automatically and applied correctly.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,31 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JavaCodeStyleSettings>
+      <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
+      <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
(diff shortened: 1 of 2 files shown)


modmuss50:
The editorconfig file works great for me, im not sure if its plugin I added but I think it comes enabled by default with idea.

haykam821:
I use [this extension](https://github.com/editorconfig/editorconfig-vscode) for editor config support on Visual Studio Code. As far as I'm aware, there's no way to use IntelliJ code style configurations on Visual Studio Code.

shartte:
It even says this in the editor footer, and _STILL_ uses the incorrect import order, without anything else applied project-local:

grafik

After actually copying the files in this PR over from my local fabric-api project, it finally _does_ work.

p.s.: I am on the latest IntelliJ version, but this has never worked right for me, even months ago.

Juuxel:
Also for import ordering, I've noticed that idea likes to put them in its default order when importing, and then I have to force the .editorconfig ordering with ctrl+alt+O / `optimise imports`

modmuss50:
Closing this as the editorconfig and spotless settings are working well, we dont use this on any other projects. Thanks.

PR #569, opened 2022-01-11
Configure ValidateAccessWidenerTask.targetJar lazily

Juuxel:
This should prevent crashes if the `validateAccessWidener` task is initialised manually (by other plugins or scripts iterating all tasks, for example) before Loom's `afterEvaluate`s init the dep managers.

(The example I gave is an actual issue someone else ran into in the wild)


modmuss50 (approved):
Seems I fixed this by accident in 0.11 👍

PR #573, opened 2022-01-15
Kotlin metadata annotation remapping

### LINKED ISSUE #365: kotlinx.serialization generated serializers not working in remapped jar
I have a class annotated with `@Serializable` right [here](https://github.com/bluefireoly/fabrikmc/blob/main/fabrikmc-core/src/main/kotlin/net/axay/fabrik/core/math/FabrikPosition.kt), which means that a serializer for that class will be generated when building the project with gradle.

Now the problem is that in every project depending on my API (and using the remapped jar, because of `modImplementation` or because this is the only jar I uploaded to the maven repository) cannot use the serializer generated by the compiler plugin for kotlinx.serialization.

Inside of IntelliJ it just shows that the serializer was not found. When building with Gradle the following error comes up:
`java.lang.AssertionError: Class with incorrect id found: expected net/axay/fabrik/core/math/FabrikPosition.$serializer, actual net/axay/fabrik/core/math/FabrikPosition.serializer`
It seems like there is a problem with the class id.

Why do I know that the remapped jar causes this problem? I assume that this is the case because the serializer is not missing (working normally) in the dev jar.

### LINKED ISSUE #558: Kotlin library jar in intermediary gets remapped to wrong mappings
When adding a Kotlin mod library using `modImplementation`, and this library publishes its sources, you currently cannot use this library with different mappings than the ones that library chose to use for its own development.

**in the maven repo** there are:
- the actual library jar, correctly remapped to intermediary
- the sources jar, not remapped because this is not supported for Kotlin sources

**during compilation**:
- a jar remapped from the intermediary jar *not* to the mappings chosen in the current environment, but to the mappings used in the sources jar, is used (which is unwanted)
- this causes incompatible type errors, making it impossible to use the Kotlin library with other mappings than those used when this library was developed

For example, if my library uses **Yarn** mappings, others can only use it if they use Yarn as well, they cannot use it with **Mojmap**.

modmuss50:
This is untested as setting up a reprodiction case for the real issue is a bit of a pain. Remapping class names should be working, type and values is a bit of a pain as there is no desc. 

Do we want kotlin in loom's source? It may be possible to use the internal kotlin classes from java but might be a pain. Gradle includes kotlin so there is minimal overhead from doing this. Moving this out into its own library might be a better solution. I do not plan on using kotlin for anything else in loom.

Hopefully fixes #365 and #558 would be nice for someone to give this a test.

This is my first go at Kotlin, so if there is a better way of doing something please let me know. Thanks @Juuxel for providing a starting point.


build.gradle
```diff
@@ -19,6 +20,11 @@ tasks.withType(JavaCompile).configureEach {
 	it.options.release = 17
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+	kotlinOptions {
+		jvmTarget = "16" // Kotlin doesnt support 17!?
```
jakobkmar: 1.6.0+ (most recent version is `1.6.10`) is required for java 17

modmuss50: Ah, ok. We are tied to the version of kotlin gradle uses here sadly. The latest nightly uses 1.6.10 but not 7.3 this seems to work but does output a warning.

build.gradle
```diff
@@ -100,12 +111,22 @@ dependencies {
...
 
 	compileOnly 'org.jetbrains:annotations:23.0.0'
 }
 
+configurations.all {
+	if (name == "compileOnly") return
+
+	dependencies.removeIf {
+		// Rely on the kotlin version gradle uses everywhere else
+		group == "org.jetbrains.kotlin"
```
Juuxel: You can do `kotlin.stdlib.default.dependency = false` in `gradle.properties` to disable the stdlib dependency added by the Kotlin plugin

src/test/kotlin/net/fabricmc/loom/test/kotlin/KotlinClassMetadataRemappingAnnotationVisitorTest.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import org.junit.jupiter.api.Test
+
+// See: https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/test/kotlinx/metadata/test/MetadataSmokeTest.kt#L67
+class KotlinClassMetadataRemappingAnnotationVisitorTest {
+
+    @Test
+    fun simpleTest() {
+
+    }
```
Juuxel: Does this actually do anything?

build.gradle
modmuss50: I will try that, however its needed to compile against, but not be used at runtime/exposed via the pom.

src/main/kotlin/net/fabricmc/loom/kotlin/remapping/RemappingKmClassVisitor.kt
```diff
@@ -0,0 +1,157 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.kotlin.remapping
+
+import kotlinx.metadata.*
+import org.objectweb.asm.commons.Remapper
+
+class RemappingKmClassVisitor(private val remapper: Remapper, delegate: KmClassVisitor?): KmClassVisitor(delegate) {
```
Juuxel:
```suggestion
class RemappingKmClassVisitor(private val remapper: Remapper, delegate: KmClassVisitor?) : KmClassVisitor(delegate) {
```
Consistent code style please (you might want to consider using [ktlint](https://github.com/pinterest/ktlint) in some form or another)

src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
```diff
@@ -0,0 +1,119 @@
...
+        val header = KotlinClassHeader(kind, metadataVersion, data1, data2, extraString, packageName, extraInt)
+        return KotlinClassMetadata.read(header)
+    }
+
+    private fun writeClassHeader(header: KotlinClassHeader) {
+        val newNode = AnnotationNode(api, desc)
+        newNode.values = this.values.toMutableList()
+
+        newNode.run {
+            for (i in 0 until values.size step 2) {
```
Juuxel:
```suggestion
            for (i in values.indices step 2) {
```
A bit cleaner imo

build.gradle
Juuxel: Yeah, that's what it should do in this context. Gradle already adds a kotlin stdlib dependency I think?

src/test/kotlin/net/fabricmc/loom/test/kotlin/KotlinClassMetadataRemappingAnnotationVisitorTest.kt
modmuss50: No I was going to write unit tests for it and clearly got bored. 😂

src/main/kotlin/net/fabricmc/loom/kotlin/remapping/RemappingKmClassVisitor.kt
modmuss50: ktlint works with spotless, thanks for the suggestion 👍

build.gradle
modmuss50: Yeah, that seems to have worked, thanks 👍

PR #574, opened 2022-01-16
Add server only jar configuration option.

modmuss50:
Add an option to have a server only dev env, also further cleanup.

```
loom {
      serverOnlyMinecraftJar()
   }
```


src/main/java/net/fabricmc/loom/configuration/providers/minecraft/SplitMinecraftProvider.java
```diff
@@ -45,20 +44,20 @@ public SplitMinecraftProvider(Project project) {
...
-		return List.of(minecraftClientOnlyJar.toPath(), minecraftCommonJar.toPath());
+		return List.of(minecraftClientOnlyJar, minecraftCommonJar);
 	}
 
 	@Override
 	public void provide() throws Exception {
 		super.provide();
 
-		boolean requiresRefresh = isRefreshDeps() || !minecraftClientOnlyJar.exists() || !minecraftCommonJar.exists();
+		boolean requiresRefresh = isRefreshDeps() || !Files.exists(minecraftClientOnlyJar) || !Files.exists(minecraftCommonJar);
```
Juuxel:
```suggestion
		boolean requiresRefresh = isRefreshDeps() || Files.notExists(minecraftClientOnlyJar) || Files.notExists(minecraftCommonJar);
```

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftJarConfiguration.java
```diff
@@ -0,0 +1,133 @@
...
+		List.of("client", "server")
+	),
+	SERVER_ONLY(
+		ServerOnlyMinecraftProvider::new,
+		IntermediaryMinecraftProvider.ServerOnlyImpl::new,
+		NamedMinecraftProvider.ServerOnlyImpl::new,
+		ProcessedNamedMinecraftProvider.ServerOnlyImpl::new,
+		SingleJarDecompileConfiguration.ServerOnly::new,
+		List.of("server")
+	),
```
Juuxel: I'm very doubtful that there needs to be yet another set of these minecraft providers. From this class I can see references to 12 separate MinecraftProvider implementation classes, in addition to the different minecraft provider base types for different mapping types, which really feels like bad design.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/SplitMinecraftProvider.java
modmuss50: Ah I didn’t know that was a thing. Idea could do with a hint for this. 👍

Chocohead: Strictly speaking they're not complete opposites, in the case where it is not possible to determine both `exists` and `notExists` would return `false`. Doesn't matter in most cases though

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftJarConfiguration.java
modmuss50: Yes, I somewhat agree with you, im not super happy about the amount of classes and the mess of generics just bellow this. However having everything split out does make it easier to see whats going and less brittle. Adding any more would make the class count explode however I cannot think of any more that might be needed. Moving their construction all into this single location vastly helps as well.

I will have another pass at it to try and reduce the amount of classes, but I dont really fancy rewriting it all again just now.

PR #580, opened 2022-01-24
Add support for mixins written in groovy

patokeni:
Mixins written in groovy or compiled by groovy tasks won't get their refmaps in correct places. This adds support for processing them.


modmuss50 (approved):
Looks easy enough, as with the scala and kotlin ones not something that should be recommended but if it works great.

PR #586, opened 2022-01-26
Setup remapped variants in afterEvaluate

modmuss50:
Works around issues caused by creating the lazy remap tasks too early.


Juuxel (approved):
> Works around issues caused by creating the lazy remap tasks too early.

Not to mention possible problems if the tasks are never configured at all

PR #588, opened 2022-01-27
Add option to isolate the tiny remapper instance between sub-projects.

modmuss50:
```
remapJar {
   remapperIsolation = true
}
```


Juuxel (approved):
Seems to work around the issue. I still think it'd be better to not let files float around across jars even with the default config, however.

PR #590, opened 2022-01-29
Add file mappings layer

Juuxel:
Supersedes #454.

- The new layer supports any (single-file or Enigma) format supported by mapping-io
- The files can be nested in a jar/zip or standalone bare files
  - Detected automatically
- The default settings are geared towards Yarn, but they can be changed:
  - Source namespace: intermediary (= mappings get merged through intermediary names)
  - Mapping path inside zip: `mappings/mappings.tiny`
  - Default fallback namespaces for formats without namespaces: `intermediary -> named`
- The mappings can be read from any `FileSpec`

TODO:
- Naming
  - [x] `fileMappings` vs `mappings`
  - [ ] Should `mappingPath` be called something like `innerMappingPath` to distinguish it from an actual path to an outer mapping file
- Tests
  - [x] Enigma test
  - [x] Bare file test
  - [x] Different source name test
  - [x] Different fallback namespaces test

Example code:

```groovy
mappings loom.layered {
    mappings("net.fabricmc:yarn:$project.yarn_version:v2")
    mappings(file("proguardMappings.txt")) {
        fallbackNamespaces("named", "official")
        mergeNamespace 'official'
        // or mergeNamespace(net.fabricmc.loom.api.mappings.layered.MappingsNamespace.OFFICIAL)
    }
    mappings(file("my_mappings.zip")) {
        enigmaMappings()
        mappingPath("/mappings")
    }
}
```


src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpecBuilderImpl.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	@Override
+	public FileMappingsSpecBuilderImpl mappingPath(@Nullable String mappingPath) {
+		this.mappingPath = mappingPath;
+		return this;
+	}
+
+	@Override
+	public FileMappingsSpecBuilderImpl namespaces(String sourceNamespace, String targetNamespace) {
```
shedaniel: Would it be better to name this `fallbackNamespaces(...)`? Considering that these are only used for fallbacks. This method would be confusing alongside `sourceNamespace(...)` below

Juuxel: I thought of that, but it would also feel weird providing "fallbacks" when you know the format won't have namespaces of its own. I can change it though.

`sourceNamespace` is also a bit odd, calling it "merge namespace" could be clearer.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpec.java
```diff
@@ -0,0 +1,55 @@
...
+		return new FileMappingsLayer(fileSpec.get(context), mappingPath, fallbackSourceNamespace, fallbackTargetNamespace, mappingFormat, mergeNamespace);
+	}
+
+	// Enums don't have a consistent hash code
+	@Override
+	public int hashCode() {
+		String mappingFormat = this.mappingFormat != null ? this.mappingFormat.name() : null;
+		String mergeNamespace = this.mergeNamespace.name();
+		return Objects.hash(fileSpec, mappingPath, fallbackSourceNamespace, fallbackTargetNamespace, mappingFormat, mergeNamespace);
+	}
```
modmuss50: Humm, thats a bit annoying, Might be best to use a string name? A string is easier to use from the build script as well.

This applies for both the format and namespace, it also makes a bit more resilient to possible changes in MappingIO.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpecBuilderImpl.java
```diff
@@ -0,0 +1,83 @@
...
+
+import net.fabricmc.loom.api.mappings.layered.MappingsNamespace;
+import net.fabricmc.loom.api.mappings.layered.spec.FileMappingsSpecBuilder;
+import net.fabricmc.loom.api.mappings.layered.spec.FileSpec;
+import net.fabricmc.mappingio.format.MappingFormat;
+
+public class FileMappingsSpecBuilderImpl implements FileMappingsSpecBuilder {
+	private static final String DEFAULT_MAPPING_PATH = "mappings/mappings.tiny";
+	private static final String DEFAULT_FALLBACK_SOURCE_NAMESPACE = MappingsNamespace.INTERMEDIARY.toString();
+	private static final String DEFAULT_FALLBACK_TARGET_NAMESPACE = MappingsNamespace.NAMED.toString();
```
modmuss50: Nit: Any reason to have these as static fields? Just set the default value in the none static fields? Fine either way tbh

src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	FileMappingsSpecBuilder fallbackNamespaces(String sourceNamespace, String targetNamespace);
+
+	/**
+	 * Marks that the file contains Enigma mappings.
+	 * Because they are stored in a directory, the format cannot be auto-detected.
+	 *
+	 * @return this builder
+	 */
+	FileMappingsSpecBuilder enigmaMappings();
```
modmuss50, line 72: I would just expose a `FileMappingsSpecBuilder format(String format)`

`format 'enigma'` isnt that much to ask for imo?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpec.java
Juuxel: The mapping format is only internal so I'd say it doesn't matter that much. Really, I could just change it to `boolean enigma` since it's either `MappingFormat.ENIGMA` or null.

The merge namespace needs to be exposed so it might actually be nicer to also accept a string and just validate that it's either `official`‚ `intermediary` or `named`, then convert it to a `MappingsNamespace` in `getSourceNamespace`.

src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
Juuxel: The format can be automatically inferred for all the other formats by mio I think (this is a directory so it's a bit different). Also see above, changing this to `boolean enigma` would simplify the hash code

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpecBuilderImpl.java
Juuxel: Not really, I can inline them too

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpec.java
modmuss50: Yes, that a good idea, a boolean is fine 👍 And I think using a string and validating it would be cleaner.

Juuxel: Now it supports both a `MappingsNamespace` (since it's safer and an api enum so people can use that) and a validated string (both are internally a string). The Enigma thing is a boolean now.

modmuss50 (approved):
Thinking about it, this seems easy enough to into 0.11 as it doesnt change anything existing and is well tested.

PR #593, opened 2022-02-08
Intermediate mapping handling improvements.

modmuss50:
- Add API to allow other plugins to supply their own intermediate mappings.
- Add hacky option to run with no intermediate mappings.

This is far from perfect as loom still refrences `intermediary` throughout, however its should be good enough for most use cases without another massive refactor.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -35,9 +35,11 @@ import org.gradle.api.publish.maven.MavenPublication;
 import org.jetbrains.annotations.ApiStatus;
 
 import net.fabricmc.loom.api.decompilers.DecompilerOptions;
+import net.fabricmc.loom.api.mappings.intermediate.IntermediateMappingsProvider;
 import net.fabricmc.loom.api.mappings.layered.spec.LayeredMappingSpecBuilder;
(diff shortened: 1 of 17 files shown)


thecatcore (approved):

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -130,6 +132,20 @@ default void interfaceInjection(Action<InterfaceInjectionExtensionAPI> action) {
 	 */
 	Property<Boolean> getEnableTransitiveAccessWideners();
 
+	@ApiStatus.Experimental
+	IntermediateMappingsProvider getIntermediateMappingsProvider();
+
+	@ApiStatus.Experimental
+	<T extends IntermediateMappingsProvider> void setIntermediateMappingsProvider(Class<T> clazz, Action<T> action);
```
Juuxel, line 142: Why not allow directly setting the instance considering it's constructed immediately as far as I can see?

modmuss50: Sure, its easy enough to add. Will keep this version of it as it does hide away the need to use project.getObjects() to create the instance, making it nicer to call from a build script.

Juuxel: Oh I see it uses that abstract `Property` pattern, didn't see that before

PR #599, opened 2022-02-19
Add unpick support to layered file mappings, add direct url support to FileSpec.

modmuss50:
This can go into 0.11 as unpick is now disabled by default, you will need to add your own constants jar to the compileOnly classpath.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
@@ -76,7 +76,7 @@ public interface FileMappingsSpecBuilder {
 	 *
 	 * @return this builder
 	 */
-	FileMappingsSpecBuilder disableUnpick();
+	FileMappingsSpecBuilder containsUnpick();
 
 	/**
 	 * Sets the merge namespace of this mappings spec.
--- a/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
@@ -72,6 +72,13 @@ public interface FileMappingsSpecBuilder {
 	FileMappingsSpecBuilder enigmaMappings();
 
 	/**
+	 * Any unpick data in a zip file will be ignored.
+	 *
+	 * @return this builder
+	 */
+	FileMappingsSpecBuilder disableUnpick();
+
+	/**
 	 * Sets the merge namespace of this mappings spec.
 	 *
 	 * <p>The merge namespace is the namespace that is used to match up this layer's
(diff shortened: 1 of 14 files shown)


src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileMappingsSpecBuilder.java
```diff
@@ -71,6 +71,13 @@
 	 */
 	FileMappingsSpecBuilder enigmaMappings();
 
+	/**
+	 * Any unpick data in a zip file will be ignored.
```
Juuxel: ignored or used?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/HttpFileSpec.java
```diff
@@ -0,0 +1,54 @@
...
+
+public record HttpFileSpec(String url) implements FileSpec {
+	private static final Logger LOGGER = LoggerFactory.getLogger(HttpFileSpec.class);
+	@Override
+	public Path get(MappingContext context) {
+		try {
+			Path output = Files.createTempFile("loom", "httpfile");
+			Files.delete(output);
+
+			DownloadUtil.downloadIfChanged(new URL(url), output.toFile(), LOGGER);
```
Juuxel: Not sure if this is the best since it seems to never cache (considering it uses a temp file)

src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileSpec.java
```diff
@@ -63,6 +64,10 @@ static FileSpec create(Object o) {
 		Objects.requireNonNull(o, "Object cannot be null");
 
 		if (o instanceof CharSequence s) {
+			if (s.toString().startsWith("https://")) {
```
Juuxel: Maybe support for `http://` too (though no one should be using that), as well as `URL`s?

modmuss50: Only reason I didn’t was because gradle doesn’t allow it by default for maven repos. No major disadvantage especially as this is only used for mappings?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/HttpFileSpec.java
modmuss50: It doesn’t cache at this level, it’s only requested when the mapping layers are built up. I might change this 👍

src/main/java/net/fabricmc/loom/api/mappings/layered/spec/FileSpec.java
Juuxel: Ah I see. `URL` object support for this protocol would still be nice, although I don't expect that many people to use it

Juuxel (approved):

PR #605, opened 2022-02-28
Fix mappingsConstants dependency being added even if already present

IotaBread:


src/main/java/net/fabricmc/loom/configuration/providers/mappings/MappingsProviderImpl.java
```diff
@@ -158,13 +158,17 @@ private void setup(MinecraftProvider minecraftProvider, Path inputJar) throws IO
 
 	public void applyToProject(Project project, DependencyInfo dependency) {
 		if (hasUnpickDefinitions()) {
-			String notation = String.format("%s:%s:%s:constants",
-					dependency.getDependency().getGroup(),
-					dependency.getDependency().getName(),
-					dependency.getDependency().getVersion()
-			);
+			if (project.getConfigurations().getByName(Constants.Configurations.MAPPING_CONSTANTS).getDependencies().isEmpty()) {
```
modmuss50: This is always going to be empty as you add to it within this check. You need to check if the dependency is a layered.

Tests would be ideal, but prob something I should have added to start with.

modmuss50 (changes requested):
Still not the correct fix imo, this can cause silent failures.

Like I said this should just skip over layered mapping dependencies. Will be a much cleaner and peformant fix.

Issue #600, opened 2022-02-22
No way to create custom mod* remapped configurations.


NoComment1105:
Title says it all

![image](https://user-images.githubusercontent.com/67918617/155178936-5005ca4b-02c3-4fbb-bacc-8455cb6cc7eb.png)
![image](https://user-images.githubusercontent.com/67918617/155179030-a9747e8c-bf5f-427e-a997-c45d1885b8b9.png)
```bash
/home/nocomment/Documents/GitHub/NoComment1105/maybe-more-data/src/testmod/java/maybemoredata/testmod/MaybeMoreDataTestmod.java:34: error: cannot access class_3614
            TEST_BLOCK = Registry.register(Registry.BLOCK, new Identifier("maybemoredata", "test_block"),  new Block(FabricBlockSettings.of(Material.STONE).requiresTool().strength(2f, 4f)));
                                                                                                                                        ^
  class file for net.minecraft.class_3614 not found
```

NoComment1105:
o7, Thank you! Please let me know when this is possible, I look forward to it

modmuss50:
This was resolved with: https://github.com/FabricMC/fabric-loom/commit/da2992e7d9547a1ab14e500ba5ca188733f1bc3f

PR #606, opened 2022-02-28
First pass at spilt env sourcesets.

modmuss50:
This is the final stage of my gloal to split the client and server jar to provide compile time sidedness saftey.

Very WIP, breaks stuff when enabled.

`main` becomes common only, new `client` sourceset contains the client only stuff.

# TODO (un-unordered)

- [ ] Loader support, I have a hacked version [here](https://github.com/modmuss50/fabric-loader/tree/split_game_jars) to aid with testing loom
- [x] Better support for multiproject builds
- [x] Output jars should contain a list of client only entries.
- [ ] When depending on a mod that has client only entries it should be re-split and added to the correct sourceset
- [ ] Minecraft dev plugin freaks out badly, making it unseable, invistage and report to them.
- [x] Fabric API support, should just depend on https://github.com/FabricMC/fabric/pull/1918
- [x] Add client only annotation to all client only classes when building? Would help people not using a split setup.
- [x] Try and improve the situation with afterEvaluate, the new sourceset should be added when the loom extension method is called. - Setup can happen later.
- [ ] Client only unpick constants? Prob not worth the effort, at least to start with.
- [ ] Good tests, including running on a prod server and depending on mods.


modmuss50:
I think im going to merge this in, and solve splitting mod dependencies at a later date as it really needs https://github.com/FabricMC/fabric-loom/issues/600 to be solved first.

PR #608, opened 2022-03-04
Add DSL to configure mod class path groups.

modmuss50:
## Mod Classpath Settings
This provides a simple way to populate the [`fabric.classPathGroups`](https://github.com/FabricMC/fabric-loader/pull/585) system property, via gradle. This is only required when your mod spans across multiple classpath directories, such as when using split client/common sourcesets.

```gradle
loom {
    mods {
        modid {
            sourceSet sourceSets.main
            sourceSet sourceSets.client
        }
    }
}
```

## Known issues

IDEs such as intelij use a diffrent output location when building without gradle, as far as I can tell gradle has no knowledge where this goes, further investigation needs to be done to figure out what can be done to support building and running only through the IDE. If anyone has any ideas I would appreciate the input. 👍 

## Example:

With this change it is possible to run TechReborn & RebornCore with split client/common sourcesets. This was done with a prototype version of fabric loader.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/ModSettings.java
+++ b/src/main/java/net/fabricmc/loom/api/ModSettings.java
@@ -1,0 +1,58 @@
+/*
+ * This file is part of fabric-loom, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2022 FabricMC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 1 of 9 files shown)


src/main/java/net/fabricmc/loom/api/ModSettings.java
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public void sourceSet(SourceSet sourceSet) {
+		final SourceSetOutput output = sourceSet.getOutput();
+		final File resources = output.getResourcesDir();
+
+		getClasspath().addAll(output.getClassesDirs().getFiles());
+
+		if (resources != null) {
+			getClasspath().add(resources);
+		}
```
Juuxel: Maybe do these lazily so that if the source set changes before generating the run configs (this can happen by shuffling around things in the buildscript), it'll take that into account

src/main/java/net/fabricmc/loom/api/ModSettings.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.File;
+
+import org.gradle.api.Named;
+import org.gradle.api.provider.ListProperty;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.SourceSetOutput;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * A {@link Named} object for setting mod realted values. The {@link Named#getName()} should match the modid.
```
Juuxel:
```suggestion
 * A {@link Named} object for setting mod-related values. The {@linkplain Named#getName() name} should match the mod id.
```

Juuxel:
See also our prior work for this feature on Forge in Arch Loom:

<https://github.com/architectury/architectury-loom/blob/a4a61aaf70da8d33215333045956e939e20e2c3b/src/main/java/net/fabricmc/loom/api/ForgeExtensionAPI.java#L137-L150>
<https://github.com/architectury/architectury-loom/blob/dev/0.10.0/src/main/java/net/fabricmc/loom/api/ForgeLocalMod.java>

Juuxel (changes requested):

modmuss50:
@Juuxel I have made your suggested changes, I think this should now be able to fully replace your `ForgeLocalMod` class.

I still need to solve the issue of this only beign able to use gradle's output directories. Does arch loom work when building and running through the IDE? (Im looking for existing code for inspiration :D ) I couldnt see this beign the case. EDIT: Basic idea support was easy enough, will look at eclipse when I have the energy to loose.

Juuxel:
> Does arch loom work when building and running through the IDE? (Im looking for existing code for inspiration :D )

Arch loom itself should, not sure if this functionality since I haven't touched it myself (this is shedaniel's code). We do put idea into "build and run using gradle" by default since it doesn't work with the bytecode transformer we use for multiplatform projects, so I don't think anyone has tested this

shedaniel:
Is `ForgeLocalMod` related to this at all? That class is used to setup the forge environment variable `MOD_CLASSES` only, and isn’t anything else.

Juuxel:
@shedaniel It's also used for the classpath like this PR (albeit in a different form since it's that env variable)

PR #613, opened 2022-03-12
Allow supporting different mod metadata formats

TheGlitch76:
This de-hardcodes almost every reference to fabric.mod.json, and provides an API to add support for arbitrary mod metadata formats. Multiple metadata formats can exist, but mods are expected to only use one kind of metadata. If an uncompiled project uses more than one kind of mod metadata, the metadata to read is undefined and selected at random. Reading a jar file with multiple kinds of metadata will throw a runtime exception.

`fabric.mod.json` generation for JiJ mods was skipped to reduce the complexity of this PR, and because it isn't currently needed for my use-case.

The test suite passes, but I would appreciate another careful eye to be sure that I didn't accidentally remove behavior or sanity checks.

I can be contacted directly as `glitch#3274` on Discord.


modmuss50:
What is your use-case? This is a massive change with added complexity.

This stuff did need cleaning up, but its not something I have given much thought about tbh especially as we have no need for it right now.

modmuss50:
This seems generally ok if your only goal is to rename the fabric.mod.json, however I dont think its terribly useful beyond that usecase and wont be sutabile for possible future needs.

The ModMetadataHelper.Metadata makes a lot of assumptions about the metadata format, making it no use if you have another format that does not map well to the existing fabric.mod.json format. I dont know what our V2 format is going to look like when it happens, but I wouldn't be confident using this API to support it. I believe forge's format supports multiple mods for example (Correct me if im wrong?)

Right now loom isnt very clever about opening/reading something it has already read. Perf improvements could be made by caching this data. Most likely out of scope for this PR, should possibly be done before allowing other formats?

Something that is ongoing is trying to improve the testability of loom via unit tests, as intergration tests are painfully slow, this does seems like a good step towards making that easier. 👍 

I think the real solution to these issues is a lot bigger than this PR, and should be done in multiple stages. The goal of this PR to me just seems to allow you to you remove fabric's name from everything and then put the maintaince burden of doing so on our backs, without actually solving any of the existing issues.

Im happy to discuss this further, but I quite clearly wont be spending a lot of time and energy into something that doesnt benefit me/fabric.

TheGlitch76:
Discussed on Discord--will be exploring other solutions to this.

PR #631, opened 2022-04-20
Create MixinMappingsService for every target MC version.

### LINKED ISSUE #630: Mixin remapping incorrect in multi-version projects
https://github.com/KosmX/breakFabricLoom
//There is a branch with Yarn mappings

In the demo, there are two modules, for MC 1.16.5 and one for MC 1.18.2.
One of the two jars will be incorrect. For me, both files were remapped with Minecraft 1.16.5 mappings.

If there are multiple mc versions in one project, all mixins will be remapped with the __same__ mappings.

KosmX:
Attempt to fix #630

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/service/MixinMappingsService.java
+++ b/src/main/java/net/fabricmc/loom/task/service/MixinMappingsService.java
@@ -34,6 +34,7 @@ import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.util.service.SharedService;
 import net.fabricmc.loom.util.service.SharedServiceManager;
 import net.fabricmc.tinyremapper.IMappingProvider;
+import net.fabricmc.loom.configuration.providers.minecraft.MinecraftProvider;
 
 public final class MixinMappingsService implements SharedService {
 	private final SharedServiceManager sharedServiceManager;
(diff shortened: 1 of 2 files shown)


src/main/java/net/fabricmc/loom/task/service/MixinMappingsService.java
```diff
@@ -47,13 +48,13 @@ public static File getMixinMappingFile(Project project, SourceSet sourceSet) {
...
-		getService(SharedServiceManager.get(project)).mixinMappings.add(mixinMapping);
+		getService(SharedServiceManager.get(project), extension.getMinecraftProvider()).mixinMappings.add(mixinMapping);
 
 		return mixinMapping;
 	}
 
-	static synchronized MixinMappingsService getService(SharedServiceManager sharedServiceManager) {
-		return sharedServiceManager.getOrCreateService("MixinMappings", () -> new MixinMappingsService(sharedServiceManager));
+	static synchronized MixinMappingsService getService(SharedServiceManager sharedServiceManager, MinecraftProvider minecraftProvider) {
+		return sharedServiceManager.getOrCreateService("MixinMappings-" + minecraftProvider.minecraftVersion(), () -> new MixinMappingsService(sharedServiceManager));
```
modmuss50: I think this is along the correct lines, however I think it should be using `extension.getMappingsProvider().mappingsIdentifier()` as used in a few lines above insead of the minecraft version.

KosmX: `MappingsProvider` does not have `mappingsIdentifier()`.  
I need `MappingsProviderImpl`.  

Or should I pass a `String` parameter?

modmuss50 (approved):
This looks good now. 👍

PR #634, opened 2022-04-25
Fix #633 Use reference-counted FileSystemHandler for jars

### LINKED ISSUE #633: Loom and TinyRemapper can cause ClosedFileSystemException when both open the same zip file
Note: This is already being worked on but I am documenting it here in case anyone else runs into the issue before it is resolved.

## Description

I am running into this issue in my gradle multi-project, where multiple instances of plugins are all running at once.
My `:Fabric:remapJar` or `FabricApi:remapJar` tasks fail with an exception when I run `./gradlew build` for the whole project.
Stack trace: https://gist.github.com/mezz/da5efaf1fd9a2131e5cdf008c8d036de

## Analysis

The issue is that `FileSystem` (the java nio class used to open zip files) is basically a global resource within a single JVM instance, and expects everyone on the JVM to coordinate together.
Loom and TinyRemapper run on the same JVM instance, so if they both open a zip file, and then one closes it, the other will get  a `ClosedFileSystemException` the next time it tries to use that FileSystem.

## Solution

The solution is to create some kind of global state or library that these gradle plugins can share, so that they can reference-count how many times the `FileSystem` has been opened or closed, and then only close it when there are no more references.

## Current Progress

@sfPlayer1 has implemented a solution for this in this TinyRemapper here: https://github.com/FabricMC/tiny-remapper/commit/4eb5a9bda8d819dc742fa86b735e1cda6bffb10c
This issue will be resolved when FabricLoom adopts the same pattern, so that the two plugins can share the global "fsRefsProvider" system property and coordinate their openings and closings of the `FileSystem`s.

mezz:
`FileSystemHandler` is copied from TinyRemapper, with permission from @sfPlayer1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,7 +85,7 @@ dependencies {
 	}
 
 	// tinyfile management
-	implementation ('net.fabricmc:tiny-remapper:0.8.2')
+	implementation ('net.fabricmc:tiny-remapper:0.8.4')
 	implementation 'net.fabricmc:access-widener:2.1.0'
 	implementation 'net.fabricmc:mapping-io:0.2.1'
 
--- a/src/main/java/net/fabricmc/loom/util/FileSystemHandler.java
+++ b/src/main/java/net/fabricmc/loom/util/FileSystemHandler.java
@@ -24,7 +24,6 @@
 
 package net.fabricmc.loom.util;
 
-
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.URI;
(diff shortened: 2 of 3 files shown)


Player3324 (approved):

modmuss50:
This was resolved here: https://github.com/FabricMC/fabric-loom/commit/651bcec276e99b6fb59a1587f5658ebcab7150dd

Thanks a lot anyway 👍

PR #637, opened 2022-05-01
Dont depend on other prepare tasks, only run after.

modmuss50:
This may fix some issues we have been seeing.


modmuss50:
Going to merge this, let me know if it breaks anything.

PR #639, opened 2022-05-04
Use TR's ref coutning FileSystemHandler

modmuss50:
I want to rewrite most of this zip handling in loom at somepoint, but this is hopefully a quick fix.


Player3324:
would need adjusting for https://github.com/FabricMC/tiny-remapper/pull/93 if we go for those changes

modmuss50:
I have further changes I plan make around zip access, going to merge this as-is to so I can contiune without the worry of bad conflicts.

PR #642, opened 2022-05-10
Fix mod settings for sourcesets from other projects.

modmuss50:
This needs some slightly better testing, there still seems to be issues with https://github.com/FabricMC/fabric/pull/2179


modmuss50:
Just down to an issue with extending mixins across sourcesets.

```
error: Superclass net.fabricmc.fabric.mixin.event.lifecycle.WorldMixin of net.fabricmc.fabric.mixin.event.lifecycle.client.ClientWorldMixin was not found in the hierarchy of target class net.minecraft.client.world.ClientWorld
```

PR #651, opened 2022-05-19
Preserve extraInt in kotlin metadata remapper

Johni0702:
The `extraInt` (`xs`, [1]) parameter of Kotlin's Metadata annotation does not become
part of the KmClass/KmLambda and must be preserved explicitly to not be lost
during remapping.

It does carry important information though, so we must not neglect it.
In particular, bit 7 indicates that the class is used in the scope of an inline
function and is therefore implicitly part of the public API. If this flag is
missing from a class which requires it, then it becomes impossible to use that
inline function because the Kotlin compiler will crash hitting an assertion
while trying to inline the required code [2].

Based on the descriptions in [1], it looks like none of this data should be
affected by our remapping in any way, so it seems safe to simply copy over the
value from the original annotation. And that's what this commit does.

[1]: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-metadata/extra-int.html
[2]: https://i.johni0702.de/sOpA8.png

### DIFF
--- a/src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
+++ b/src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
@@ -48,7 +48,7 @@ class KotlinClassMetadataRemappingAnnotationVisitor(private val remapper: Remapp
                 val klass = metadata.toKmClass()
                 val writer = KotlinClassMetadata.Class.Writer()
                 klass.accept(RemappingKmVisitors(remapper).RemappingKmClassVisitor(writer))
-                writeClassHeader(writer.write().header)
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):

modmuss50:
Im a little unsure as to why all the tests are failing, I dont think its realted to your changes.

modmuss50:
`The specified Gradle distribution 'https://services.gradle.org/distributions-snapshots/gradle-7.5-20220129032445+0000-bin.zip' does not exist.`

Oo seems gradle removes old snapshot builds, Im going to pull this change in 0.12 as that also includes some major fixes to the kotlin remapper.

modmuss50:
Manually pushed this change into 0.12: https://github.com/FabricMC/fabric-loom/commit/1b6fa615c48083b97afdda4fd350e43c67ae3a39

Thanks a lot 👍

PR #670, opened 2022-06-14
Added jetbrains annotations to test compile

jaredlll08:
So I'm not too sure on the specifics of this one, but basically if you clone 
```
https://github.com/jaredlll08/MultiLoader-Template
```
then add:
```java
package com.example.examplemod;

import net.minecraft.client.Minecraft;
import net.minecraft.client.gui.components.AbstractSelectionList;
import net.minecraft.client.gui.narration.NarrationElementOutput;

public class BadList<E extends AbstractSelectionList.Entry<E>> extends AbstractSelectionList<E> {

    public BadList(Minecraft mc) {

        super(mc, 0, 0, 43, 0, 50);
    }

    @Override
    public void updateNarration(NarrationElementOutput narrator) {

    }
}
```
to the Common project and try and build it, you get a very useless error message of:
```
> Task :Fabric:compileTestJava FAILED
An exception has occurred in the compiler (17.0.1). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
com.sun.tools.javac.code.Symbol$CompletionFailure: class file for org.jetbrains.annotations.Nullable not found
```

With absolutely no information as to what file errored or what even requires that annotation.

It only happens when running `compileTestJava` and changing the template to only bring in Common code when `compileJava` is called builds fine (although that is not a solution).

Changing the JB annotations to instead be `implementation`, which is what the [docs](https://www.jetbrains.com/help/idea/annotating-source-code.html) uses, fixes the issue.

As noted [here](https://discuss.gradle.org/t/compileonly-dependencies-are-not-available-in-tests/15366/6) `compileOnly` does not propagate to tests and `implementation` should be used instead.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -130,6 +130,7 @@ public final class CompileConfiguration {
 		project.getDependencies().add(Constants.Configurations.LOOM_DEVELOPMENT_DEPENDENCIES, Constants.Dependencies.DEV_LAUNCH_INJECTOR + Constants.Dependencies.Versions.DEV_LAUNCH_INJECTOR);
 		project.getDependencies().add(Constants.Configurations.LOOM_DEVELOPMENT_DEPENDENCIES, Constants.Dependencies.TERMINAL_CONSOLE_APPENDER + Constants.Dependencies.Versions.TERMINAL_CONSOLE_APPENDER);
 		project.getDependencies().add(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
+		project.getDependencies().add(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
 	}
 
 	public static void configureCompile(Project p) {


modmuss50:
I dont think this is the correct fix, compileOnly is used as these anntations are only available in dev, when debugging dev we want to make sure people dont accidentally try to use the library in a way that would cause a failiure in prod.

Im a little confused by your reproduction case, as you say to put the class into the common project, but the error you'r showing is from when trying to compile the fabrictest sourceset (that doesnt exist in your repo). I dont believe you are using loom for the common project. A minimal way to reproduce this in the fabric example mod, without having Forge or other plugins in the mix might be useful.

If the annotations are just missing from the test sourceset, I think fix here would to be also add the annotation on the `testCompileOnly` config, I assume it already makes it way onto the client sourceset (when enabled).

jaredlll08:
Reproduction using just fabric here:
https://github.com/jaredlll08/temp-fabric-issue-repo

That project is a 1:1 of this project https://github.com/FabricMC/fabric-loom/tree/dev/0.12/src/test/resources/projects/multiproject although it has been updated to 1.19 (despite what the branch name says), with the `BadList` class added (after converting to yarn) and adding https://github.com/jaredlll08/temp-fabric-issue-repo/blob/1.18/example/build.gradle#L13-L15 which pulls the `core` code into `example`'s built jar.

Just so we are on the same page (since I'm not too sure about the client source set), the new solution would be
```java
        project.getDependencies().add(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
        project.getDependencies().add(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
```


If you're happy with the repro and the new fix, I can make a proper commit with it tomorrow when I wake up.

modmuss50:
Thanks for the repo project,

This is not a loom issue. In your example sub project you are adding the main sourceset of the core project, to all JavaCompile tasks! (including the test source set that your not even using) https://github.com/jaredlll08/temp-fabric-issue-repo/blob/1.18/example/build.gradle#L13-L15 Removing this fixes the build.

Im not quite sure why you have this, but this isn't how it should be done. This is causing you to compile the same classes many times resulting in dupliate classes, and bad classpaths. This will and does cause havoc.

You already depend on the core project correctly, if you want to shade the outut the of the core project into your jar, this should be done by includng the output classes in the jar task.

Darkhax:
Thanks for the feedback modmuss, I think you're right here. After making your suggested change the problematic projects are building fine. @jaredlll08 You can see my crude implementation of this at https://github.com/Darkhax-Minecraft/Tips/commit/695cefe805196018130e9cb63e6902b4c7c539d8.

modmuss50:
Yeah, that looks better. You may need to add:

```gradle
loom {
	mods {
		modid {
			sourceSet sourceSets.main
			sourceSet project(":core").sourceSets.main
		}
	}
}
```

This will help fabric loader load the two sourcesets as one mod.

jaredlll08:
After more testing a reproduction is as simple as adding the `BadList` class to the test sourceset of the fabric example mod, no multiprojects required.

I have updated this PR to add the annotation to testCompile instead of moving it all to implementation.

modmuss50 (approved):

Juuxel (approved):

Issue #666, opened 2022-06-10
Use access widener version used by the loader


modmuss50:
Right now any addtions/updates to access widener require both the loader and loom to be updated. In the future we may want to update access widener at a faster pace. Loom should use the version of access widener the loader depends on.

This can be based of similar tech used to sperate the kotlin version for the kotlin remapper.

modmuss50:
Not going to change this for AW, classtweaker will do this.

PR #673, opened 2022-06-16
[VERY WIP] Access widener abstraction #666

modmuss50:
Very WIP, should compile but wont run.


modmuss50:
Im going to revist this at a later date. Possibly wont be needed for 0.13.

PR #678, opened 2022-06-26
Replace getRemapArchives & getSetupRemappedVariants with a gradle property.

### LINKED ISSUE #677: remapArchives and setupRemappedVariants are evaluted before project configuration.
This makes it impossible to set them before they are required.

A possible fix is supplying project names via a gradle property? I cannot think of a better way?

modmuss50:
getRemapArchives & getSetupRemappedVariants are evaluted before a build script has time to configure them, making them useless. This PR replaces them with a gradle property to be set in the projects `gradle.properties` file.


opl-:
Seems this change never got documented on https://wiki.fabricmc.net/documentation:fabric_loom

PR #681, opened 2022-07-09
Replace Download util.

modmuss50:
The current download util is a bit of a mess, and has caused a few issues, this PR replaces what we had with a (hopefully) more robust, testable and maintainable solution.

Features
- eTag support (It seems Mojang's CDN and our meta/maven do not support this right now)
- Age based caching.
- Sha1 based caching.
- HTTP(S) proxy settings should now be respected.
- The downloader has better awareness of offline mode and refresh deps, cleaning up the calling code a bit.
- Uses the new Java 11 HTTP client, provides async support for downloading multiple files
- Progress handling (TODO needs hooking up to gradle)
- Better compression support. Handled by [Methanol](https://mizosoft.github.io/methanol/) 
- Unit tested with a real web server. 

This PR isnt fully complete, just wanted to get it up for inital feedback and let the tests run.


src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -56,6 +56,7 @@ public interface LoomGradleExtensionAPI {
 
 	RegularFileProperty getAccessWidenerPath();
 
+	@Deprecated(forRemoval = true)
 	Property<Boolean> getShareRemapCaches();
```
Juuxel, line 60: This seems unrelated?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/URLFileSpec.java
```diff
@@ -35,18 +33,20 @@
 
 import net.fabricmc.loom.api.mappings.layered.MappingContext;
 import net.fabricmc.loom.api.mappings.layered.spec.FileSpec;
-import net.fabricmc.loom.util.DownloadUtil;
+import net.fabricmc.loom.util.download.DownloadException;
 
-public record URLFileSpec(URL url) implements FileSpec {
+public record URLFileSpec(String url) implements FileSpec {
```
Juuxel, line 39: The hashCode should probably be changed now (or maybe DlBuilder should support URLs?)

src/test/groovy/net/fabricmc/loom/test/unit/LoomMocks.groovy
```diff
@@ -34,12 +37,14 @@ class LoomMocks {
     static IntermediaryMappingsProvider intermediaryMappingsProviderMock(String minecraftVersion, String intermediaryUrl) {
         def minecraftVersionProperty = GradleTestUtil.mockProperty(minecraftVersion)
         def intermediaryUrlProperty = GradleTestUtil.mockProperty(intermediaryUrl)
+		def downloaderProperty = GradleTestUtil.mockProperty(Download.&create as Function)
```
Juuxel: There's something weird with the indents in this method (mixed tabs and spaces?)

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/URLFileSpec.java
modmuss50: URL.hashCode [makes a network request](https://stackoverflow.com/questions/2348399/why-does-java-net-urls-hashcode-resolve-the-host-to-an-ip).

I think DLBuilder should also take a URL 👍

src/test/groovy/net/fabricmc/loom/test/unit/LoomMocks.groovy
modmuss50: Humm, looks like we dont check this, will reformat all the groovy to use tabs separately.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/URLFileSpec.java
Juuxel: > URL.hashCode [makes a network request](https://stackoverflow.com/questions/2348399/why-does-java-net-urls-hashcode-resolve-the-host-to-an-ip).

Precisely my point, the workaround isn't needed anymore since it's a String now.

modmuss50:
Test failures all seem to be known (caused by previous 0.13 changes) going to merge this now. 👍

PR #686, opened 2022-07-19
Make remapping tasks use the archive file from the Jar API

Juuxel:
Remapping only works for sources jars at this point, not sure what's going on with code jars.

Other minor changes:
- Exception messages now include the absolute path of the jar file
- `RemapSourcesJarTask` now also uses descriptive wrappers instead of a direct `new RuntimeException(...)`


Juuxel:
Seems to be impossible to do unless a lot of the remapping optimisations are removed (this can't be done with TR sharing).

PR #689, opened 2022-07-25
Ensure that deps with injected ifaces are on both the compile and runtime classpath.

### LINKED ISSUE #687: Interface injection causes a guarenteed crash when used as a compile only dependency
If a mod uses interface injection, and you try to depend on it through gradle but don't want it in your runtime, you cannot launch the game. Interface injection puts the interface onto the class for dev sources, and launching the game makes the JVM attempt to load the interface, which does not exist at runtime, causing an instant class not found error.

This effectively means that no API mods can have interface injections, and also content mods with interface injections cannot have optional compatibility made on the end of other mods, since all dependencies are forced to be present at runtime. This is a pretty huge problem.

Possible solution:
* Separate the compile/run sources and only apply interface injection to the former (this was a suggestion from someone else and I'm not sure how reasonable it is to implement)

Hacky solution:
* Detect interface injection of compile only dependencies and, at runtime, add stub interfaces that contain no methods but match the classpath to prevent crashing.

modmuss50:
Fixes #687 See issue for detailed discussion.

PR to let the tests run.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
@@ -183,8 +183,9 @@ public class InterfaceInjectionProcessor implements JarProcessor, GenerateSource
 		List<InjectedInterface> result = new ArrayList<>();
 
 		for (RemappedConfigurationEntry entry : Constants.MOD_COMPILE_ENTRIES) {
-			// Only apply injected interfaces from mods that are part of the compile classpath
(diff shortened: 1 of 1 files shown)


Juuxel (approved):
This doesn't work for the compileOnly+localRuntime combo but that can be fixed later imo

modmuss50:
Pushed a fix that will handle jars across the 2 configurations into 0.13: https://github.com/FabricMC/fabric-loom/commit/a0187c95e46c1e787b37fca2b23f03e92a5efedc I will stick with this PR for 0.12. The 0.13 changes made it a bit easier to do as well.

PR #698, opened 2022-08-09
POC for a test that runs the client, and interacts with it.

modmuss50:
I think this would be better placed in its own project used to test fabric + api (with mods) in production like environment. 

Running this in CI should be possible as well with a little effort.


modmuss50:
Here is a video of it in action: https://www.youtube.com/watch?v=NMjEE94nATM

src/test/groovy/net/fabricmc/loom/test/integration/ClientRunTest.groovy
```diff
@@ -0,0 +1,69 @@
...
+class ClientRunTest extends Specification implements GradleProjectTestTrait, RobotTrait {
+	@Unroll
+	def "Run client (minecraft #version)"() {
+		setup:
+			def gradle = gradleProject(project: "minimalBase", version: PRE_RELEASE_GRADLE)
+			gradle.buildGradle << """
+                dependencies {
+                    minecraft "com.mojang:minecraft:${version}"
+                    mappings loom.layered() {
+					}
```
Juuxel, line 50: Uhh, mixed tabs and spaces?

modmuss50: Prob, I forgot to fix the formatter. I dont plan on merging this anytime soon tbh, it needs quite a bit of work to make it worth it.

modmuss50:
Fabric API now does this, I dont think this should be a loom thing.

PR #711, opened 2022-08-31
New progress tracker Fixes #710

### LINKED ISSUE #710: Outdated Java warning does not work on < Java 11
Happens because a library we use depends on J11 in its gradle module metadata:

https://pastebin.com/PZZjz4Qj

modmuss50:


build.gradle
```diff
@@ -77,7 +77,6 @@ dependencies {
 	implementation ('org.ow2.asm:asm-commons:9.3')
 	implementation ('org.ow2.asm:asm-tree:9.3')
 	implementation ('org.ow2.asm:asm-util:9.3')
-	implementation ('com.github.mizosoft.methanol:methanol:1.7.0')
```
modmuss50, line 80: I think for the next loom version I will shade + repackage some of these libs.

PR #712, opened 2022-09-03
Better way to handle include configuration

Niko-sk2x:
Instead of reinventing resolving the configuration manually, just resolve it and actually make the `remapJar` task depend on the configuration. Depending on the configuration ensures that gradle runs the subproject tasks that generate the outputs. This wouldn't work with the dependencies being specified as inputs.

This handles project dependencies correctly, including specifying a target configuration, attributes and using artifact transforms.

I'm not sure which branch I should target, so it's on 1.0.

This is technically a breaking change because project dependencies may be included slightly differently in nontrivial cases.

I don't think the visited set is really needed (what would it be for?) but I can put it back in if it's for some reason required.

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/nesting/IncludedJarFactory.java
+++ b/src/main/java/net/fabricmc/loom/build/nesting/IncludedJarFactory.java
@@ -28,31 +28,21 @@ import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collection;
 import java.util.Locale;
-import java.util.Set;
 
-import com.google.common.collect.Sets;
 import com.google.common.hash.Hashing;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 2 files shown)


Juuxel:
> This is technically a breaking change because project dependencies may be included slightly differently in nontrivial cases.

Also a breaking change since it removes an API method (I think tasks are API classes?): `RemapJarTask.getNestedJars`.

modmuss50:
Yes, this is a major breaking change. `getNestedJars` is used quite a bit.

Niko-sk2x:
This PR was made because we thought we needed it as it broke some things for us, but we ended up doing things in a different way that makes it unnecessary and I just don't have the motivation and time to try to make it work without being a breaking change.

I'm still fairly certain this is in fact the correct way to handle include configuration by how gradle works.

lukebemish:
This is likely superseded by https://github.com/FabricMC/fabric-loom/pull/1080

PR #719, opened 2022-09-10
Fix mods on custom remap configs always being applied to the main sourceset.

modmuss50:
This does remove them from the test configs as well, im unsure if this is something we want to do, but I think it makes sense to do so. The API will fallback to the old broken state if a sourceset isnt specified.

No idea how no one (including me) failed to spot/report this 😆


modmuss50:
Let me know ASAP if this causes you issues, hopefully should be fine 👍

PR #722, opened 2022-09-11
Nightly Gradle Version Testing

isXander:
A daily cron job that tests loom against the latest nightly gradle version.

I'm not sure what gradle command to run precisely as the test push workflow looks super complicated.

### DIFF
--- a/.github/workflows/nightly-test.yml
+++ b/.github/workflows/nightly-test.yml
@@ -1,9 +1,9 @@
-name: Test latest Gradle RC
+name: Gradle Nightly Test
 on:
   schedule:
     - cron: 0 0 * * * # daily
 jobs:
-  gradle-rc:
+  nightly-test:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/nightly-test.yml
+++ b/.github/workflows/nightly-test.yml
@@ -1,0 +1,17 @@
+name: Test latest Gradle RC
+on:
+  schedule:
+    - cron: 0 0 * * * # daily
+jobs:
+  gradle-rc:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-java@v3
+      with:
+        distribution: temurin
+        java-version: 17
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-test.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-java@v3
+      with:
+        distribution: temurin
+        java-version: 17
+    - uses: gradle/gradle-build-action@v2
+      with:
+        gradle-version: nightly
+    - run: gradle test # just test build configuration
```
modmuss50, line 17: This will take many many many hours, this just isnt how the test are ran on GHA. Take a look at the other workflow, ideally this cron task could be added to the other one.

.github/workflows/nightly-test.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-java@v3
+      with:
+        distribution: temurin
+        java-version: 17
+    - uses: gradle/gradle-build-action@v2
+      with:
+        gradle-version: nightly
```
modmuss50, line 16: This wont run the intergration tests with the nightly gradle, will only try to build loom its self with it, somewhat useful to know, but not what we are going for.

modmuss50 (changes requested):

.github/workflows/nightly-test.yml
Juuxel: If you look below, it does run `gradle test` (which includes the integration tests), but without the matrix it'll take quite a long time

modmuss50: Ah yeah, I forgot, this will run the test with the nightly gradle specified here, as well as the one in the test constants, only needs to run them on the "current" gradle version.

isXander: I'll get round to this at some point, thanks

modmuss50: Sounds good 👍 Let me know if you need some pointers, Gradle/Loom can be confusing :)

PR #731, opened 2022-09-30
Add client sourceset to injected interfaces.

modmuss50:


modmuss50:
Looks like the latest 8.0 nightly has broken us again :)

PR #736, opened 2022-10-07
Patch download to work with symlinks

### LINKED ISSUE #735: Loom doesn't work with symlinks when downloading.

Phoenix-Starlight:
Resolves #735
~~Todo: Get the darned test to work~~


Phoenix-Starlight:
Okay, after some wacky test results, involving post run somehow failing it now passes.

src/test/groovy/net/fabricmc/loom/test/unit/download/DownloadFileTest.groovy
```diff
@@ -30,13 +30,30 @@ import net.fabricmc.loom.util.download.Download
...
 import java.time.Duration
 import java.time.Instant
 
 class DownloadFileTest extends DownloadTest {
+	def "Directory: Symlink"() {
+		setup:
+			server.get("/symlinkFile") {
+				it.result("Hello World")
+			}
+			Files.deleteIfExists(Paths.get(".", "faketemp"))
```
modmuss50: I would also make faketemp a sub dir in a temp dir, also assign it to a varible.

Phoenix-Starlight:
Ignoring the fail where it fails to sign: most likely some one off thing.
It's cleaner but adds a dependency, which I am not sure will break filament which was brought up. But apparently it includes transitive dependencies?.. not sure.

modmuss50 (approved):
I think this is fine 👍

PR #737, opened 2022-10-09
New Jar Processor API, Shared Service improvements, FMJ 2.0 support.

modmuss50:
Warning massive PR opening for inital feedback and to allow tests to run:

- New Jar Processor API that has better awareness of what jar is being transformed. (Allows for client/common only transforms)
  - Mapping processor api builtin (was hacked ontop before)
  - Old API is fully implimented in the new API
  - Easier to unit test
  - Peformance improvements, due to the input mods be loaded once and not for each processor
  - Builtin caching, no need for each processor to manually handle this.
- Experimental FMJ 2.0 support, may change depending on loader developments.
  - Also supports v0, and v1
  - Unit tested
  - New API to allow creating `FabricModJson` objects from a range of locations, abstracting out a lot of previously duplicated logic.
- Refactored the shared service util to be fundamentally compatible with build caching. The current impl may be slower for multi project builds, I plan to look into improving this. Still some work to be done to get build caching working.
- Fully backwards compatible with 1.0, so the plan is to release as 1.1
- Store MC jars in a maven structure, fixes source attachment in Idea

Todo:
- [ ] Replace AW with class tweakers
- [ ] Support project deps for jar processor inputs
- [ ] Document new and deprecated APIs


modmuss50:
This is not complete, but I will contiune development in the 1.1 branch as this PR is got a bit out of control size wise 👍

PR #744, opened 2022-10-23
Update Checkstyle config to match Filament

Juuxel:
See FabricMC/fabric#478 for why the language is set to English and https://github.com/FabricMC/yarn/pull/3301#discussion_r1002730499 for allowing `var`


modmuss50 (approved):
We discussed this and think it makes sense in a gradle plugin. I have no plans to do the same in projects such as Fabric API.

Will most likely put this into the upcoming 1.1 branch 👍

PR #747, opened 2022-10-29
Add multi project optimisation option to allow opting into shared tiny remapper.

modmuss50:
Im not 100% sure on the name, happy for other suggestions.

This option should be enabled in projects with many (5+) sub projects. This has been made opt-out due to fundamental issues raised. The peformance hit for most people with no sub projects will be 0, projects with a small handful of subprojects should only be a tiny bit slower.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -108,4 +108,6 @@ public interface LoomGradleExtension extends LoomGradleExtensionAPI {
 	boolean refreshDeps();
 
 	void setRefreshDeps(boolean refreshDeps);
+
+	boolean multiProjectOptimisation();
 }
--- a/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradleExtension.java
@@ -109,5 +109,11 @@ public interface LoomGradleExtension extends LoomGradleExtensionAPI {
 
 	void setRefreshDeps(boolean refreshDeps);
 
+	/**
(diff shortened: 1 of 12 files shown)


Juuxel:
> The peformance hit for most people with no sub projects will be 0

Might actually be faster if there's one less task to run (haven't checked yet)

src/main/java/net/fabricmc/loom/task/AbstractRemapJarTask.java
```diff
@@ -228,4 +229,9 @@ protected static Function<File, String> relativePath(List<String> rootPaths) {
 			return s;
 		};
 	}
+
+	@Internal
+	protected LoomGradleExtension getLoomExtension() {
```
Juuxel, line 236: This should also be `@ApiStatus.Internal`

src/main/java/net/fabricmc/loom/LoomGradleExtension.java
```diff
@@ -108,4 +108,6 @@ default List<Path> getMinecraftJars(MappingsNamespace mappingsNamespace) {
 	boolean refreshDeps();
 
 	void setRefreshDeps(boolean refreshDeps);
+
+	boolean multiProjectOptimisation();
```
Juuxel, line 118: Would benefit from a javadoc comment like this:

```
If true, multi-project optimisation mode is enabled. This mode makes builds with many Loom projects
much faster by increasing sharing and disabling some functionality.

<p>You can enable it by setting the Gradle property {@code fabric.loom.multiProjectOptimisation} to {@code true}.
```

Juuxel (changes requested):

modmuss50:
Going to merge, test failures are unrealted. 👍

PR #760, opened 2022-11-27
Clear Fernflower context after decompiling

Juuxel:
This clears up some of the resources used by Fernflower and mirrors [what is done in the CLI decompiler](https://github.com/FabricMC/intellij-fernflower/blob/49f5561beca2d4efbf94762452858f032841b31e/src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java#L119-L126).


modmuss50 (approved):
Sure, wont make a huge diffrence as it runs in a forked JVM that will be shutdown right after this.

Juuxel:
Not right after since the linemapping is also done in the forked one, but yeah. (LQF was having issues with FileSystemExceptions since QF replaces the input with NIO which requires clearing to properly close the FS, but I thought it could be good for cleanup here too.)

zml2008 (approved):

PR #770, opened 2022-12-13
Add getClientOnlySourceSetName

modmuss50:
Used to allow creating a split sources jar from the none defualt client sourcset. Use case is the fabric API testmod jar.


src/main/java/net/fabricmc/loom/task/AbstractRemapJarTask.java
```diff
@@ -103,6 +105,9 @@ public abstract class AbstractRemapJarTask extends Jar {
 	@Input
 	public abstract ListProperty<String> getAdditionalClientOnlyEntries();
 
+	@Input
+	public abstract Property<String> getClientOnlySourceSetName();
```
Juuxel, line 111: Fyi this could also also be an `@Internal Property<SourceSet>` if you added a custom input property on `getClientOnlySourceSet().get().getName()`, but probably overkill

Juuxel (approved):

PR #772, opened 2022-12-14
Greatly reduced memory usage by excluding unnecessary classpaths from remap task

Luna5ama:
Using `JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME`/`complileClasspath` causes all the compile time dependencies such as user-defined external libraries and Minecraft libraries to be included in the remapping classpath. Those could be a huge amount of jars and classes, which could consume a lot of memory when they get read by TinyRemapper. Especially in projects with multiple modules, it could force a full GC and slow down the build process.
Only the `modCompileClasspath` (libraries to be remapped) and `minecraftNamed` (remapped mc jar for development environment) are needed for remapping as they are the only ones relevant to mapping.

### Result
Before:
before
before time

After:
after
after time

Note:
I know I was using a fork of fabric loom but the change should still be relevant.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
+++ b/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
@@ -88,7 +88,14 @@ public abstract class RemapJarTask extends AbstractRemapJarTask {
 	public RemapJarTask() {
 		super();
 
-		getClasspath().from(getProject().getConfigurations().getByName(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME));
+		// Lazy provider because the configuration is not yet registered at this point
+		getClasspath().from(getProject().provider(() -> {
+			String sourceSetName = MinecraftSourceSets.get(getProject()).getCombinedSourceSetName();
+			// Use copy to avoid inherited configuration (Minecraft libraries)
+			return getProject().getConfigurations().getByName(sourceSetName).copy();
+		}));
+		getClasspath().from(getProject().getConfigurations().getByName("modCompileClasspath"));
+
 		getAddNestedDependencies().convention(true).finalizeValueOnRead();
 
 		Configuration includeConfiguration = getProject().getConfigurations().getByName(Constants.Configurations.INCLUDE);


modmuss50:
> all the compile time dependencies such as user-defined external libraries and Minecraft libraries to be included in the remapping classpath

These most definitely should be on the remap classpath. I am aware of this issue but believe it to a bug/lack of optimistation in tiny remapper. 

My memory profiling from a little whileback resulted in this (Taken from Fabric API with 3 copies of TR, now only 2):
image

I believe this issue started when TR got MRJ (multi release jar) support. Possibly around this TODO: https://github.com/FabricMC/tiny-remapper/blob/master/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java#L1010

I have already spoken @sfPlayer1 about this a while back and they are going to be taking a closer look at it. While this PR may appear to work on the surface I believe it to be fundamentally wrong and will break down with some edge cases that are not covered by the automated tests.

Lets aim to fix the root cause of the issues and not have hacky work arounds like this. 👍

Luna5ama:
> > all the compile time dependencies such as user-defined external libraries and Minecraft libraries to be included in the remapping classpath
> 
> These most definitely should be on the remap classpath. I am aware of this issue but believe it to a bug/lack of optimistation in tiny remapper.

I am not sure how are they needed to be in the remap classpath. From what I understand the Minecraft jar has to be included is because the mapping names have to be propagated, from the Minecraft superclasses to their subclasses, including the ones in the mod, to ensure references to members in subclasses get remapped. Minecraft libraries aren't dependent on Minecraft classes anyways therefore they should be fine.

Same with user-defined dependencies, if they are using any Minecraft-dependent dependencies such as another mod, they will use the `modImplementation`, `modApi` configurations anyways, The change already includes the `modCompileClasspath` to remap classpath so it should be handled.


> I believe this issue started when TR got MRJ (multi release jar) support. Possibly around this TODO: https://github.com/FabricMC/tiny-remapper/blob/master/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java#L1010
> 
> I have already spoken @sfPlayer1 about this a while back and they are going to be taking a closer look at it. While this PR may appear to work on the surface I believe it to be fundamentally wrong and will break down with some edge cases that are not covered by the automated tests.
> 
> Lets aim to fix the root cause of the issues and not have hacky work arounds like this. 👍

That is another main cause for the ram usage for sure. I also noticed it in my heap dump. This is a quick hacky fix for sure but this will also help reduce the memory usage by a good amount besides a proper fix on MRJ.  And surely more examinations are needed to ensure it works properly in different scenarios. 

Would be great to see any of those fixes being deployed so multi-module builds could be run in parallel without running out of memory.

modmuss50:
> I am not sure how are they needed to be in the remap classpath. From what I understand the Minecraft jar has to be included is because the mapping names have to be propagated, from the Minecraft superclasses to their subclasses, including the ones in the mod, to ensure references to members in subclasses get remapped. Minecraft libraries aren't dependent on Minecraft classes anyways therefore they should be fine.

Yeah, thats correct, but we need to make sure we have a fully populated classpath to get all edge cases. The namespace of the classpath needs to match the source namespace when remapping. `modCompileClasspath` contains all of the intermediary named dependcies when `named` is required. However fixing this still would not solve all of the edge cases, dependecies that use Minecraft classes can still be one one of the regular configurations. An example of this may be a sub project.

> That is another main cause for the ram usage for sure. I also noticed it in my heap dump.

Did you see anything else? The MJR states is the only thing I saw when profiling this. I am a little wary about using your graphs as a comparision between the two versions, even in your "fixed" screenshot the heap size still seems to be close to 3GB? (In my testing a while back I think a Fabric API build was only a little over 1GB)

I would much prefer to see solid numbers from a heap dump. Especially not from a forked version of loom that may or may not contain improvements I have made.

Loom 1.1 (`exp/1.1` branch) does make some improvements, a good place to take a heapdump is: https://github.com/FabricMC/fabric-loom/blob/exp/1.1/src/main/java/net/fabricmc/loom/util/service/SharedServiceManager.java#L70 just before it cleans everything up. Make sure to enable `fabric.loom.multiProjectOptimisation` if you are using 1.1.

Luna5ama:
> > I am not sure how are they needed to be in the remap classpath. From what I understand the Minecraft jar has to be included is because the mapping names have to be propagated, from the Minecraft superclasses to their subclasses, including the ones in the mod, to ensure references to members in subclasses get remapped. Minecraft libraries aren't dependent on Minecraft classes anyways therefore they should be fine.
> 
> Yeah, thats correct, but we need to make sure we have a fully populated classpath to get all edge cases. The namespace of the classpath needs to match the source namespace when remapping. `modCompileClasspath` contains all of the intermediary named dependcies when `named` is required. However fixing this still would not solve all of the edge cases, dependecies that use Minecraft classes can still be one one of the regular configurations. An example of this may be a sub project.

That is a good point, I could also make it include `namedElements` dependencies.


> Did you see anything else? The MJR states is the only thing I saw when profiling this. I am a little wary about using your graphs as a comparision between the two versions, even in your "fixed" screenshot the heap size still seems to be close to 3GB? (In my testing a while back I think a Fabric API build was only a little over 1GB)

Yes it was the MRJ state, there are two or more of them in each remapper instance. They both were problematic because multiple MRJ states holding copies of the extra `ClassInstance` from the whole `compileClasspath`. My Gradle Daemon heap usage was high because the project I was working on has 6 subprojects running `remapJar`, 1.16.5, 1.18.2, 1.19.2, both fabric and forge. In your heap dump screenshot shows that each TR was 500MB, so the 3GB usage should make sense.


> Loom 1.1 (`exp/1.1` branch) does make some improvements, a good place to take a heapdump is: https://github.com/FabricMC/fabric-loom/blob/exp/1.1/src/main/java/net/fabricmc/loom/util/service/SharedServiceManager.java#L70 just before it cleans everything up. Make sure to enable `fabric.loom.multiProjectOptimisation` if you are using 1.1.

I will try that and see if it does improve anything.

Luna5ama:
> > Loom 1.1 (`exp/1.1` branch) does make some improvements, a good place to take a heapdump is: https://github.com/FabricMC/fabric-loom/blob/exp/1.1/src/main/java/net/fabricmc/loom/util/service/SharedServiceManager.java#L70 just before it cleans everything up. Make sure to enable `fabric.loom.multiProjectOptimisation` if you are using 1.1.
> 
> I will try that and see if it does improve anything.

Tested with a built on the 1.1 branch and project with 6 different subproject (just to replicate the original scene) that contains nothing but an example `ModInitializer`. Unfortunately, either `fabric.loom.multiProjectOptimisation` enabled or disabled seem to improve the memory usage.

Enabled:
usage_enabled

Disabled:
usage_disabled

And for the heap dump, I realized there isn't a way to capture a perfect one because the tasks finish asynchronously. But this is what I got by trying to capture it at max usage:

heapdump_enabled_a
heapdump_enabled_b

modmuss50:
Superseeded by: #987 I have been a bit more careful to only remove the minecraft libs.

PR #774, opened 2022-12-16
Fix Javadoc on injected inner interfaces

jpenilla:


jpenilla:
Might actually need to trim the up to the last `$` (or maybe just use fully qualified name when it contains `$`), the decompile can contain imports like `import com.example.Inner.Class` instead of importing `import com.example.Inner` and referencing `Inner.Class`. Will look into this more later

modmuss50:
This was fixed here: https://github.com/FabricMC/fabric-loom/pull/900

Thanks for the PR

PR #791, opened 2023-01-06
Fix #790

### LINKED ISSUE #790: Loom build with kapt caused `Multiple output file properties with name 'mixin-ap-main'`
Repo: https://github.com/SettingDust/fabric-kapt-issue

When gradle `compileJava` with kapt it will 👇
```
* What went wrong:
Execution failed for task ':compileJava'.
> Multiple output file properties with name 'mixin-ap-main'
```

Caused by Java and Kapt add property twice
https://github.com/FabricMC/fabric-loom/blob/dev/1.0/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java#L262-L277
```
passMixinArguments:98, AnnotationProcessorInvoker (net.fabricmc.loom.build.mixin)
configureMixin:147, AnnotationProcessorInvoker (net.fabricmc.loom.build.mixin)
configureMixin:76, KaptApInvoker (net.fabricmc.loom.build.mixin)
setupMixinAp:271, CompileConfiguration (net.fabricmc.loom.configuration)
```
```
passMixinArguments:98, AnnotationProcessorInvoker (net.fabricmc.loom.build.mixin)
configureMixin:147, AnnotationProcessorInvoker (net.fabricmc.loom.build.mixin)
setupMixinAp:262, CompileConfiguration (net.fabricmc.loom.configuration)
```

modmuss50:
Tried to add kapt to an intergration test but it fails for some weird reason. Manually tested and it seems to work fine.

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/mixin/AnnotationProcessorInvoker.java
+++ b/src/main/java/net/fabricmc/loom/build/mixin/AnnotationProcessorInvoker.java
@@ -64,15 +64,17 @@ public abstract class AnnotationProcessorInvoker<T extends Task> {
 	protected final Project project;
 	protected final MixinExtension mixinExtension;
 	protected final Map<SourceSet, T> invokerTasks;
+	private final String name;
 	private final Collection<Configuration> apConfigurations;
 
 	protected AnnotationProcessorInvoker(Project project,
-										Collection<Configuration> apConfigurations,
-										Map<SourceSet, T> invokerTasks) {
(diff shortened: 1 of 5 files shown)


SettingDust (approved):
Added different names for different invoker. Should work

modmuss50:
I think this is a fine small fix for 1.0, I do think that we may also need to change the mixin mapping file name to include this name to prevent conflicts there. But I will make that change in the exp/1.1 branch.

PR #792, opened 2023-01-07
Add ability to mark client config as authenticated using DevLogin

ApexModder:
This PR brings the ability to mark client run configs as being authenticated. These configs use Covers1624 DevLogin tool to login & authenticate against Mojangs account servers on launch.

See [here](https://github.com/covers1624/DevLogin#devlogin) for more details on DevLogin.

---

I have added Covers maven repo to be able to pull in the DevLogin jar, and have needed to change the main class for the client run config generated to `net.covers1624.devlogin.DevLogin` aswell as append a new program arg `--launch_target <original main>`.
[_Following the [wiki](https://github.com/covers1624/DevLogin/wiki/ForgeGradle-Setup) as a setup guide_]

---

To mark a config as being authenticated the user must specify `authenticated` in their gradle file like so
```groovy
loom {
	runs {
		client {
			// ...
			authenticated()
			// ...
		}
	}
}
```

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
@@ -37,6 +37,7 @@ import org.gradle.api.plugins.ExtensionAware;
 import org.gradle.api.plugins.PluginAware;
 import org.jetbrains.annotations.NotNull;
 
+import net.fabricmc.loom.configuration.ide.RunConfigSettings;
 import net.fabricmc.loom.extension.LoomFiles;
 import net.fabricmc.loom.util.MirrorUtil;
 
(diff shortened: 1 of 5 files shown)


modmuss50:
Hi, thanks for the PR. Sadly I don’t think this is something we want especially in this form.

Using a 3rd party maven repo like this is far from ideal, if/when that maven repo goes down all builds will break.

Handing authentication is something we have decided not to participate in, it’s hard for us to ensure that a users auth data is kept secure, I especially do not want to depend on 3rd party code that would be hard for us to trust enough to bare the responsibility if there was an issue.

One idea I have had for a while would be to allow loom to install its run configs into the vanilla launcher.

ApexModder:
> Using a 3rd party maven repo like this is far from ideal, if/when that maven repo goes down all builds will break.

While I understand what you are saying, relying on third parties can easily break builds at random unforeseen times for seemingly no reason.

I have written it so that it only adds the third party maven & includes the dependency when a user marks a runConfig as authenticated, no authed configs, the repo nor the dependency will be added.

I have also been speaking with @covers1624 on Discord and he made mention that if these PR's [_one also exists for [ForgeGradle](https://github.com/MinecraftForge/ForgeGradle/pull/896)_] were to go through, he would potentially move/mirror DevLogin onto the main Forge maven rather than his own maven, but obviously, couldn't open a none functioning PR, so I used his maven for the meantime to get the PR functioning.

> Handing authentication is something we have decided not to participate in, it’s hard for us to ensure that a users auth data is kept secure, I especially do not want to depend on 3rd party code that would be hard for us to trust enough to bare the responsibility if there was an issue.

The implementation of DevLogin from my usages and reading of the code are pretty safe & secure, it all runs locally on your machine and only network call outs it does is to official Mojang & Microsoft services to authenticate you.
After it successfully logs in, the data is then cached locally on your machine in `~/.devlogin/` and called back every time you launch the game, using expiring tokens.
If the tokens have expired, using the cached data it attempts to auto-refresh the token and re-authenticate you.

> One idea I have had for a while would be to allow loom to install its run configs into the vanilla launcher.

This does intrigue me though, I'm assuming if this was to happen would function like CurseForge launcher, hit play and the Vanilla launcher opens with only 1 profile, pre-set up for development in this case.
While I like the idea of using Vanilla launcher, I wouldn't want to essentially have to hit _play_ twice in order to launch the game, especially with how many times and frequently you can launch the game during development.
Also how would debugging work this way, I assume debugger could still be attached when launched from Vanilla launcher.

modmuss50:
I think this PR would be best implemented as a separate plug-in. Loom should already offer all of the API needed, if not I’d be happy to help make this work.

These 3rd party auth methods use a reverse engineered API with no official documentation. If Mojang were to provide an official library and/or documentation I would be much more happy to add something like this.

I could be wrong but I think most people only want to be authenticated to test on a real server once they have finished a feature that was developed using single player or a local offline server.


I think we would still be able to debug using the vanilla launcher. See one of my open PRs adding prod run tasks, this could also help out there.

PR #793, opened 2023-01-07
Use soft reference for shared service caching

Luna5ama:
This would allow JVM garbage collector clear inactive shared services on demand, eg. when heap memory runs out. Making it possible to build large projects with different mc versions (They can't share remapper service). It should achieve similar effects to #772 without breaking functionality.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
+++ b/src/main/java/net/fabricmc/loom/task/RemapJarTask.java
@@ -122,7 +122,10 @@ public abstract class RemapJarTask extends AbstractRemapJarTask {
 				params.getNestedJars().from(getNestedJars());
 			}
 
-			params.getTinyRemapperBuildServiceUuid().set(UnsafeWorkQueueHelper.create(getProject(), tinyRemapperService.get()));
+			Supplier<TinyRemapperService> temp = tinyRemapperService;
+			tinyRemapperService = null; // Release the strong reference
+			params.getTinyRemapperBuildServiceUuid().set(UnsafeWorkQueueHelper.create(getProject(), temp.get()));
+
 			params.getRemapClasspath().from(getClasspath());
 
 			final boolean legacyMixin = extension.getMixin().getUseLegacyMixinAp().get();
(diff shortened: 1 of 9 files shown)


modmuss50:
I don’t think this is quite right as it may no longer share the services between different projects when it makes sense to, this vastly slowing the build.

The 1.1 branch by default creates and closes a service manger per remap task when multi project optimisations are not enabled.

It’s possible this may not be working quite right as I haven’t yet fully tested it. Creating an integration test that has many mc versions may be a good idea?

Luna5ama:
> I don’t think this is quite right as it may no longer share the services between different projects when it makes sense to, this vastly slowing the build.

Each `PrepareRemapTask` calls `RemapJarTask.getTinyRemapperService()` from its corresponding remap task. It triggers the memorized supplier, runs `TinyRemapperService.getOrCreate(this)`, and stores the reference to the remapper service. It retains this strong reference until the remap task starts running and passes it to the `RemapAction`. So it won't be cleared accidentally before the remap task kicks in, and it will also not be cleared until the task finish and release the reference.
If multiple remap tasks are sharing the remapper service, they all will retain the reference to the remapper service during the prepare remap task. Therefore they will still be able to share the remapper service.

> It’s possible this may not be working quite right as I haven’t yet fully tested it. Creating an integration test that has many mc versions may be a good idea?

Sure that will be great, since the current tests don't have coverage for that.

modmuss50:
I have spent a few hours profiling and optimising loom here in this PR here: https://github.com/FabricMC/fabric-loom/pull/796 this allows me to build FabricAPI with `-Xmx2G'.

Loom 1.1 uses a ref counting [BuildSharedServiceManager](https://github.com/FabricMC/fabric-loom/blob/exp/1.1/src/main/java/net/fabricmc/loom/util/service/BuildSharedServiceManager.java) for the remap jar task to close the TinyRemapper instance once its finished.

I think I will cherry pick your test into my branch to help test this.

modmuss50:
All of the shared service stuff has been removed in 1.8, this should no longer be an issue.

PR #796, opened 2023-01-10
Gradle 8 tests, perf optimisations

modmuss50:
Add gradle 8 to the test matrix
Reuse gradle home between tests, should hopefully speed them up a lot.
Misc perf and mem optimisations


modmuss50:
Hey @Luna5ama I have commited your test to this branch, it seems to pass for me locally. Can you verify this?

I think there are still some improvements to be made, we can possibly use [BuildServiceSpec.getMaxParallelUsages()](https://docs.gradle.org/current/javadoc/org/gradle/api/services/BuildServiceSpec.html#getMaxParallelUsages--) to create a lock that by default (without multi project optimistations) forces the remap jar/sources tasks to not run asynchronously.

I have a few other ideas that I might explore as well, such as nested SharedServiceManager's to help ensure that no services are kept around longer than needed.

modmuss50:
Going to merge this, its not perfect but better than nothing 👍 I will contiune to look at this in follow up PRs

PR #799, opened 2023-01-11
Add RemapConfigurationSettings.getApplyDependencyTransforms to close #797

### LINKED ISSUE #797: Interface injection chooses an overly broad dependency set
I have a project where my main source set has a limited subset of fabric api modules on the compile classpath, and a separate testmod source set has all of fabric API available at compile- and runtime.

Currently interface injection sums all remapped configurations together, which injects unresolvable classes onto the compile classpath for some configuration, producing errors like:

```
C:\Users\abby\dev\kyori-text\adventure-platform-fabric\src\main\java\net\kyori\adventure\platform\fabric\impl\server\ServerPlayerAudience.java:236: error: cannot access FabricItem
    final ItemStack bookStack = new ItemStack(Items.WRITTEN_BOOK, 1);
                                ^
  class file for net.fabricmc.fabric.api.item.v1.FabricItem not found
C:\Users\abby\dev\kyori-text\adventure-platform-fabric\src\main\java\net\kyori\adventure\platform\fabric\impl\server\ServerPlayerAudience.java:237: error: cannot access FabricItemStack
    final CompoundTag bookTag = bookStack.getOrCreateTag();
                                         ^
  class file for net.fabricmc.fabric.api.item.v1.FabricItemStack not found
```

Is there a sensible fully automated default here? My feeling is that this may require adding additional configurability to the interface injection extension, maybe taking a list of remapped configurations to use for injection?

I've [temporarily worked around this issue](https://github.com/KyoriPowered/adventure-platform-fabric/blob/mc/1.19/build.gradle.kts#L143) by disabling dependency interface injection entirely.

modmuss50:
Closes #797

By default this will no longer apply transitive access widners or injected interfaces found in mods from remapping configurations other than `main` or `client`. An API is provided to allow opting back into this behaviour.

This is technically a breaking change, but I dont think many if anyone depended on this odd behaviour.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/RemapConfigurationSettings.java
+++ b/src/main/java/net/fabricmc/loom/api/RemapConfigurationSettings.java
@@ -60,6 +60,7 @@ public abstract class RemapConfigurationSettings implements Named {
 		getOnCompileClasspath().finalizeValueOnRead();
 		getOnRuntimeClasspath().finalizeValueOnRead();
 		getPublishingMode().convention(PublishingMode.NONE).finalizeValueOnRead();
+		getApplyDependencyTransforms().convention(defaultDependencyTransforms()).finalizeValueOnRead();
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


Juuxel:
I'd consider the old behaviour a bug anyway - it'll break the main source set's runtime if unknown itfs are injected.

zml2008:
This looks reasonable to me, thanks!!

PR #806, opened 2023-01-15
Restructure mod configuration remapping to better align with how Gradle works

### LINKED ISSUE #801: Weird behavior with compileOnlyApi dependencies of mod dependencies
I have a mod library ([`cloud-fabric`](https://github.com/Incendo/cloud)) that declares a `compileOnlyApi` dependency on `org.apiguardian:apiguardian-api:1.1.2` (not a mod). When adding `cloud-fabric` to `modImplementation` in a separate project, it seems like the `compileOnlyApi` dependency isn't properly pulled in, as evidenced by the output of build scans/dependencies task, and the massive amount of warnings on build: https://github.com/jpenilla/squaremap/actions/runs/3762624455/jobs/6395483461#step:7:131
(adding the dependency to `compileOnly` manually is a workaround, which I've since used in the linked repo)

I also have non-mod libraries (for example `cloud-paper`), which declares the same `compileOnlyApi` dependency, and when depending on it using `implementation` in a separate project, I get no similar warnings and the dependency appears in scans/dependencies as expected, which is what leads me to think it's something with loom.

I've spent some time trying to debug this but haven't really gotten anywhere.

`cloud.commandframework:cloud-fabric:1.8.0` is published on maven central, adding it to `modImplementation` is enough to reproduce the issue, to get the warnings just reference an annotated method (for example `StringArgument.builder("");`) (might also need `-Xlint:all`, not sure as I always have it on)

*Every* module in cloud has this `compileOnlyApi` dep, which means not only is the direct dependency lost, but also the dependencies of for example `cloud-core` that `cloud-fabric` also depends on.

jpenilla:
Resolves #801

Basic testing has been done with standard & split source projects, but this is still pretty rough and likely has issues I haven't noticed yet. Tests fail as they aren't updated.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -25,7 +25,6 @@
 package net.fabricmc.loom.api;
 
 import java.io.File;
-import java.util.List;
 
 import org.gradle.api.Action;
 import org.gradle.api.NamedDomainObjectContainer;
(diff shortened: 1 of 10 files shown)


modmuss50:
Please do not target large changes like this to the 1.0 branch.

Juuxel (changes requested):
Breaking changes in minor updates :ohno:

semver go brr

jpenilla:
Long term, restructuring this in a way that breaks API to align better with Gradle is probably desirable as it fixes some edge case issues and removes the cruft needed to keep the old API intact, but the linked issue should be resolved by #807. When/if the aforementioned restructuring is wanted, this PR won't be that useful anymore as it will need to be rebased on #807, so I am closing it.

PR #807, opened 2023-01-16
Rewrite the internals of mod configuration remapping to fix bugs

### LINKED ISSUE #572: Depending on a mod with transitive non-mods deps exposes those deps in pom.xml
For example, if a mod depends on Fabric Language Kotlin with `modImplementation`, its POM will contain
- A `runtime` dep on FLK as expected
- **Extra `runtime` deps on...**
  - kotlin stdlib modules
  - kotlinx-serialization
  - kotlin-reflect

when set up normally with software components.

This happens because we add the non-mod artifacts from `mod*` configs into the standard configs - like `implementation` - directly. Not sure what a good fix would be, though.

### LINKED ISSUE #801: Weird behavior with compileOnlyApi dependencies of mod dependencies

Juuxel:
Fixes #801. Fixes #572.

Needs a bit of testing, esp. with split builds and split dependencies.

- Loom now remaps mods to "collector configs" that are specific to the execution type (compile-time/API vs runtime).
  - For example, `modImplementation` remaps to `modCompileClasspathMain` and `modRuntimeClasspathMain`.
  - This lets us use Gradle's `org.gradle.usage` attributes to select whether it's a compile-time or runtime dependency
  - Note: Remap configurations sourcing from `api` are partially left intact due to their special logic in the `namedElements` configuration.
- Setting the proper usages fixes #801.
- No dependency files are added to the real target configurations (like `api` and `implementation`) anymore.
  - This fixes #572, since `runtimeClasspath` and `compileClasspath` don't leak transitive dependencies unlike `implementation` etc.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/Strings.java
+++ b/src/main/java/net/fabricmc/loom/util/Strings.java
@@ -1,0 +1,47 @@
+/*
+ * This file is part of fabric-loom, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2023 FabricMC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.util;
+
+public final class Strings {
+	public static String capitalize(String word) {
+		// Don't capitalize empty strings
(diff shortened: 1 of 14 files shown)


src/main/java/net/fabricmc/loom/util/Strings.java
```diff
@@ -0,0 +1,47 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.util;
+
+public final class Strings {
+	public static String capitalize(String word) {
```
Juuxel, line 28: I couldn't find this in any libraries and didn't want to pull in another one, let me know if there is a library that we already have that has it

modmuss50: I think this is fine, I dont see any existing library usage. You could have possibly used SourceSet.getTaskName I think but its clearly not really intended for this usecase.

Juuxel: I used it originally but I need the `main` source set name in this case, otherwise it'd conflict with the generic `modCompileClasspath` used for all compile cp mods.

modmuss50 (approved):
Looks great, tests are also useful 👍

PR #818, opened 2023-01-24
Update MCE

### LINKED ISSUE #813: Loom 1.1.2 on Windows does not release file locks at end of build
This appears to be a new regression in 1.1. Per the CI log from [adventure-platform-fabric](https://github.com/KyoriPowered/adventure-platform-fabric/actions/runs/3983153992/jobs/6828230484), it seems like a run of loom with an actively running daemon (aka not the first run, but subsequent runs) will always fail due to a held file lock.

I was able to reproduce this locally, though on my first try for some reason Loom found an existing cache lock. Honestly I'm not sure the best way to track down where the dangling file handle is being held -- my first guess would be some sort of JMX bean, but I didn't find anything there. I ended up taking a heap dump of the Gradle daemon JVM, and while I didn't see anything conclusive (and really, I'm not sure if a heap dump is the most effective way to do this investigation), I did see that some of the objects associated with the Mercury run were being retained -- something with shared services maybe?

modmuss50:
MCE change: https://github.com/FabricMC/fabric-mixin-compile-extensions/pull/20
Fixes #813

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/mixin/AnnotationProcessorInvoker.java
+++ b/src/main/java/net/fabricmc/loom/build/mixin/AnnotationProcessorInvoker.java
@@ -141,8 +141,7 @@ public abstract class AnnotationProcessorInvoker<T extends Task> {
 				project.getLogger().info("Adding mixin to classpath of AP config: " + processorConfig.getName());
 				// Pass named MC classpath to mixin AP classpath
 				processorConfig.extendsFrom(
-						configs.getByName(minecraftSourceSets.getCombinedSourceSetName()),
-						configs.getByName(Constants.Configurations.MOD_COMPILE_CLASSPATH_MAPPED),
+						configs.getByName(Constants.Configurations.LOADER_DEPENDENCIES),
 						configs.getByName(Constants.Configurations.MAPPINGS_FINAL)
 				);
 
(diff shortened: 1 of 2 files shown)


modmuss50:
All tests passed, other than the unit test that I broke as part of another PR. I will merge this please let me know of any issues. Thanks again zml 👍

src/main/java/net/fabricmc/loom/build/mixin/AnnotationProcessorInvoker.java
```diff
@@ -141,8 +141,7 @@ public void configureMixin() {
 				project.getLogger().info("Adding mixin to classpath of AP config: " + processorConfig.getName());
 				// Pass named MC classpath to mixin AP classpath
 				processorConfig.extendsFrom(
-						configs.getByName(minecraftSourceSets.getCombinedSourceSetName()),
-						configs.getByName(Constants.Configurations.MOD_COMPILE_CLASSPATH_MAPPED),
+						configs.getByName(Constants.Configurations.LOADER_DEPENDENCIES),
 						configs.getByName(Constants.Configurations.MAPPINGS_FINAL)
```
zml2008, line 145: I don't think the `mappingsFinal` configuration is needed here? It seems like the mappings file used is passed below as a normal path, and the Mixin AP graps it from that file rather than a classpath resource.

Admittedly I don't think it'd cause too much harm, but if the AP class loader holds on to the file handle it might prevent changes to locally generated mappings.

PR #835, opened 2023-02-24
Generate a fabric-loader-junit config file on the classpath.

modmuss50:
Very draft ATM, just trying to get something working for 1.1. I expect in 1.2 ill rework this and provide a DSL to allow setting these options, alongside DLI options.

Requires: https://github.com/FabricMC/fabric-loader/pull/776


modmuss50:
Not needed.

PR #838, opened 2023-02-28
Change how `include` disables transitive dependencies to allow platform dependencies to work

jpenilla:


Juuxel (approved):

jpenilla:
Looking at this closer, I'm not sure it's needed to disable transitivity at all due to how includes are resolved via top-level dependencies. Removing the transitive disable logic altogether would also allow transitive platform dependencies to work.

modmuss50:
This could do with a test tbh. Make sure that A) it still includes depdencies correctly and B) does not include transitive deps. I think this is fine for now though.

Sorry it seems I tottaly missed this PR...

modmuss50 (approved):

PR #840, opened 2023-03-06
fix: remapping of jars without a fabric.mod.json

marcus8448:
Jars that have `Fabric-Loom-Remap: true` without a `fabric.mod.json` fail to remap, as loom searches for accesswideners in the FMJ (which doesn't exist in the jar).

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/mods/AccessWidenerUtils.java
+++ b/src/main/java/net/fabricmc/loom/configuration/mods/AccessWidenerUtils.java
@@ -60,6 +60,7 @@ public class AccessWidenerUtils {
 		if (!FabricModJsonFactory.isModJar(inputJar)) {
 			return null;
 		}
+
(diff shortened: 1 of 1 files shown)


src/main/java/net/fabricmc/loom/configuration/mods/AccessWidenerUtils.java
```diff
@@ -57,6 +57,9 @@ public static byte[] remapAccessWidener(byte[] input, Remapper remapper) {
 	}
 
 	public static AccessWidenerData readAccessWidenerData(Path inputJar) throws IOException {
+		if (!FabricModJsonFactory.isModJar(inputJar)) {
+			return null;
+		}
```
modmuss50, line 62: Missing a new line, this wont pass checkstyle.

marcus8448:
Could someone rerun the GH actions? It seems that the tests failed due to SSL errors.

modmuss50 (approved):

PR #848, opened 2023-03-29
Support intermediary metadata

modmuss50:
At a high level this basically copies the manifest from intermediary to the generated mappings.jar file, this will be used to allow intermediary to drive loaders generated Minecraft version.

In the future this might be expandable to allow mapping layers to add their own metadata, it may also be useful for older versions that do not use the default intermediary provider.

This is a well tested none breaking change for loom 1.1.

Merge pending intermediary and loader changes.

### DIFF
--- a/src/test/groovy/net/fabricmc/loom/test/util/ZipTestUtils.groovy
+++ b/src/test/groovy/net/fabricmc/loom/test/util/ZipTestUtils.groovy
@@ -49,16 +49,16 @@ class ZipTestUtils {
         return file
     }
 
-	static String manifest(String key, String value) {
-		return manifest(Map.of(key, value))
-	}
+    static String manifest(String key, String value) {
+        return manifest(Map.of(key, value))
+    }
 
     static String manifest(Map<String, String> entries) {
         def manifest = new Manifest()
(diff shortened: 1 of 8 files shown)


src/test/groovy/net/fabricmc/loom/test/util/ZipTestUtils.groovy
```diff
@@ -49,10 +49,16 @@ class ZipTestUtils {
         return file
     }
 
-    static String manifest(String key, String value) {
+	static String manifest(String key, String value) {
+		return manifest(Map.of(key, value))
+	}
+
+    static String manifest(Map<String, String> entries) {
```
Juuxel: Mixed spaces and tabs 😍

We should probably figure something out for the inconsistent indentation in Groovy files, some files have spaces and others have tabs.

modmuss50: Yeah, fixed this here. I will resolve this for all files in another PR later.

modmuss50: See: https://github.com/FabricMC/fabric-loom/pull/850

modmuss50:
I decided against doing this for now.

PR #849, opened 2023-03-29
Update Dependencies and move to Gradle 8.1 (Kotlin 1.8.10)

JT122406:


modmuss50:
Hi, thanks for the PR. For the moment updating Gradle like this is not something I want to do just yet. My plan was to move to Gradle 8.1 with the next loom version as there are some neat features we can make use of. This will also require bumping the min required Gradle version in the bootstrap.

Ill leave this open to come back to it when I start work on the next version of loom.

JT122406:
sounds good!

build.gradle
```diff
@@ -95,26 +95,26 @@ dependencies {
 	implementation ('net.fabricmc:mercury:0.3.0')
 
 	// Kotlin
-	implementation('org.jetbrains.kotlinx:kotlinx-metadata-jvm:0.5.0') {
+	implementation('org.jetbrains.kotlinx:kotlinx-metadata-jvm:0.6.0') {
```
modmuss50, line 98: Was this tested, I have just updated to 0.6.0 and it seems to have broken and deprecated the old API we were using?

JT122406: We can revert it but It shouldn't have broken anything

modmuss50:
Many thanks for the PR, I have intergrated it into my PR here: https://github.com/FabricMC/fabric-loom/pull/866

The required Kotlin changes are very siginfiact, so it was just easier to do it in my PR. 👍

JT122406:
Sounds good @modmuss50, Thank you for your time and your work here!

Issue #461, opened 2021-07-30
Loom doesn't set configuration roles


Juuxel:
From the [Gradle docs](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs):

> In short, a configuration’s role is determined by the `canBeResolved` and `canBeConsumed` flag combinations: [table not included]
> For backwards compatibility, both flags have a default value of true, but as a plugin author, you should always determine the right values for those flags, or you might accidentally introduce resolution errors.

This is very easy to add but is a bit tedious.

PR #857, opened 2023-04-11
Rewrite Minecraft Library handling.

modmuss50:
WIP, I plan this will be part of Loom 1.2

This PR rewrites the Minecraft library processing with a more structured and testable set of "library processors". The current code is a mess of special cases and work arounds for various issues on various platforms. 

Previously this was only really used on lesser used platforms/versions so wasnt a major issue if things broke, however current shipping Minecraft versions (1.19.4) use an LWJGL version that does not work well on newer Java versions. This will mean that we will need to upgrade more LWJGL versions than we did previously.


src/main/java/net/fabricmc/loom/util/Platform.java
```diff
@@ -24,14 +24,36 @@
 
 package net.fabricmc.loom.util;
 
-public record Architecture(String name) {
-	public static final Architecture CURRENT = new Architecture(System.getProperty("os.arch"));
+public interface Platform {
```
Juuxel, line 27: What's the point of making this an interface with a single impl? Mocking?

modmuss50: Yes, Mocking. Previously it was impossible to test any of this as it changed based upon the platform its running on. I will be able to write tests for the hardware that I dont use regularly.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -68,60 +69,87 @@
...
-		project.getDependencies().add(Constants.Configurations.LOOM_DEVELOPMENT_DEPENDENCIES, Constants.Dependencies.TERMINAL_CONSOLE_APPENDER + Constants.Dependencies.Versions.TERMINAL_CONSOLE_APPENDER);
+		project.getDependencies().add(Configurations.LOOM_DEVELOPMENT_DEPENDENCIES, Constants.Dependencies.DEV_LAUNCH_INJECTOR + Constants.Dependencies.Versions.DEV_LAUNCH_INJECTOR);
+		project.getDependencies().add(Configurations.LOOM_DEVELOPMENT_DEPENDENCIES, Constants.Dependencies.TERMINAL_CONSOLE_APPENDER + Constants.Dependencies.Versions.TERMINAL_CONSOLE_APPENDER);
 		project.getDependencies().add(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
 		project.getDependencies().add(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
 	}
 
-	public static void configureCompile(Project project) {
+	private NamedDomainObjectProvider<Configuration> createNoneTransitive(String name) {
```
Juuxel: `createNonTransitive` I think

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -68,60 +69,87 @@
...
+		return configurations.register(name, configuration -> configuration.setTransitive(false));
+	}
+
+	private NamedDomainObjectProvider<Configuration> createTransitive(String name) {
+		return configurations.register(name, configuration -> configuration.setTransitive(true));
+	}
```
Juuxel: Could be a nice opportunity to resolve #461

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -68,60 +69,87 @@
...
 		project.getDependencies().add(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME, Constants.Dependencies.JETBRAINS_ANNOTATIONS + Constants.Dependencies.Versions.JETBRAINS_ANNOTATIONS);
 	}
 
-	public static void configureCompile(Project project) {
+	private NamedDomainObjectProvider<Configuration> createNoneTransitive(String name) {
+		return configurations.register(name, configuration -> configuration.setTransitive(false));
+	}
+
+	private NamedDomainObjectProvider<Configuration> createTransitive(String name) {
+		return configurations.register(name, configuration -> configuration.setTransitive(true));
```
Juuxel: Isn't `transitive == true` the default?

Juuxel (changes requested):

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
modmuss50: Yeah, I'll remove this as its only used in one place and easy enought to just do `configurations.register(name)`

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
modmuss50: Yeah good idea, going to be a lot of fun figuring it out though 😢

Juuxel: For *almost all* Loom configurations it'll be `canBeResolved = true, canBeConsumed = false`. `namedElements` is the opposite.

(can be resolved = has *resolvable* dependencies, can be consumed = has dependencies and artifacts to publish to *consumers*)

modmuss50: Ill likely do this as another PR, I think it would be best to split `CompileConfiguration` into two (it basically already is split, just in one file). Can still be done for 1.2 👍

PR #861, opened 2023-04-16
Upload test heap dump after OOM

modmuss50:
PR to try and figure out the falkey tests.


modmuss50:
Moving this change into: https://github.com/FabricMC/fabric-loom/pull/846

PR #866, opened 2023-04-17
Rewrite Kotlin metadata remapping.

### LINKED ISSUE #842: `KotlinPluginUtils.getKotlinMetadataVersion` returns unexpected version if kotlinx-metadata-jvm `>=0.6.0` is loaded
Reproducible notably by trying to use https://github.com/Kotlin/binary-compatibility-validator `>=0.13.0`. 

With release 0.6.0 the `Implementation-Version` field changed to a new format, at the moment being `0.6.0-1.8.0-26`. This causes `remapJar` to fail as it tries to resolve kotlinx-metadata with that version, while the actual version is only `0.6.0`

modmuss50:
The old visitor API was depreacated/broken, however the new API has some holes. With a bit of reflection and java tricks that can be solved.

I am going to look at improving the test for this before merging, as I have a feeling this wont be the last time this code will break.

Any Kotlin tips would be appricated 👍


Juuxel (approved):
LGTM, your Kotlin looks good too. The internals will probably break in some update, but that shouldn't matter as long as nothing else on the Gradle classpath upgrades the library.

PR #867, opened 2023-04-17
Allow mods to opt into being put onto the annotation processor classpath.

modmuss50:
This is designed for "mods" such as Mixin Extras that add an annotation processor.

Add the following manifest entries:

```
Fabric-Loom-Remap: false
Fabric-Loom-Annotation-Processor: true
```

I dont see it necessary to support remapped mods on the annotation processor classpath, as it gets a bit more complex with split mods and transitive dependencies.


Juuxel:
If it's not remapping the mods, I don't see why the users can't just specify
```gradle
dependencies {
  annotationProcessor 'theMod'
}
```
themselves

modmuss50:
Going to go with a diffrent approach to this.

PR #873, opened 2023-04-23
Fix incorrect log level in logger config

Juuxel:
The old level="all" meant that if the trace level would pretend to be enabled when it was not. This can make some libraries like [ModLauncher](https://github.com/McModLauncher/modlauncher/blob/0ef7e830a51df41af4de50967f57d832ae8d60a0/src/main/java/cpw/mods/modlauncher/ClassTransformer.java#L134) very spammy.

This is also *slightly* broken: if fabric.log.level is a finer level than fabric.debug.log.level, it will report the less fine level.

### DIFF
--- a/src/main/resources/log4j2.fabric.xml
+++ b/src/main/resources/log4j2.fabric.xml
@@ -52,7 +52,7 @@
 	</Appenders>
 	<Loggers>
 		<Logger level="${sys:fabric.log.level:-info}" name="net.minecraft"/>
-		<Root level="all">
+		<Root level="${sys:fabric.log.debug.level:-debug}">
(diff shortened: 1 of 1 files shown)


modmuss50:
This does break `-Dfabric.log.level=trace`, I'm not too sure if thats something we want to do.

Juuxel:
True, but with the default settings this isn't very correct either: `LoggerFactory.getLogger("abc").isTraceEnabled() == true`.

I don't think there's any other way to solve this than making a Log4J plugin with a [custom lookup](https://logging.apache.org/log4j/2.x/manual/lookups.html) that chooses the finest log level of the two.

PR #874, opened 2023-04-23
Small pref improvement in fabric API configuration. 

modmuss50:
Saves a few seconds configuring fabric API. Nothing else is really standing out when profiling :) .


modmuss50:
It broke the build :(

PR #875, opened 2023-04-24
Update Kotlin to 1.8.10 and Gradle to 8.1.1

JT122406:
Kotlin 1.8.10 is included with gradle 8 and in the project we mark that we much match kotlin version to the one in gradle so this handles that.  Along with bumping Gradle from 8.1 to 8.1.1 https://docs.gradle.org/8.1.1/release-notes.html
As 8.1.1 is suggested over 8.1 due to some bugs

image

### DIFF
--- a/.github/workflows/test-push.yml
+++ b/.github/workflows/test-push.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [8.1.0-jdk17]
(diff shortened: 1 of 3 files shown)


modmuss50:
Many thanks for this PR, however I have replaced it with: https://github.com/FabricMC/fabric-loom/pull/903

My PR validates the kotlin version matches the version used by Gradle, this will pevent this from being an issue again 👍

Issue #877, opened 2023-05-01
Loom is very slow to download files, often hanging for 10+ minutes


jellysquid3:
I'm not sure if this is an issue with Gradle or Loom itself, but in the interest of having _someone_ be aware of it...

It seems that on any Linux distro (Fedora, Ubuntu, Gentoo, and even Arch), that Loom will often take an extremely long time to download certain files, and will even sometimes hang for tens of minutes (most often while downloading assets.) 

An easy way to reproduce this on my machine is to checkout the sources for [Sodium](https://github.com/CaffeineMC/sodium-fabric/tree/1.19.4/dev) and then try to run the `./gradlew build` task after clearing the `~/.gradle/caches` directory. This almost never works and ends up getting stuck on `:downloadAssets`, where it will take 20-30 minutes to get past whatever files it is stuck on, before resuming normally.

On occasion, it seems to spit out the following network error for every file it was stuck on, which then halts the build immediately:

```
A problem occurred configuring root project 'sodium-fabric'.
> Could not resolve all files for configuration ':minecraftLibraries'.
   > Could not resolve com.mojang:authlib:3.17.30.
     Required by:
         project :
      > Could not resolve com.mojang:authlib:3.17.30.
         > Could not get resource 'https://libraries.minecraft.net/com/mojang/authlib/3.17.30/authlib-3.17.30.pom'.
            > Could not GET 'https://libraries.minecraft.net/com/mojang/authlib/3.17.30/authlib-3.17.30.pom'.
               > The server may not support the client's requested TLS protocol versions: (TLSv1.2, TLSv1.3). You may need to configure the client to allow other protocols to be used. See: https://docs.gradle.org/7.6/userguide/build_environment.html#gradle_system_properties
                  > Remote host terminated the handshake
```

While it's easy to think that this is a problem with my specific computer or network connection, it is reproducible on _any_ Linux machine that I have access to, even when using a VPN or a different network entirely (including cellular hotspot.)  But I've never been able to reproduce it on a Windows machine.
 
The only consistent thing about this issue is that Loom will usually log the following warnings before it hangs:

```
Forcing downloading $HOME/.gradle/caches/fabric-loom/assets/objects/a0/a0d43b09bbd3a65039e074cf4699175b0c4724b8 as existing lock file was found. This may happen if the gradle build was forcefully canceled.
```

[...]

PR #879, opened 2023-05-01
Fix issues with downloading assets

mjwells2002:
this change makes HttpClient instance static to ensure the internal thread pool is used, preventing thousands of threads from being spawned

leading to 2x speed improvment in downloading game assets, and atleast on my system solving issue #877 as commented on the issue thread


mjwells2002:
Closing given that #880 is this but better

modmuss50:
Ah, sorry, I got a little carried away.

What you have here isnt thread safe as `getHttpClient` can be called on many threads, easy to fix of course but I dont think there is a lot of benfit in doing it lazily. Thanks for the PR anyway, and most importantly thanks for spotting this. 👍

mjwells2002:
Ahh, i hadnt caught that getHttpClient is used on multiple threads

Issue #473, opened 2021-08-21
[1.17.1][Loom 0.9-SNAPSHOT] `class_1309.method_5773` does not get remapped to intermediary in the refmap from mojang mappings


CorgiTaco:
Generated refmap from build(Mojang names warning) https://gist.github.com/CorgiTaco/f9718553744e3cb1309c002f4e5ea765#file-refmap-json-L18

My mixin(Mojang names warning): https://github.com/CorgiTaco/Enhanced-Celestials/blob/d63926556964ad376114df41f1ab010dc93b1425/src/main/java/corgitaco/enhancedcelestials/mixin/MixinLivingEntity.java#L27-L33

My build.gradle: https://github.com/CorgiTaco/Enhanced-Celestials/blob/f8aca864cdd16f77f29649332a306091e51723a3/build.gradle

This causes the jar to break in production.

Logiquo:
Can confirm the mapping generation is incorrect. For `loom.mappings-layered+hash.961.minecraft.1.17.1` there is 
```
c	atu	net/minecraft/class_1309	net/minecraft/world/entity/LivingEntity
	m	()V	tick	tick	tick
```
which stops MCE from searching the parents.

Logiquo:
@modmuss50 I guess this could be a bug in layer mapping? Could you have a look at this?

Logiquo:
The reason is the mapping fed into ap is incorrect. With a closer look, the problem is layer mapping.
Specifically, Mojang distributes a complete mapping (i.e. contains all field/method names regardless of inheritance relationship) where an intermediary distributes an incomplete mapping (i.e. only put method/field name in super-class).
It will cause the resulting mapping tree

```
<mojang name>: [null, <obf name>]
```

where null should be an intermediary name. This situation would happen for every inheritance method due to the different distribution methods described above.

And `MappingSourceNsSwitch` will fill the null into `<mojang name>`. Which cause the following mapping to be generated

```
<mojang name> <mojang name>
```

modmuss50:
Run `./gradlew build --refresh-dependencies` to fix.

PR #881, opened 2023-05-02
Fix constructor mappings with layered mappings

char3210:
When using layered mappings, constructor parameter names were not mapped because intermediary mappings do not include constructors, and so they were discarded when switching the source namespace to intermediary. This adds another forwarding visitor that adds the `<init>` method to the destination namespaces. 
Note that it isn't possible to just remap all the missing names (see #473)

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsDependency.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsDependency.java
@@ -49,6 +49,7 @@ import net.fabricmc.loom.api.mappings.layered.MappingContext;
 import net.fabricmc.loom.api.mappings.layered.MappingLayer;
 import net.fabricmc.loom.api.mappings.layered.MappingsNamespace;
 import net.fabricmc.loom.configuration.providers.mappings.extras.unpick.UnpickLayer;
+import net.fabricmc.loom.configuration.providers.mappings.utils.AddConstructorMappingVisitor;
 import net.fabricmc.loom.util.ZipUtils;
 import net.fabricmc.mappingio.adapter.MappingDstNsReorder;
 import net.fabricmc.mappingio.adapter.MappingSourceNsSwitch;
@@ -102,7 +103,8 @@ public class LayeredMappingsDependency implements SelfResolvingDependency, FileC
 
 			MappingDstNsReorder nsReorder = new MappingDstNsReorder(tiny2Writer, Collections.singletonList(MappingsNamespace.NAMED.toString()));
 			MappingSourceNsSwitch nsSwitch = new MappingSourceNsSwitch(nsReorder, MappingsNamespace.INTERMEDIARY.toString(), true);
-			mappings.accept(nsSwitch);
+			AddConstructorMappingVisitor addConstructor = new AddConstructorMappingVisitor(nsSwitch);
+			mappings.accept(addConstructor);
 
 			Files.deleteIfExists(mappingsFile);
 			ZipUtils.add(mappingsFile, "mappings/mappings.tiny", writer.toString().getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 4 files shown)


modmuss50:
It would be nice to expand the tests to cover this, maybe add something to `ParchmentMappingLayerTest`?

char3210:
> It would be nice to expand the tests to cover this, maybe add something to `ParchmentMappingLayerTest`?

👍

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/AddConstructorMappingVisitor.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import net.fabricmc.mappingio.MappedElementKind;
+import net.fabricmc.mappingio.MappingVisitor;
+import net.fabricmc.mappingio.adapter.ForwardingMappingVisitor;
+
+/**
+ * Adds the constructor method <code>&lt;init&gt;</code> to the destination namespaces,
+ * as long as the source name is <code>&lt;init&gt;</code> and the destination mapping is null.
```
Juuxel:
```suggestion
 * Adds the constructor method {@code <init>} to the destination namespaces,
 * as long as the source name is {@code <init>} and the destination mapping is null.
```
Much more readable.

char3210: didn't know that was a thing cool

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/AddConstructorMappingVisitor.java
```diff
@@ -0,0 +1,83 @@
...
+	@Override
+	public void visitNamespaces(String srcNamespace, List<String> dstNamespaces) throws IOException {
+		namespaceVisited = new boolean[dstNamespaces.size()];
+		super.visitNamespaces(srcNamespace, dstNamespaces);
+	}
+
+	@Override
+	public boolean visitMethod(String srcName, String srcDesc) throws IOException {
+		if ("<init>".equals(srcName)) {
+			inConstructor = true;
```
modmuss50, line 56: I would set `inConstructor` to false when the method name isnt `<init>` incase `visitElementContent` is not called.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/utils/AddConstructorMappingVisitor.java
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+
+		return super.visitElementContent(targetKind);
+	}
+
+	@Override
+	public void visitDstName(MappedElementKind targetKind, int namespace, String name) throws IOException {
+		namespaceVisited[namespace] = true;
+		next.visitDstName(targetKind, namespace, name);
```
modmuss50:
```suggestion
		super.visitDstName(targetKind, namespace, name);
```

src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsDependency.java
```diff
@@ -102,7 +103,8 @@ private void writeMapping(LayeredMappingsProcessor processor, List<MappingLayer>
 
 			MappingDstNsReorder nsReorder = new MappingDstNsReorder(tiny2Writer, Collections.singletonList(MappingsNamespace.NAMED.toString()));
 			MappingSourceNsSwitch nsSwitch = new MappingSourceNsSwitch(nsReorder, MappingsNamespace.INTERMEDIARY.toString(), true);
-			mappings.accept(nsSwitch);
+			AddConstructorMappingVisitor addConstructor = new AddConstructorMappingVisitor(nsSwitch);
```
modmuss50, line 106: Would this be better in `MojangMappingLayer`?

modmuss50 (changes requested):

src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsDependency.java
char3210: I don't exactly see how this could be moved to MojangMappingLayer, and the main cause of the issue is that intermediary mappings don't include `<init>` rather than mojang mappings

modmuss50 (approved):
I had a closer look, this looks correct 👍 Thanks a lot

PR #882, opened 2023-05-04
Move jar merger to loom.

modmuss50:
- Uses loom/TR's jar file system cache
- Uses loom ASM version constant. (Saves updating stitch as well)
- Checkstyle/misc code cleanup
- Stitch is only used for the legacy V1 mappings now. Yarn already depends on loom, and will use this for merging the jars.


modmuss50:
Going to merge this now, there is no reason to wait. its a 1-1 copy so shouldnt have any issues (in theory 👍)

liach:
Will stitch be totally merged into loom in the long run then?

modmuss50:
This is the last of the "none legacy" part of stitch that we use and will need to maintain. So effectively its all been merged into loom or elsewhere. Loom will contiune to depend on stitches old auto naming stuff for v1 mappings (pre 1.14.4 basically), I dont think is worth moving into loom as its unlikely to ever need changing again. (Dropping v1 support for newer versions is something I might explore)

Issue #885, opened 2023-05-09
Use of reproducibleFileOrder produces invalid jars


lukebemish:
Java expects a manifest in a jar to be the first or second zip entry. For example: https://github.com/openjdk/jdk17u/blob/aa8c46ca52c2f983470ab4c05e64c1ae85aa3b8e/src/java.base/share/classes/java/util/jar/JarInputStream.java#L78-L85
Currently, something is off with loom's remap jar tasks; while when used with `reproducibleFileOrder = false` they seem to maintain the file order they were given, with the manifest first as it was in the pre-remapping jar, if `reproducibleFileOrder` on the task is set to `true`, then the special ordering with the manifest first is destroyed, producing a jar which may cause strange behavior. For a simple reproduction case, see https://github.com/lukebemish/loom-jar-ordering-demo

lukebemish:
I think a good question here is: does the remapped jar have a guarantee to have the same file order as the pre-remapping jar? Because if so, then I don't even need to set that option on that task. However, it's not obvious to me that that is necessarily true across systems or whatever, so some guarantee of that would be useful

modmuss50:
Where does this cause an issue? As Fabric API has had this enabled for a long while now? https://github.com/FabricMC/fabric/blob/a37e8e70b73d4283ebd6081921fb32b9fb61e7b1/build.gradle#L240-L243 Or does tasks.withType not work as I expect?

lukebemish:
If you are consuming the produced jars with some external tool (in my case, something that runs on our maven server), this external tool assumes stuff is a valid jar. If it is using OpenJDK, OpenJDK's implementation of `JarInputStream` (linked in my original comment) assumes that a manifest is the first or second entry of a jar file. `remapJar` with `reproducibleFileOrder` set to true can (and sometimes does!) produce jars where this is not the case

lukebemish:
This does not seem to be the case for fabric api - I would guess that the ordering is redone when the jar is signed to put the manifest first

modmuss50:
Ah ok, that makes sense. I was wondering why we had not seen any issues before. The solution to this issue is to just sort the manifest first. Here: https://github.com/FabricMC/fabric-loom/blob/dev/1.2/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java#L57

PR #887, opened 2023-05-11
Change zip entry ordering to place manifest first

lukebemish:
This should solve #885; I've tested it with the reproduction case I provided for that issue, and it correctly reorders the zip entries.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
+++ b/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
@@ -45,6 +45,51 @@ public class ZipReprocessorUtil {
 
 	private ZipReprocessorUtil() { }
 
+	private static final String MANIFEST_LOCATION = "META-INF/MANIFEST.MF";
+	private static final String META_INF = "META-INF/";
+
+	// See https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html#signed-jar-file
+	private static boolean isSpecialFile(String zipEntryName) {
+		if (!zipEntryName.startsWith(META_INF)) {
+			return false;
+		}
+
+		String[] parts = zipEntryName.split("/");
+
+		if (parts.length != 2) {
+			return false;
+		}
+
+		return parts[1].startsWith("SIG-")
+				|| parts[1].endsWith(".SF")
+				|| parts[1].endsWith(".DSA")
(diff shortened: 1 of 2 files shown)


lukebemish:
Marking as a draft as it looks like there's other files this will need to handle too: https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html#signed-jar-file

lukebemish:
Alright, that implementation should match what java expects for jars. For posterity's sake, the use of a `JarInputStream` (which isn't great to use, but is still used quite often) assumes that:
> * The Manifest is the first entry in the stream (or the second entry if the first entry in the stream is META-INF/ and the second entry is META-INF/MANIFEST.MF).
> * All signature-related entries immediately follow the Manifest

(from https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/util/jar/JarInputStream.html)

modmuss50:
I think the https://github.com/FabricMC/fabric-loom/blob/dev/1.2/src/test/groovy/net/fabricmc/loom/test/integration/ReproducibleBuildTest.groovy tests are going to need updating. Might be kinda nice to read the jars with JarInputStream as well.

lukebemish:
Hmm, wait, that doesn't seem to be working quite as expected... yeah, let me try and see if I can get all the hashes correct

lukebemish:
Alright, that should be correct. At any rate, the test runs locally, and I even switched the line endings around to get both source possibilities..

modmuss50:
The tests arent working still. The mod jar seems to now be `174c9b52f4bc6d489548d11b42e853cf` I reccomend getting the hash from the github actions build.

lukebemish:
> The tests arent working still. The mod jar seems to now be `174c9b52f4bc6d489548d11b42e853cf` I reccomend getting the hash from the github actions build.

That is... worrying, then, since that's not the hash when I run it locally (and, well, it should be if we're going for reproducible builds)

lukebemish:
Alright, I've changed the hashes - I'm still worried about why it differs locally vs on the test here, but honestly I have no clue. The sources jar hashes may be out of date - I don't have an easy way to get those now

lukebemish:
Alright, the source hashes should be updated. I'm still not sure why it didn't match locally, but it seems alright here at least.

modmuss50 (approved):

PR #893, opened 2023-05-24
Look into supporting config caching

modmuss50:
This is highly broken and more of an experiement to see what is broken, it turns out a huge amount of stuff is!

~~Main blocker is Gradle's config cache serialization codec's do not seem to supports [records at all](https://gist.github.com/modmuss50/8f7fea795eb5de2c10af1d3027474d62). I only found this out after I finally got it to a place where it might have a chance of working. I dont see a way forward until this is supported by Gradle.~~ (Worked around by wrapping the object, and reading/writing to a json string.

This may be slower in some cases as the tasks can no longer share the same services, really needs testing. This also funimentally breaks MPO.


modmuss50:
Closing in favour of #1053 we will take an incremental approach to supporting this.

PR #899, opened 2023-06-08
Enable reproducible builds by default

### LINKED ISSUE #898: Enable reproduce builds by default.
Basically what the title says.

modmuss50:
Closes #898

This should likely target Loom 1.3

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -40,7 +40,6 @@ import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.tasks.AbstractCopyTask;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskContainer;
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -83,6 +84,12 @@ public void run() {
 			javadoc.setClasspath(main.getOutput().plus(main.getCompileClasspath()));
 		});
 
+		// Make all outputs to be reproducible
+		getTasks().withType(AbstractArchiveTask.class).configureEach(abstractArchiveTask -> {
```
Juuxel: Since `withType` is a live collection and `configureEach` is lazy, does this affect user-defined tasks? If it does, I'm wondering whether it'd be better to just set defaults for AbstractRemapJarTasks.

modmuss50: That might be better, I had hoped to be able to set the convention on these.

lukebemish:
Let's not do this till #887 is merged or you'll be making a lot of invalid jars

PR #900, opened 2023-06-10
Fix interface injection of inner class interfaces

### LINKED ISSUE #798: Injecting inner class interfaces does not add appropriate InnerClass attributes
When injecting an interface [that is an inner class](https://github.com/KyoriPowered/adventure-platform-fabric/blob/mc/1.19/src/main/resource-templates/fabric.mod.yaml#L32), the appropriate superinterface information will be added to the classfile, but an InnerClass link like `javac` would generate is not added. This does not affect the actual build, but seems to cause some IDEs to choke up on the modified game classes (*looks at JetBrains*) 

I was sent the following IDE error by a downstream consumer of my library mod, caused by this inner class injection:

```
Cannot access 'net.kyori.adventure.sound.Sound$Emitter' which is a supertype of 'net.minecraft.server.network.ServerPlayerEntity'. Check your module classpath for missing or conflicting dependencies
```

(when hovering over a reference to `ServerPlayerEntity#networkHandler` in a Kotlin project)

I was able to work around this with [a patch that synthesizes inner class attributes](https://github.com/jpenilla/quiet-loom/blob/7e7c4e14a4d8a8251f8301249a2d4ef0fee6ef09/patches/0009-Quilt-Fix-interface-injection-of-inner-class-interfa.patch), but it's far from a clean patch there, just something I whipped up in an evening. Since interface injections can come from the active project as well as external dependencies, there's not necessarily bytecode that's been compiled yet to check to see if an internal name actually refers to an inner class rather than a funkily named top-level class, so we have to make a guess based on the class name alone (at least in some cases).

Whether it's worth 'fixing' this with something that hacky just because of one IDE vendor's incompetence is debatable, but I for one would very much be happy to have the mod devs stop bugging me about the issue.

zml2008:
Fixes #798

This ports forward a patch I had from ages ago without much re-review since I'm unfortunately short on time these days, let me know how it looks and i can maybe hopefully potentially give it another once-over in a week or two. See the issue for a rundown of the original reasoning behind the patch.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):
Looks good 👍 

If there are issues we can easily fix them later.

PR #907, opened 2023-06-13
Move unpick to GenerateSourcesTask and rebuild input jars

modmuss50:


src/main/java/net/fabricmc/loom/task/GenerateSourcesTask.java
```diff
@@ -90,26 +96,37 @@ public abstract class GenerateSourcesTask extends AbstractLoomTask {
...
+	 * The jar name to decompile, {@link MinecraftJar#getName()}.
 	 */
-	@InputFile
-	public abstract RegularFileProperty getInputJar();
-
-	/**
-	 * The jar used at runtime.
-	 */
-	@InputFile
-	public abstract RegularFileProperty getRuntimeJar();
```
Juuxel, line 102: As tasks are part of the API (I think? They're very clearly and intentionally exposed to buildscripts), isn't this + adding a bunch of new required properties + removing the unpick task a breaking change?

modmuss50: The Task API surface is not clearly defined (Something to possibly improve later), as far as I see it all of these properties past and present are not part of the public API, I doubt (hope) that no one was actually setting these.

modmuss50:
Should be good to merge now (assuming the tests pass). This doesnt actually change anything with the line numbers, but will no longer apply the line mapping over the top of a previosly line mapped jar.

If there are still issues with the line mapping it should be a lot easier to reproduce in a consitant way. With the new intergration test that I added I am confident we can finally get to the bottom of this issue. So please report any issues you have with debugging after this PR has gone in. 👍

PR #917, opened 2023-06-30
Create RunConfigSettings with Gradle's object factory

modmuss50:


src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
```diff
@@ -209,8 +222,13 @@ public void environment(String environment) {
 		setEnvironment(environment);
 	}
 
+	/**
+	 * @deprecated Replace with {@link RunConfigSettings#setConfigName(String)}
+	 */
 	public void name(String name) {
```
Juuxel:
```suggestion
	@Deprecated
	public void name(String name) {
```

modmuss50: That might be helpful 👍 Thanks

modmuss50:
Updated the fabric API test to run against my PR https://github.com/FabricMC/fabric/pull/3158 fixing the new deprecations added in this version.

src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
```diff
@@ -160,11 +173,11 @@ public void setEnvironment(String environment) {
 	}
 
 	public String getConfigName() {
-		return name;
+		return configName;
 	}
 
 	public void setConfigName(String name) {
```
Juuxel, line 173: Should it also have a `configName(String)` method to match the rest? I don't personally think it's needed, but it could be good for completeness. I guess it'd go away in 2.0 if this class moves to lazy properties though.

modmuss50: Good question, for completeness yes, but for the sake of it being removed in 2.0 I dont think its needed.

modmuss50: Just thinking about this, it should really be a display name, it feels a little weird deprecating this to something we arent 100% happy with.

Juuxel: Well, it could be `Property<String> getDisplayName()` instead? With the old methods delegating to the property.

modmuss50: Thats easier said than done, the default name is set in RunConfig, I think for now ill just do the easy change to make this use an object factory without breaking or deprecating anything. I really just want to get 1.3 done.

PR #923, opened 2023-07-04
Fix included none mod libraries not being reproducible.

### LINKED ISSUE #921: Include configuration is not reproducible
Using the `include` configuration to add non-mod libraries as "mods" to the remap jar creates non-reproducible output.  
For a minimal example I modified the example mod, built the mod, and checked the difference in output over subsequent builds.

What I believe causes this issue is that the generated `fabric.mod.json` does not use a static timestamp, but instead is the time at which the build is run.

Here's the exact modifications I made to the example build script, using Loom 1.3 since reproducible builds are enabled by default:
```diff
diff --git a/build.gradle b/build.gradle
index a8bdc34..cdf8eb4 100644
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-       id 'fabric-loom' version '1.2-SNAPSHOT'
+       id 'fabric-loom' version '1.3.5'
        id 'maven-publish'
 }

@@ -36,13 +36,7 @@ dependencies {
        mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
        modImplementation "net.fabricmc:fabric-loader:${project.loader_version}"

-       // Fabric API. This is technically optional, but you probably want it anyway.
-       modImplementation "net.fabricmc.fabric-api:fabric-api:${project.fabric_version}"
-
-       // Uncomment the following line to enable the deprecated Fabric API modules.
-       // These are included in the Fabric API production distribution and allow you to update your mod to the latest modules at a later more convenient time.
-
-       // modImplementation "net.fabricmc.fabric-api:fabric-api-deprecated:${project.fabric_version}"
+       include("commons-io:commons-io:2.13.0")
 }
```

modmuss50:
Closes #921

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
+++ b/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
@@ -95,44 +95,80 @@ public class ZipReprocessorUtil {
 			return;
 		}
 
-		try (ZipFile zipFile = new ZipFile(file)) {
+		try (var zipFile = new ZipFile(file)) {
 			ZipEntry[] entries;
 
 			if (reproducibleFileOrder) {
-				entries = zipFile.stream().sorted(Comparator.comparing(ZipEntry::getName, ZipReprocessorUtil::specialOrdering)).toArray(ZipEntry[]::new);
+				entries = zipFile.stream()
+						.sorted(Comparator.comparing(ZipEntry::getName, ZipReprocessorUtil::specialOrdering))
+						.toArray(ZipEntry[]::new);
 			} else {
-				entries = zipFile.stream().toArray(ZipEntry[]::new);
+				entries = zipFile.stream()
+						.toArray(ZipEntry[]::new);
 			}
 
-			ByteArrayOutputStream outZip = new ByteArrayOutputStream(zipFile.size());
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
```diff
@@ -95,44 +95,80 @@ public static void reprocessZip(File file, boolean reproducibleFileOrder, boolea
...
-					while ((length = inputStream.read(buf)) > 0) {
-						zipOutputStream.write(buf, 0, length);
-					}
+			try (var fileOutputStream = new FileOutputStream(file)) {
+				outZip.writeTo(fileOutputStream);
+			}
+		}
+	}
+
+	public static void appendZipEntry(File file, String path, byte[] data) throws IOException {
```
Juuxel, line 131: Might want to add a comment about how this differs from `ZipUtils.add`

modmuss50: Done 👍

modmuss50:
Uhh, ffs github lost the commit adding the comment, I pushed before merging..

LostLuma:
Thanks! 🦫

PR #938, opened 2023-08-02
Fix and test record access flags in 23w31a.

modmuss50:
This fixes empty records not having the correct record attribute, without this some decompilers fail to decompile the class as a record.

Fixes: https://bugs.mojang.com/browse/MC-264564


modmuss50:
Fixed upstream, going to close this PR as its not strictly required for our usecase.

PR #945, opened 2023-08-17
Add extension option to configure known INVOKEDYNAMIC BSMs

lukebemish:
This PR makes use of the new feature added in my recent tiny-remapper PR to specify indy BSMs that shouldn't throw errors during remapping; a set property is added to the loom API extension to configure these, with reasonable defaults set up, including the groovy indy BSM that has been causing significant log spam issues for GroovyMC in the past.


src/main/java/net/fabricmc/loom/task/service/TinyRemapperService.java
```diff
@@ -126,8 +126,8 @@ private static IMappingProvider gradleMixinMappingProvider(SharedServiceManager
 	// Set to true once remapping has started, once set no inputs can be read.
 	private boolean isRemapping = false;
 
-	public TinyRemapperService(List<IMappingProvider> mappings, boolean useMixinExtension, @Nullable KotlinClasspath kotlinClasspath) {
-		TinyRemapper.Builder builder = TinyRemapper.newRemapper();
+	public TinyRemapperService(List<IMappingProvider> mappings, boolean useMixinExtension, @Nullable KotlinClasspath kotlinClasspath, LoomGradleExtension extension) {
```
modmuss50: I wouldnt pass the extension through like this, with the way things are moving in Gradle the idea is to move these services away from doing this. (kinda hard to explain here, see Graadle's docs on config caching.

I would change this to pass the `Set<String>` through the constructor

src/main/java/net/fabricmc/loom/task/service/TinyRemapperService.java
```diff
@@ -87,7 +87,7 @@ public static synchronized TinyRemapperService getOrCreate(SharedServiceManager
 				mappings.add(gradleMixinMappingProvider(serviceManager, project.getGradle(), extension.getMappingConfiguration().mappingsIdentifier, from, to));
```
modmuss50, line 90: (I cannot put a comment on the lines above, but the StringJoiner on line 68 needs to contain either these entries, or some sort of hash, so if multiple projects have diffrent values they will get their own remapper.

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionApiImpl.java
```diff
@@ -98,6 +101,11 @@ protected LoomGradleExtensionApiImpl(Project project, LoomFiles directories) {
 		this.log4jConfigs = project.files(directories.getDefaultLog4jConfigFile());
 		this.accessWidener = project.getObjects().fileProperty();
 		this.customManifest = project.getObjects().property(String.class);
+		this.knownIndyBsms = project.getObjects().setProperty(String.class).convention(Set.of(
+				"java/lang/invoke/StringConcatFactory",
+				"java/lang/runtime/ObjectMethods",
+				"org/codehaus/groovy/vmplugin/v8/IndyInterface"
+		));
```
modmuss50, line 108: Add a `this.knownIndyBsms.finalizeValueOnRead();` after this line.

modmuss50 (changes requested):

modmuss50 (approved):

PR #946, opened 2023-08-22
Update to Gradle 8.3, and update all other deps.

modmuss50:
Gradle 8.3 is required to update `kotlinx-metadata-jvm` due to the Kotlin 1.9.0 requirement.


src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
```diff
@@ -58,39 +58,40 @@ class KotlinClassMetadataRemappingAnnotationVisitor(private val remapper: Remapp
...
+                var kpackage = metadata.kmPackage
                 kpackage = KotlinClassRemapper(remapper).remap(kpackage)
-                val remapped = KotlinClassMetadata.writeMultiFileClassPart(kpackage, metadata.facadeClassName, metadata.annotationData.metadataVersion, metadata.annotationData.extraInt).annotationData
+                val wrapper = KotlinClassMetadataWrapper(metadata)
+                val remapped = KotlinClassMetadata.writeMultiFileClassPart(kpackage, metadata.facadeClassName, wrapper.annotationData.metadataVersion, wrapper.annotationData.extraInt)
                 writeClassHeader(remapped)
                 validateKotlinClassHeader(remapped, header)
             }
-            is KotlinClassMetadata.MultiFileClassFacade, is KotlinClassMetadata.Unknown, null -> {
+            is KotlinClassMetadata.MultiFileClassFacade, is KotlinClassMetadata.Unknown -> {
```
modmuss50, line 94: Note for future self:

As per the [migration guide](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/Migration.md#migrating-from-06x-to-070), we may want to try catch here.

build.gradle
```diff
@@ -85,9 +84,9 @@ dependencies {
 	}
 
 	// tinyfile management
-	implementation ('net.fabricmc:tiny-remapper:0.8.7')
+	implementation ('net.fabricmc:tiny-remapper:0.8.8')
 	implementation 'net.fabricmc:access-widener:2.1.0'
-	implementation 'net.fabricmc:mapping-io:0.2.1'
+	implementation 'net.fabricmc:mapping-io:0.4.2'
 
 	implementation ('net.fabricmc:lorenz-tiny:4.0.2') {
```
modmuss50: Might be a good idea to build a new copy of lorenz-tiny against the latest mapping-io.

gradle/libs.versions.toml
```diff
@@ -2,7 +2,19 @@
 kotlin = "1.9.0"
 
 [libraries]
+# Decompilers
+fernflower = { module = "net.fabricmc:fabric-fernflower", version = "2.0.0" }
+cfr = { module = "net.fabricmc:cfr", version = "0.2.1" }
+
+# Runtime depedencies
+mixin-compile-extenions = { module = "net.fabricmc:fabric-mixin-compile-extensions", version = "0.6.0" }
```
Juuxel: Should be `-extensions` (won't make a suggestion as it's used in quite a few places)

modmuss50: Thanks :)

src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
modmuss50: Going to leave this for now, better to crash and then figure out the error vs contiuning. If anyone has any issues with this please open an issue 👍

PR #949, opened 2023-09-09
Support reading "development" dependencies from the installer json.

modmuss50:
This is needed for some upcoming loader changes.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/InstallerData.java
+++ b/src/main/java/net/fabricmc/loom/configuration/InstallerData.java
@@ -24,18 +24,26 @@
 
 package net.fabricmc.loom.configuration;
 
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ExternalModuleDependency;
 import org.gradle.api.artifacts.repositories.MavenArtifactRepository;
+import org.gradle.api.plugins.JavaPlugin;
 
 import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.LoomRepositoryPlugin;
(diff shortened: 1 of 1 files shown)


src/main/java/net/fabricmc/loom/configuration/InstallerData.java
```diff
@@ -45,35 +53,61 @@ public void applyToProject(Project project) {
...
-			project.getLogger().debug("Adding dependency ({}) from installer JSON", name);
+			LOGGER.debug("Adding dependency ({}) from installer JSON", name);
 
 			ExternalModuleDependency modDep = (ExternalModuleDependency) project.getDependencies().create(name);
-			modDep.setTransitive(false);
+			modDep.setTransitive(false); // Match the launcher in not being transitive
 			loaderDepsConfig.getDependencies().add(modDep);
 
-			// TODO: work around until https://github.com/FabricMC/Mixin/pull/60 and https://github.com/FabricMC/fabric-mixin-compile-extensions/issues/14 is fixed.
+			// Work around https://github.com/FabricMC/Mixin/pull/60 and https://github.com/FabricMC/fabric-mixin-compile-extensions/issues/14 is fixed.
```
Juuxel:
```suggestion
			// Work around until https://github.com/FabricMC/Mixin/pull/60 and https://github.com/FabricMC/fabric-mixin-compile-extensions/issues/14 is fixed.
```

Juuxel (approved):

src/main/java/net/fabricmc/loom/configuration/InstallerData.java
modmuss50: I dont think we can ever remove this tbh, it makes sense to keep.

Juuxel: Read the whole comment until the end 😅

modmuss50: Good idea :D Should be better now 👍

PR #951, opened 2023-09-09
Add Vineflower support

modmuss50:
Adds support for the [Vineflower](https://github.com/Vineflower/vineflower) decompiler hosted on Maven central. The code is based off the exisitng fernflower support, each decompiler is split into its own sourceset and lazily downloaded from maven.

Custom vineflower versions can be specifed like so.
```gradle
dependencies {
	vineflowerDecompilerClasspath "org.vineflower:vineflower:1.9.3"
}
```

The existing loom decompiler API is left unchnaged, and can be used to configure custom options as required.


modmuss50:
Build failuire is caused by this Kotlin bug... https://youtrack.jetbrains.com/issue/KT-60879

PR #952, opened 2023-09-09
Fix reproducable zips across timezones. 

### LINKED ISSUE #926: Reproducible ZIP timestamps are dependent on machine timezone
Currently when building with reproducibility enabled the timestamps inside the archive are dependent on the timezone of the machine building the mod. This diverges from the way Gradle itself handles this, as it always outputs a consistent timestamp of `1980-02-01 00:00`, regardless of build environment (at least from my testing).

A few examples obtained using `unzip -v modid-1.0.0.jar` on the example mod built in different environments:

Built via GHA, where I assume the machine time zone is UTC:
```sh
Archive:  modid-1.0.0-actions-build.jar
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
     208  Defl:N      148  29% 1980-02-01 01:00 ba7d7dad  modid.mixins.json
```
And from building the exact same commit on my local machine (currently CEST?):
```sh
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
     208  Defl:N      148  29% 1980-02-01 00:00 ba7d7dad  modid.mixins.json
```

I assume this is a difference in how timestamps are handled by Java's built in ZIP-related library and the apache compress library which is used by Gradle instead, since the constant timestamp used is exactly the same.

modmuss50:


modmuss50:
Fixed, ~~the root cause was using Febuary and not January...~~

This java commit here was very helpful in fixing and understanding this: https://github.com/openjdk/jdk/commit/99136026b8d7468ff80ebaf4372fd7709edc1c76

PR #962, opened 2023-10-12
Run tests against Java 21 and latest Gradle nightly

modmuss50:
No need to merge, just running the tests to see if there are any issues.


modmuss50:
Humm, there is 1 intresting failure:

> Build was configured to prefer settings repositories over project repositories but repository 'MavenCentralLWJGL' was added by plugin 'fabric-loom'

I think the only option here will be to always get LWJGL from central 🤔

PR #973, opened 2023-11-01
Raspberry Pi support

modmuss50:
I recently got a Raspberry Pi 5, first thing I did was to try and run Minecraft on it. Turns it out its actually playable, once you set the `MESA_GL_VERSION_OVERRIDE=4.3` env var. Loom already handles adding the linux arm64 natives, so why not set this env var for a flawless experience on a Pi5.

I have limited this to debian bookworm (Raspbian) as its the easyist thing for me to test with.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
@@ -310,7 +310,7 @@ public class RunConfigSettings implements Named {
 		defaultMainClass(Constants.Knot.KNOT_CLIENT);
 
 		if (Platform.CURRENT.isRaspberryPi()) {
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/configuration/ide/RunConfigSettings.java
```diff
@@ -308,6 +308,11 @@ public void client() {
 		startFirstThread();
 		environment("client");
 		defaultMainClass(Constants.Knot.KNOT_CLIENT);
+
+		if (Platform.CURRENT.isRaspberryPi()) {
+			getProject().getLogger().lifecycle("Raspberry Pi detected, setting MESA_GL_VERSION_OVERRIDE=4.3");
```
Juuxel: Nit: Maybe `info` is enough? I'm not sure this needs to be printed out every configuration cycle.

modmuss50: Done 👍

PR #974, opened 2023-11-02
Update MIO to 0.5.0

modmuss50:


src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsLayer.java
```diff
@@ -78,7 +78,7 @@ private void visit(Path path, MappingVisitor mappingVisitor) throws IOException
 		);
 		MappingNsRenamer renamer = new MappingNsRenamer(nsSwitch, fallbackNamespaceReplacements);
 
-		MappingReader.read(path, enigma ? MappingFormat.ENIGMA : null, renamer);
+		MappingReader.read(path, enigma ? MappingFormat.ENIGMA_DIR : null, renamer);
```
NebelNidas, line 81: Should probably rename the `enigma` boolean to `enigmaDir` too

PR #976, opened 2023-11-05
Support mixins without refmaps in mod dependencies

Juuxel:
This probably needs some testing, I'm also not sure how I'd make a new integration test for this.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
+++ b/src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
@@ -32,8 +32,8 @@ import java.util.List;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
-import org.gradle.api.logging.Logger;
-import org.gradle.api.logging.Logging;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import net.fabricmc.loom.LoomGradlePlugin;
 import net.fabricmc.loom.util.FileSystemUtil;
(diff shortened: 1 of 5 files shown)


src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
```
modmuss50: Nit: use SLF4j logging libraries.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
```diff
@@ -0,0 +1,79 @@
...
+
+	public static boolean hasMixinsWithoutRefmap(Path modJar) throws IOException {
+		final List<String> mixinConfigs = getMixinConfigs(modJar);
+
+		if (!mixinConfigs.isEmpty()) {
+			try (FileSystemUtil.Delegate fs = FileSystemUtil.getJarFileSystem(modJar)) {
+				for (String mixinConfig : mixinConfigs) {
+					final Path configPath = fs.getPath(mixinConfig);
+
+					if (Files.exists(configPath)) {
```
modmuss50: Nit: remove some of the indents, with a contiune, no need to wrap the whole thing in this if. Could maybe move the bellow code into its own method as well.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
```diff
@@ -0,0 +1,79 @@
...
+			}
+		}
+
+		return false;
+	}
+
+	private static List<String> getMixinConfigs(Path modJar) {
+		// Nullable because we don't care here if we can't read it.
+		// We can just assume there are no mixins.
+		final FabricModJson fabricModJson = FabricModJsonFactory.createFromZipNullable(modJar);
```
modmuss50, line 71: We should be mindful of peformance here as its called for every mod dep every gradle run, I would move this within the above `FileSystemUtil.getJarFileSystem` to prevent opening it twice?

I think the FMJ has already been read at this point in time, id have to look at the spagetti code to see how hard it would be to pass this through to here.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
```diff
@@ -0,0 +1,79 @@
...
+
+import net.fabricmc.loom.LoomGradlePlugin;
+import net.fabricmc.loom.util.FileSystemUtil;
+import net.fabricmc.loom.util.fmj.FabricModJson;
+import net.fabricmc.loom.util.fmj.FabricModJsonFactory;
+
+public final class MixinDetector {
+	private static final Logger LOGGER = Logging.getLogger(MixinDetector.class);
+
+	public static boolean hasMixinsWithoutRefmap(Path modJar) throws IOException {
```
modmuss50, line 42: Should be easy to unit test :)

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
```diff
@@ -0,0 +1,79 @@
...
+					if (Files.exists(configPath)) {
+						try (BufferedReader reader = Files.newBufferedReader(configPath)) {
+							final JsonObject json = LoomGradlePlugin.GSON.fromJson(reader, JsonObject.class);
+
+							if (!json.has("refmap")) {
+								// We found a mixin config with no refmap, exit the loop.
+								return true;
+							}
+						} catch (JsonParseException e) {
+							LOGGER.warn("Could not parse mixin config {} from jar {}", mixinConfig, modJar.toAbsolutePath(), e);
```
modmuss50: Would a fatal error be better here? Its just going to fail later on.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
Juuxel: Gradle's logging API is also used elsewhere (`Checksum`), but sure, I can change it since it's the same API.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
Juuxel: Yeah, I can combine the file system access. If there is a simple point where you can pass the FMJ through, that works as well.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
modmuss50: It honestly doesnt matter much, for newer stuff I think I try to use SLF4j. Maybe ill go and cleanup the others.

src/main/java/net/fabricmc/loom/configuration/mods/MixinDetector.java
Juuxel: I changed it to a RuntimeException (also resolves the logger issue completely since there's no more logger)

modmuss50 (approved):
Looks good, nice tests 👍

PR #977, opened 2023-11-07
Gradle 8.4, Kotlin 1.9.10

JT122406:
Updates to Gradle 8.4 and Kotlin 1.9.10 which is the version included in 8.4

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-kotlin = "1.9.0"
+kotlin = "1.9.10"
 asm = "9.5"
 commons-io = "2.13.0"
 gson = "2.10.1"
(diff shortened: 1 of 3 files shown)


modmuss50:
Thanks for the PR, is there a need for this? Generally we only bump the gradle version when required. In 1.3 this was due to the kotlin metadata library requiring a newer Kotlin version.

When bumping the gradle version the min version should also be bumped in the bootstrap.

JT122406:
Seemed like y'all were bumping gradle with almost every new loom version

modmuss50:
> Seemed like y'all were bumping gradle with almost every new loom version

Maybe we should make that rule? Usually there is a good reason to (either new features, or a forced update due to another dep).

JT122406:
Yeah lol

JT122406:
but you see what I mean

PR #980, opened 2023-11-08
Add Fabric-Loom-Mixin-Remap-Type manifest entry

modmuss50:
This guarantees that a mod with tiny-remapper remapped mixins will be remapped back with TR.

Better intergration tests to follow.


src/main/java/net/fabricmc/loom/configuration/mods/ArtifactMetadata.java
```diff
@@ -100,4 +112,15 @@ private Predicate<ArtifactMetadata> getShouldRemap() {
 			return shouldRemap;
 		}
 	}
+
+	public enum RefmapRemapType {
```
Juuxel: Should this be `MixinRemapType`, like the attribute?

src/main/java/net/fabricmc/loom/configuration/mods/ArtifactMetadata.java
```diff
@@ -58,11 +61,20 @@ public static ArtifactMetadata create(ArtifactRef artifact) throws IOException {
 			if (Files.exists(manifestPath)) {
 				final var manifest = new Manifest(new ByteArrayInputStream(Files.readAllBytes(manifestPath)));
 				final Attributes mainAttributes = manifest.getMainAttributes();
-				final String value = mainAttributes.getValue(MANIFEST_REMAP_KEY);
+				final String remapKey = mainAttributes.getValue(MANIFEST_REMAP_KEY);
```
Juuxel: Nit: this isn't really a "remap key", I'm pretty sure the `_KEY` in the attribute field name is "key" as in a map key in the attribute map

modmuss50:
I will create a PR for previous 1.x builds to throw an error when reading jars that contain this manifest entry as they wont be backwards compatible.

PR #984, opened 2023-11-19
Remapper extensions

modmuss50:
Adds an API to allow other gradle plugins to have a place within tiny remapper.

This can be used by:

- Custom tools to remap anntoations
- Alternative JVM language remappers

TODO:
- Move the kotlin remapper stuff to be implemented ontop of this API? (still kept within Loom)


src/main/java/net/fabricmc/loom/extension/RemapperExtensionHolder.java
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	public void apply(TinyRemapper.Builder tinyRemapperBuilder, String sourceNamespace, String targetNamespace, ObjectFactory objectFactory) {
+		final RemapperExtension<?> remapperExtension = newInstance(objectFactory);
+
+		tinyRemapperBuilder.extraPostApplyVisitor(new RemapperExtensionImpl(remapperExtension, sourceNamespace, targetNamespace));
+
+		if (remapperExtension instanceof TinyRemapperExtension tinyRemapperExtension) {
+			if (tinyRemapperExtension.getAnalyzeVisitorProvider() != null) {
+				tinyRemapperBuilder.extraAnalyzeVisitor(tinyRemapperExtension.getAnalyzeVisitorProvider());
```
Player3324: I wouldn't re-query these to allow expensive in place creation

modmuss50: Sorted, thanks.

PR #985, opened 2023-11-20
Skip remapping in `AbstractRemapJarTask`s when source and target namespaces match

jpenilla:
The "remap jar" tasks have much more functionality than simply remapping jars at this point, such as adding namespace metadata, nesting jars, ensuring reproducible builds, etc. Some custom build logic may want to take advantage of these features without the full overhead of no-op remapping with TinyRemapper/Mercury.


modmuss50:
Good idea! This could do with a intergration tests, can likely just add it to an existing one. Such as the [simple](https://github.com/FabricMC/fabric-loom/blob/dev/1.4/src/test/resources/projects/simple/build.gradle) one.

jpenilla:
Added to the `simple` integration test.

modmuss50 (approved):
Looks great, thanks 👍

PR #987, opened 2023-11-20
Remove Minecraft libraries from the remap classpath.

modmuss50:
Lets see how much this breaks :D

Same idea as: https://github.com/FabricMC/fabric-loader/pull/856


modmuss50:
Tests passed, going to merge, Let me know if this breaks anything.

PR #991, opened 2023-11-22
Filter out realms connection error message from debug log

dicedpixels:
Filter out realms connection error exception message from debug log. Using another `<RegexFilter>` did not work, hence the current regex.

img

There are two log entries related to not being able to connect to Realms. Only the log entry related to the exception, `Couldn't connect to realms` is removed.

Sample log attached.

[log.txt](https://github.com/FabricMC/fabric-loom/files/13442639/log.txt)

### DIFF
--- a/src/main/resources/log4j2.fabric.xml
+++ b/src/main/resources/log4j2.fabric.xml
@@ -4,8 +4,12 @@
 
 		<!--	System out	-->
 		<Console name="SysOut" target="SYSTEM_OUT">
-			<!-- Filter out the authentication error when starting in development -->
-			<RegexFilter regex="^Failed to verify authentication|^Couldn't connect to realms$" onMatch="DENY" onMismatch="ACCEPT"/>
+			<!-- Filter out the authentication errors when starting in development -->
+			<Filters>
+				<RegexFilter regex="^Failed to verify authentication$" onMatch="DENY" onMismatch="NEUTRAL"/>
+				<RegexFilter regex="^Failed to fetch user properties$" onMatch="DENY" onMismatch="NEUTRAL"/>
+				<RegexFilter regex="^Couldn't connect to realms$" onMatch="DENY" onMismatch="NEUTRAL"/>
+			</Filters>
 			<PatternLayout>
 				<LoggerNamePatternSelector defaultPattern="%style{[%d{HH:mm:ss}]}{blue} %highlight{[%t/%level]}{FATAL=red, ERROR=red, WARN=yellow, INFO=green, DEBUG=green, TRACE=blue} %style{(%logger{1})}{cyan} %highlight{%msg%n}{FATAL=red, ERROR=red, WARN=normal, INFO=normal, DEBUG=normal, TRACE=normal}" disableAnsi="${sys:fabric.log.disableAnsi:-true}">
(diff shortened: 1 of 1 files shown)


modmuss50:
You filter out `Couldn't connect to realms` but in the PR desc you say the error is `Could not authorize you against Realms server: Invalid session id`. Which one is correct?

dicedpixels:
> You filter out `Couldn't connect to realms` but in the PR desc you say the error is `Could not authorize you against Realms server: Invalid session id`. Which one is correct?

I've fixed the PR description. What I wanted to highlight is that only the exception `Couldn't connect to realms` is removed.

`Could not authorize you against Realms server: Invalid session id` is printed as an info so, I did not remove it.

dicedpixels:
It seems like 1.20.3-pre3 changed the `InvalidCredentialsException` message. Now it's this: 
image

If the way I've done the `RegexFilter` is okay, I can include this as well.

modmuss50:
> If the way I've done the `RegexFilter` is okay, I can include this as well.

Yes, if we can include them all that would be great, shame they keep changing it 😆 

Maybe it would be easier to read if there was only 1 message per `RegexFilter`? Im not too bothered either way.

dicedpixels:
> Maybe it would be easier to read if there was only 1 message per `RegexFilter`? Im not too bothered either way.

I agree too but defining the filters like

```
<Filters>
  <RegexFilter regex="^Failed to verify authentication$" onMatch="DENY" onMismatch="ACCEPT"/>
  <RegexFilter regex="^Failed to fetch user properties$" onMatch="DENY" onMismatch="ACCEPT"/>
  <RegexFilter regex="^Couldn't connect to realms$" onMatch="DENY" onMismatch="ACCEPT"/>
</Filters>
```
doesn't seem to work as expected. It filters using the 1st regex and skips the other two. I couldn't find anything explaining this behavior online either. I tried different regexes as well.

modmuss50:
> > Maybe it would be easier to read if there was only 1 message per `RegexFilter`? Im not too bothered either way.
> 
> I agree too but defining the filters like
> 
> ```
> <Filters>
>   <RegexFilter regex="^Failed to verify authentication$" onMatch="DENY" onMismatch="ACCEPT"/>
>   <RegexFilter regex="^Failed to fetch user properties$" onMatch="DENY" onMismatch="ACCEPT"/>
>   <RegexFilter regex="^Couldn't connect to realms$" onMatch="DENY" onMismatch="ACCEPT"/>
> </Filters>
> ```
> 
> doesn't seem to work as expected. It filters using the 1st regex and skips the other two. I couldn't find anything explaining this behavior online either. I tried different regexes as well.

Ah, thats annoying. Lets just do it as one line then, its not the end of the world.

dicedpixels:
> Ah, thats annoying. Lets just do it as one line then, its not the end of the world.

PR updated with @justanothercorpusguy's suggestion and it works. 

`onMismatch="ACCEPT"` indeed stops further processing. `NEUTRAL` does not.

https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/Filter.Result.html#NEUTRAL

modmuss50 (approved):
Thanks :)

PR #1003, opened 2023-12-15
Exclude all *.existing.json files from jar

JT122406:
Kinda a extension of excluding .cache files from the jar but excludes the .existing.json files that some use for lang datagen from the jar as they cannot be used by the game


modmuss50:
> that some use for lang datagen

Is this something that Minecraft generates? Or is it just something that people use in their own mods. If the latter I dont think it should be in loom, its trivial to add to your own mod if needed. I also think its possible that people will have json files named like this that people will want to keep.

PR #1005, opened 2023-12-16
Allow disabling RunConfig appending project path

MattSturgeon:
I'd like to have more control over run config names in sub-projects, for example something like:

```gradle
allprojects {
  loom {
    runs {
      client {
        configName "Run ${project.path.replaceAll ':', ' '}"
        appendProjectPathToConfigName false
      }
      remove server
    }
  }
}
```

This PR adds that capability.

All tests passed when running `:check` locally.

I'd like to have introduced a test for both the default & modified behavior, however I'm a bit lost in your test suite and didn't want to spend much time on this PR without knowing how it'd be received.

If someone more knowledgeable would like to write a test or guide me through the process that'd be appreciated.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/FabricApiExtension.java
+++ b/src/main/java/net/fabricmc/loom/configuration/FabricApiExtension.java
@@ -28,7 +28,9 @@ import java.io.File;
 import java.io.UncheckedIOException;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


modmuss50 (approved):
This is fine for now, its not great but the only way to do it without breaking.

RunConfigSettings is something I'd like to make breaking changes to anyway if/when we do a loom 2.0.

PR #1018, opened 2024-01-05
Cache FMJs and ArtifactMetadata when reading during configuration time

jpenilla:
We have to read every jar on remapped configs to see if they are a mod or should otherwise be remapped. By caching we can avoid re-reading jars that are java-runtime & java-api or are on multiple remapped configs. The cache scope could probably be widened to help more with multi-project builds, but I am leaving that for future work as I think it would also require some sort of invalidation.


modmuss50 (approved):
Looks great, nice spot 👍This is likely not worth doing across projects, cross project stuff in Gradle should be a last resort.

PR #1026, opened 2024-01-16
Support merging pre 1.3 Minecraft versions.

SpaceWalkerRS:
- ~~Added two methods in `LoomGradleExtension` to query whether the obfuscated jars can be merged (determined by querying the MC version from the project dependency).~~
- Added `clientOfficial` and `serverOfficial` mapping namespaces.
- ~~Added "separate jars" Minecraft provider, intermediary Minecraft provider, and named Minecraft provider.~~
  - ~~The intermediary and named Minecraft providers merge the jars after remapping them separately.~~

* Edit: I rended up refactoring most of this. Adding this feature as a new jar configuration lead to some problems, but I was able to merge it into the other jar configurations quite seamlessly. This way if the `MERGED` configuration is used, Loom will figure out itself whether it should merge and then remap or remap first and then merge.


src/main/java/net/fabricmc/loom/api/mappings/layered/MappingsNamespace.java
```diff
@@ -60,6 +70,8 @@ public enum MappingsNamespace {
 	public static @Nullable MappingsNamespace of(String namespace) {
 		return switch (namespace) {
 		case "official" -> OFFICIAL;
+		case "client" -> CLIENT;
+		case "server" -> SERVER;
```
modmuss50: I think this should be `officialClient` and `officialServer`

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftJarConfiguration.java
```diff
@@ -49,6 +49,13 @@ public enum MinecraftJarConfiguration {
 		SingleJarDecompileConfiguration::new,
 		List.of("client", "server")
 	),
+	SEPARATED(
+		SeparateJarsMinecraftProvider::new,
+		IntermediaryMinecraftProvider.SeparatedImpl::new,
+		ProcessedNamedMinecraftProvider.SeparatedImpl::new,
+		SingleJarDecompileConfiguration::new,
+		List.of("client", "server")
+	),
```
modmuss50: Would `LEGACY_MERGED` be a better name for all of this?

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionImpl.java
```diff
@@ -129,6 +131,24 @@ public void setMinecraftProvider(MinecraftProvider minecraftProvider) {
 		this.minecraftProvider = minecraftProvider;
 	}
 
+	@Override
+	public boolean canMergeObfuscatedJars() {
+		if (canMergeObfuscatedJars == null) {
+			canMergeObfuscatedJars = getMinecraftProvider().getVersionInfo().isVersionOrNewer("2012-07-25T22:00:00+00:00" /* 1.3 release date */);
+		}
```
modmuss50: Just use a Provider, it handles this for you.

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionImpl.java
```diff
@@ -129,6 +131,24 @@ public void setMinecraftProvider(MinecraftProvider minecraftProvider) {
...
+	}
+
+	@Override
+	public boolean canMergeObfuscatedJarsOrThrow() {
+		if (!canMergeObfuscatedJars()) {
+			throw new UnsupportedOperationException("The obfuscated client and server jars for Minecraft versions prior to 1.3 cannot be merged. Please provide a common intermediary or use `loom { server/clientOnlyMinecraftJar() }`");
+		}
+
+		return true;
+	}
```
modmuss50: This doesnt need to be part of the extension.

modmuss50 (commented):
Just a very rough high level review, im going to have to sit down and get familar with this stuff again.

Have you got any ideas on how this can be tested? I guess we will need to include some mappings in the tests, might be best to discuss this on discord.

src/main/java/net/fabricmc/loom/api/mappings/layered/MappingsNamespace.java
SpaceWalkerRS: I used `client` and `server` for the sake of compatibility with existing Babric intermediaries. Would that be worth it or should I change it anyway?

modmuss50: Change it, you can trivially rename those when you make your IntermediateMappingsProvider.

SpaceWalkerRS:
I think the mappings merger is broken, but I'm not familiar enough with MIO to fix it.

thecatcore:
> I think the mappings merger is broken, but I'm not familiar enough with MIO to fix it.

This might be because in babric intermediary, if an entry is client or server only, then there is no value for the other side namespace, which might be evaluated as null in MIO if I remember correctly.

SpaceWalkerRS:
IF I understand it right, MIO will need an update to support the double source namespace, both for reading the intermediary and for reading/writing the merged mapping file.

For classes, you could conceivable hack it together by just putting the second source namespace in the first target namespace slot, but this won't work for fields and methods, since you need both the source name and source descriptor.

SpaceWalkerRS:
On second thought, wouldn't updating the tiny format like that mean TR needs an update as well as MIO? I'm starting to think it's not worth it trying to merge the mappings into a single file and instead just making some more modifications to Loom to have separate client and server side intermediary and merged (in this case meaning it has both the intermediary and named target namespaces) mapping files.

Juuxel:
Edit: wrong discussion

thecatcore:
Hm, I was wondering, how would loader work with this without any changes currently? Is there a way to make it not care about official/obfuscated namespace when in dev environement or is it already the case?

modmuss50:
> Hm, I was wondering, how would loader work with this without any changes currently? Is there a way to make it not care about official/obfuscated namespace when in dev environement or is it already the case?

I believe thats already the case, its worth checking though.

SpaceWalkerRS:
The two tests for legacy projects failed - I pushed a fix for the no op intermediate mappings provider , hopefully that resolves that.

There are no tests for 'legacy merged' projects yet, though, but I have no clue how to write those tests.

SpaceWalkerRS:
Most issues are now resolved. The only failing tests are those for <1.3 merged projects, since they require a merged intermediary to work.

src/main/java/net/fabricmc/loom/api/mappings/layered/MappingsNamespace.java
```diff
@@ -37,6 +37,16 @@ public enum MappingsNamespace {
 	 */
 	OFFICIAL,
 
+	/**
+	 * Official names for the Minecraft client jar, usually obfuscated.
```
modmuss50, line 41: Nit: Maybe suggest this is only used pre 1.3, when the client/server had differing names.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
```diff
@@ -29,19 +29,31 @@
...
 import net.fabricmc.loom.api.mappings.intermediate.IntermediateMappingsProvider;
 
 /**
  * A bit of a hack, creates an empty intermediary mapping file to be used for mc versions without any intermediate mappings.
  */
 public abstract class NoOpIntermediateMappingsProvider extends IntermediateMappingsProvider {
 	private static final String HEADER = "tiny\t2\t0\tofficial\tintermediary";
+	private static final String HEADER_LEGACY = "tiny\t2\t0\tintermediary\tclientOfficial\tserverOfficial\t";
+
+	public Property<Boolean> legacy;
```
modmuss50: As this class is already created via the object factory you can just do:

`public abstract Property<Boolean> getIsLegacy()`

This removes the need for you pass the Project through. You can likely remove the `configure` method, and set the convention for this in the extension where its created.

The `finalizeValueOnRead` can be set in a no args constructor marked with "@Inject"

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
```diff
@@ -29,19 +29,31 @@
...
 
+import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.api.mappings.intermediate.IntermediateMappingsProvider;
 
 /**
  * A bit of a hack, creates an empty intermediary mapping file to be used for mc versions without any intermediate mappings.
  */
 public abstract class NoOpIntermediateMappingsProvider extends IntermediateMappingsProvider {
 	private static final String HEADER = "tiny\t2\t0\tofficial\tintermediary";
+	private static final String HEADER_LEGACY = "tiny\t2\t0\tintermediary\tclientOfficial\tserverOfficial\t";
```
modmuss50: Is Legacy the best name for this? Im not sure, I wonder if there is a better word to denote that the client/server had differing obf names. Legacy can mean a lot, and within loom its mostly used for anything pre 1.14.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/TinyMappingsService.java
```diff
@@ -43,13 +48,37 @@ public TinyMappingsService(Path tinyMappings) {
...
+
+			try {
+				mappingTree.accept(new MappingSourceNsSwitch(tree, srcNs));
+			} catch (IOException e) {
+				throw new UncheckedIOException("Failed to switch source namespace", e);
+			}
+
+			mappingTree = tree;
+			lastSrcNs = srcNs;
+		}
```
modmuss50: This feels wrong, I think you should have seperate instances of this service for each namespace. `getOrCreateService` takes a key, that should likely include the namespace.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
```diff
@@ -70,8 +81,27 @@ public static void mergeAndSaveMappings(Path from, Path out, IntermediateMapping
 		try (var writer = new Tiny2FileWriter(Files.newBufferedWriter(out, StandardCharsets.UTF_8), false)) {
 			officialTree.accept(writer);
 		}
+	}
 
-		LOGGER.info(":merged mappings in " + stopwatch.stop());
+	private static void legacyMergeAndSaveMappings(Path from, Path out, IntermediateMappingsService intermediateMappingsService) throws IOException {
+		MemoryMappingTree intermediaryTree = new MemoryMappingTree();
+		intermediateMappingsService.getMemoryMappingTree().accept(intermediaryTree);
```
modmuss50, line 88: Question: Whats the intended source namespace for this tree? In the normal path `MappingSourceNsSwitch` is used to ensure that intermediary is used.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
```diff
@@ -58,9 +59,14 @@ public abstract class MinecraftProvider {
...
 
 	private final Project project;
+	private final Property<Boolean> canMergeJars;
 
 	public MinecraftProvider(ConfigContext configContext) {
 		this.project = configContext.project();
+		this.canMergeJars = this.project.getObjects().property(Boolean.class).convention(this.project.provider(() -> {
+			return getVersionInfo().isVersionOrNewer("2012-07-25T22:00:00+00:00" /* 1.3 release date */);
+		}));
+		this.canMergeJars.finalizeValueOnRead();
```
modmuss50: Its likely cleaner to move this into provide(), likely after downloading the jars. Maybe just have a nullable `Boolean` and throw if its null in canMergeJars(), there is no need to use a Gradle `Property` here.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/SplitMinecraftProvider.java
```diff
@@ -56,6 +56,10 @@ public List<Path> getMinecraftJars() {
 	public void provide() throws Exception {
 		super.provide();
 
+		if (!canMergeJars()) {
+			throw new UnsupportedOperationException("the split jar configuration is not supported at this moment for minecraft versions where the obfuscated client and server jars cannot be merged!");
+		}
+
```
modmuss50: Although not wrong I dont think this is needed, as it will bail out where it checks if `serverBundleMetadata` is not null. This was only added in Minecraft 1.17.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/MappedMinecraftProvider.java
```diff
@@ -42,6 +42,21 @@ interface ProviderImpl extends MappedMinecraftProvider {
...
+		 * available here as well, as an intermediate step.
+		 */
+
+		default MinecraftJar getClientJar() {
+			return new MinecraftJar.Client(getJar(MinecraftJar.Type.CLIENT));
+		}
+
+		default MinecraftJar getServerJar() {
+			return new MinecraftJar.Server(getJar(MinecraftJar.Type.SERVER));
+		}
```
modmuss50: Im having 2nd thoughts on if its a good idea or not to use the same provider for this. The idea was for it to be clear what jars are going to be present, in this case its not that clear. Let me know if you want me to have a look at this.

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -86,8 +86,41 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
...
+
+                    modImplementation "net.fabricmc:fabric-loader:0.12.12"
+                }
+			"""
+
+		when:
+		def result = gradle.run(task: "configureClientLaunch")
+
+		then:
+		result.task(":configureClientLaunch").outcome == SUCCESS
```
modmuss50: We really need to test something more substanial here, a `build` at least with a small mod that compiles against both common and client classes.. It will likely need its own test project + file. Im not against adding some open source mappings from another project into looms repo. 

The tests above are mostly to ensure that the client assets are downloaded correctly, and do not actually test the jars created.

modmuss50 (commented):
This looks great, well done for diving head first into the deepest depths of loom 👍

The main thing is to improve the tests here, as this stuff wont be manually tested and will be easy to break. Let me know if you have any questions, im back looking at loom so should be able to keep up some more.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
SpaceWalkerRS: > As this class is already created via the object factory you can just do:
>
> `public abstract Property<Boolean> getIsLegacy()`
>
> This removes the need for you pass the Project through. You can likely remove the configure method, and set the convention for this in the extension where its created.

I had some trouble with that. I think that is what I tried first (because the other properties in the super class worked that way) but it wouldn't compile because of an abstract method error.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
SpaceWalkerRS: perhaps "obf merged" (for >=1.3) and "obf split" (for <1.3) for these sorts of things?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
SpaceWalkerRS: This assumes the intermediate mappings for <1.3 will be provided as (intermediary) -> [clientOfficial, serverOfficial]. This is done because intermediary is the only complete namespace. The other two have 'holes' which is problematic for field and method descriptors.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/SplitMinecraftProvider.java
SpaceWalkerRS: ah that's my bad, I'll remove this then!

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
SpaceWalkerRS: I'll be honest, I have no clue how to write proper tests for this. I saw the legacy projects tests and just copied the ancient minecraft one and changed it up a little. Perhaps it's best I revert my changes to this file and ask you or someone else who knows what they're doing to write the tests.

I can provide some mappings for those though, Ornithe's Stitch fork should have everything in place now to generate those.

modmuss50: Im quite happy to either help you write them, or write it myself. The tests arent that scary, its just like running a normal gradle project, you just need to set it up, and then validate that the output is as you expect.

For this I think it would be nice to use some real mappings, (they could just be included in the loom repo as a zip file). And then have a very simple mod that just compiles against some clases.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
modmuss50: That makes sense, it might be worth checking that the source ns of the MemoryMappingTree is intermediary after redaing in the mappings.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
modmuss50: `NoOpIntermediateMappingsProvider` has no child classes, so im not sure what issue you had. Maybe try again and let me know if you get stuck again?

SpaceWalkerRS: Alright. I'll push that and see if it works - it's very possible I was doing something wrong in my previous attempt.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
SpaceWalkerRS: I'll move it to just after the version info has been downloaded - that's the earliest it could be done

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/MappedMinecraftProvider.java
SpaceWalkerRS: I'd appreciate it if you looked into it. The reason I did it this way is because these methods are used in all the impls of these providers, e.g. in `NamedMinecraftProvider`:
```java
@Override
public List<RemappedJars> getRemappedJars() {
	return minecraftProvider.canMergeJars()
		? List.of(
			new RemappedJars(minecraftProvider.getMergedJar(), getMergedJar(), MappingsNamespace.OFFICIAL)
		) : List.of(
			new RemappedJars(minecraftProvider.getClientJar(), getClientJar(), MappingsNamespace.CLIENT_OFFICIAL),
			new RemappedJars(minecraftProvider.getServerJar(), getServerJar(), MappingsNamespace.SERVER_OFFICIAL)
		);
}
```
I figured it would make more sense to have these 'mapped client' and 'mapped server' jars in the mapped minecraft providers as opposed to the base minecraft provider.

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
thecatcore: A modified version of babric intermediary could be used to test this.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/MappedMinecraftProvider.java
SpaceWalkerRS: I tried moving these methods out of `ProviderImpl` and into the specific `MergedImpl`s where they're used. Is that a better solution?

src/main/java/net/fabricmc/loom/api/processor/ProcessorContext.java
```diff
@@ -41,5 +41,5 @@ public interface ProcessorContext {
 
 	LazyCloseable<TinyRemapper> createRemapper(MappingsNamespace from, MappingsNamespace to);
 
-	MemoryMappingTree getMappings();
+	MemoryMappingTree getMappings(String srcNs);
```
modmuss50: Why do you need to specity the source ns everwhere? I woudl generally expect it to be intermediary, is this not the case?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediateMappingsService.java
```diff
@@ -80,7 +82,11 @@ public static IntermediateMappingsService create(IntermediateMappingsProvider in
 			throw new UncheckedIOException("Failed to provide intermediate mappings", e);
 		}
 
-		return new IntermediateMappingsService(intermediaryTiny);
+		String expectedSrcNs = minecraftProvider.canMergeJars()
+				? MappingsNamespace.OFFICIAL.toString() // >=1.3
+				: MappingsNamespace.INTERMEDIARY.toString(); // <1.3
+
+		return new IntermediateMappingsService(intermediaryTiny, expectedSrcNs);
```
modmuss50: This doesnt seem right? Wouldnt you always expect IntermediateMappingsService to have the intermediary names as the source NS?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
```diff
@@ -70,8 +81,27 @@ public static void mergeAndSaveMappings(Path from, Path out, IntermediateMapping
...
+
+		try (BufferedReader reader = Files.newBufferedReader(from, StandardCharsets.UTF_8)) {
+			Tiny2FileReader.read(reader, intermediaryTree);
+		}
+
+		MemoryMappingTree officialTree = new MemoryMappingTree();
+		MappingNsCompleter nsCompleter = new MappingNsCompleter(officialTree, Map.of(MappingsNamespace.CLIENT_OFFICIAL.toString(), MappingsNamespace.INTERMEDIARY.toString(), MappingsNamespace.SERVER_OFFICIAL.toString(), MappingsNamespace.INTERMEDIARY.toString()));
+		intermediaryTree.accept(nsCompleter);
```
modmuss50, line 96: Nit, why use the 2nd MemoryMappingTree, can this not go stright to the writer?

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MergedMinecraftProvider.java
```diff
@@ -48,50 +48,73 @@ protected void initFiles() {
 
 	@Override
 	public List<Path> getMinecraftJars() {
-		return List.of(minecraftMergedJar);
+		return canMergeJars() ? List.of(minecraftMergedJar) : List.of(getClientJar(), getServerJar());
```
modmuss50: This doesnt feel right, you would always expect `MergedMinecraftProvider` to have a single jar.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MergedMinecraftProvider.java
```diff
@@ -48,50 +48,73 @@ protected void initFiles() {
...
+		return getMinecraftServerJar().toPath();
+	}
+
 	public Path getMergedJar() {
+		if (!canMergeJars()) {
+			throw new UnsupportedOperationException("the obfuscated client and server jars cannot be merged in this minecraft version");
+		}
+
 		return minecraftMergedJar;
 	}
```
modmuss50: See above, this is goes against how this was designed. WHen using `MergedMinecraftProvider` you would always expect to get a single merged jar.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/IntermediaryMinecraftProvider.java
```diff
@@ -58,9 +59,37 @@ public MergedImpl(Project project, MergedMinecraftProvider minecraftProvider) {
 
 		@Override
 		public List<RemappedJars> getRemappedJars() {
-			return List.of(
-				new RemappedJars(minecraftProvider.getMergedJar(), getMergedJar(), MappingsNamespace.OFFICIAL)
-			);
+			return minecraftProvider.canMergeJars()
+				? List.of(
```
modmuss50: Nit: use an if, its easier to read.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/IntermediaryMinecraftProvider.java
```diff
@@ -58,9 +59,37 @@ public MergedImpl(Project project, MergedMinecraftProvider minecraftProvider) {
...
+				);
+		}
+
+		@Override
+		public List<MinecraftJar> provide(ProvideContext context) throws Exception {
+			List<MinecraftJar> jars = super.provide(context);
+
+			if (minecraftProvider.canMergeJars()) {
+				return jars;
+			}
```
modmuss50: Again, this feels wrong as you are in IntermediaryMinecraftProvider.Merged but getting two none merged jars. Using two of the eixsting SingleJar's should achieve the intended result?

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/NamedMinecraftProvider.java
```diff
@@ -59,15 +59,43 @@ public MergedImpl(Project project, MergedMinecraftProvider minecraftProvider) {
...
+		// in mc versions prior to 1.3, the client and server jars
+		// are remapped separately before being merged
+
+		private MinecraftJar getClientJar() {
+			return new MinecraftJar.Client(getJar(MinecraftJar.Type.CLIENT));
+		}
+
+		private MinecraftJar getServerJar() {
+			return new MinecraftJar.Server(getJar(MinecraftJar.Type.SERVER));
+		}
```
modmuss50: The named jar is the one you compile against, why are they split here?

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -86,8 +86,41 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
 		'1.3.2'			| _
 	}
 
+	// TODO: need intermediaries for this test to succeed
```
modmuss50: This needs doing.

modmuss50 (commented):
This still feels a bit hacked on tbh, it is made a little harder by the fact that the existing stuff was never designed with this in mind.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediateMappingsService.java
SpaceWalkerRS: The IntermediaryMappingsProvider extracts the mappings out of the jar as is, so the IntermediateMappingsService has the mappings as `official -> intermediary`. A custom intermediary provider for <1.3 would have to provide them as `intermediary -> [officialClient, officialServer]`.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MergedMinecraftProvider.java
SpaceWalkerRS: That would mean adding a new MinecraftJarConfiguration, right? I'll have to make another attempt at that, the first time I tried it it seemed like I'd have to rewrite a bunch of stuff for it. But it's been a while so I can't remember the specific problems I ran into.

modmuss50: Hold off for a bit, im cleaning some stuff up to help with this. 👍

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/IntermediaryMinecraftProvider.java
SpaceWalkerRS: I'm not sure there is really another way here. The obfuscated jars cannot be merged so you need to do the jar merging *after* the intermediary mapped jars have been provided. Even if I added new mapped minecraft provider impls for this, I'd still need to do the merging here.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MergedMinecraftProvider.java
SpaceWalkerRS: Ah I remember the issue I ran into when creating a new jar configuration for this. The jar configuration is used to create the minecraft provider, which in turn retrieves and parses the minecraft info, which is used to check if a version can be merged or not. Basically, it would force the user to manually choose which 'kind' of merged they need while there is no reason loom couldn't figure that out for itself. But for that the minecraft info needs to be parsed before the jar configuration is created.

modmuss50: First thing I have done is remove that mess of generics: https://github.com/FabricMC/fabric-loom/pull/1070 makes things a lot easier :D (I hope)

I have then created a rough outline of how I think this could work here: https://github.com/modmuss50/fabric-loom-1/commit/32df745660b70803c81a4003aa786eecd2ad2181

I have created some new minecraft and intermediary providers that delgate to the existing single jar providers for the client and server. This basically means you end up with both the client and server jar as if you were using `clientOnlyMinecraftJar` and `serverOnlyMinecraftJar`. You end with with 2 seperate named client/server jars, the special handling to merge them can then happen in `NamedMinecraftProvider.LegacyMergedImpl`

Let me know what you think, I havent tried to run it partly becuase I dont have mappings to test with, and partly becuase you might notice something tottaly wrong with it 😆 

>  it would force the user to manually choose which 'kind' of merged they need while there is no reason loom couldn't figure that out for itself.

This is a valid issue, and still a problem with what I have done above. Ill look into seeing what could be done about that. It likely requires moving the `MinecraftMetadataProvider` out of the `MinecraftProvider`, I think this can be handled separately. Its not the end of the world for it to be specified for now.

SpaceWalkerRS: > First thing I have done is remove that mess of generics: #1070 makes things a lot easier :D (I hope)

Neat, that should make it less of a head ache for sure!

> Let me know what you think, I havent tried to run it partly becuase I dont have mappings to test with, and partly becuase you might notice something tottaly wrong with it 😆

I do have some questions about it, mostly 'cause I'm not familiar enough with Loom's inner workings.

It is not an issue that the intermediary and named providers provide separate client and server jars instead of one merged jar? Or are these jars otherwise never exposed and used anywhere else? If so, could the merging then be done at the start of the processed named provider step?

And something that feels more like a nitpick - is there a reason behind the use of delegate providers rather than passing in the client and server jars as defined in `MinecraftProvider`?

> This is a valid issue, and still a problem with what I have done above. Ill look into seeing what could be done about that. It likely requires moving the `MinecraftMetadataProvider` out of the `MinecraftProvider`, I think this can be handled separately. Its not the end of the world for it to be specified for now.

Fair enough, I'll look out for that then!

modmuss50: > It is not an issue that the intermediary and named providers provide separate client and server jars instead of one merged jar?

Yes, I think that might be an issue. I forgot that you can still create a merged intermediary jar, thats easily solved though and would follow the same pattern as the named provider.

> And something that feels more like a nitpick - is there a reason behind the use of delegate providers rather than passing in the client and server jars as defined in MinecraftProvider?

Good point, that is a waste of time, you could use the same provider.

src/main/java/net/fabricmc/loom/api/processor/ProcessorContext.java
SpaceWalkerRS: It was added because the impl calls `MappingConfiguration.getMappingsService` which now asks for the source namespace. But you're right, it's not needed as a parameter here, and can just be hardcoded to intermediary in `ProcessorContextImpl`.

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
SpaceWalkerRS: If I have a mapping file in `/source/test/recources/`, what would be the way to go about writing the test so that it uses it?

SpaceWalkerRS:
Alright I tried implementing it how you suggested, it wasn't too bad to change it up. I hope I didn't miss any changes that could be reverted but there we are. 

I also tried adding some mappings for the legacy project test, but I'm not sure how to go about making the test use those. Any help with that would be appreciated.

this PR now depends on #1070

modmuss50:
Please see: https://github.com/FabricMC/fabric-loom/pull/1072

This should allow you to adjust the default jar configuration used based on minecraft versions.

SpaceWalkerRS:
> Please see: #1072
> 
> This should allow you to adjust the default jar configuration used based on minecraft versions.

That was quick, thanks so much! It should be a small change to this PR, so I'll wait until that is merged in.

src/test/resources/mappings/1.2.5-intermediary.tiny
```diff
@@ -0,0 +1,8905 @@
+v1	intermediary	officialClient	officialServer
```
modmuss50: Can you add a small file next to this explaining where it came from and its license?

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -102,7 +131,7 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
 						// No names
 					}
 
-                    modImplementation "net.fabricmc:fabric-loader:0.12.12"
+                    modImplementation "net.fabricmc:fabric-loader:0.15.7" // 0.15.2 fixed MC 1.2.5 not launching
```
modmuss50: Nit: no need for this comment, just update it.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/LegacyMinecraftProvider.java
```diff
@@ -0,0 +1,75 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+import net.fabricmc.loom.configuration.ConfigContext;
+
+/**
+ * Minecraft versions prior to 1.3 obfuscate the server and client jars differently.
+ * The obfuscated jars must be provided separately, and can be merged after remapping.
+ */
+public final class LegacyMinecraftProvider extends MinecraftProvider {
```
modmuss50: Didnt you suggest that this can just be the regular MergedMinecraftProvider?

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -87,12 +87,41 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
...
+		'1.2.5'			| _
+	}
+
+	@Unroll
+	def "Ancient minecraft client-only(minecraft #version)"() {
 		setup:
 		def gradle = gradleProject(project: "minimalBase", version: PRE_RELEASE_GRADLE)
 		gradle.buildGradle << """
 				loom {
-                    noIntermediateMappings()
```
modmuss50: How does this work? There are other versions here.

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -87,12 +87,41 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
 	}
 
 	@Unroll
-	def "Ancient minecraft (minecraft #version)"() {
+	def "Ancient minecraft merged (minecraft #version)"() {
+		setup:
+		def gradle = gradleProject(project: "minimalBase", version: PRE_RELEASE_GRADLE)
+		gradle.buildGradle << """
+				loom {
```
modmuss50: How does this work? I dont see how the 1.2.5-intermediary.tiny is used?

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -87,12 +87,41 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
...
+                    legacyMergedMinecraftJar()
+                }
+
+                dependencies {
+                    minecraft "com.mojang:minecraft:${version}"
+                    mappings loom.layered() {
+						// No names
+					}
+
+                    modImplementation "net.fabricmc:fabric-loader:0.15.7" // 0.15.2 fixed MC 1.2.5 not launching
```
modmuss50: Nit: comment not needed

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -87,12 +87,41 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
...
+                }
+			"""
+
+		when:
+		def result = gradle.run(task: "configureClientLaunch")
+
+		then:
+		result.task(":configureClientLaunch").outcome == SUCCESS
```
modmuss50: I think we need to test more than just configuring the client launch. We should be able to build a jar that compliles against mc code. Maybe even try running a server with this? (might be easier said than done though)

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
```diff
@@ -37,11 +40,19 @@
...
-	private static final String HEADER = "tiny\t2\t0\tofficial\tintermediary";
+	private static final String HEADER_OFFICIAL_MERGED = "tiny\t2\t0\tofficial\tintermediary";
+	private static final String HEADER_OFFICIAL_SPLIT = "tiny\t2\t0\tintermediary\tclientOfficial\tserverOfficial\t";
+
+	@Inject
+	public NoOpIntermediateMappingsProvider() {
+		getIsOfficialSplit().finalizeValueOnRead();
+	}
+
+	public abstract Property<Boolean> getIsOfficialSplit();
```
modmuss50: It might be better to include the minecraft release time in `IntermediateMappingsProvider` so other providers can decide to use the new namespaces?

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -155,7 +156,11 @@ private synchronized void setupMinecraft(ConfigContext configContext) throws Exc
 
 		final MinecraftMetadataProvider metadataProvider = MinecraftMetadataProvider.create(configContext);
 
-		final var jarConfiguration = extension.getMinecraftJarConfiguration().get();
+		var jarConfiguration = extension.getMinecraftJarConfiguration().get();
+
+		if (jarConfiguration == MinecraftJarConfiguration.MERGED && !metadataProvider.getVersionMeta().isVersionOrNewer("2012-07-25T22:00:00+00:00" /* 1.3 release date */)) {
```
modmuss50: NIT: Maybe move the release time to `Constants`? Will save duplicating it everywhere.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/IntermediaryMinecraftProvider.java
```diff
@@ -85,24 +125,26 @@ protected void configureRemapper(RemappedJars remappedJars, TinyRemapper.Builder
...
 		}
 
 		public static SingleJarImpl server(Project project, SingleJarMinecraftProvider minecraftProvider) {
-			return new SingleJarImpl(project, minecraftProvider, SingleJarEnvType.SERVER);
+			return new SingleJarImpl(project, minecraftProvider, SingleJarEnvType.SERVER, minecraftProvider.canMergeJars() ? MappingsNamespace.OFFICIAL : MappingsNamespace.SERVER_OFFICIAL);
 		}
 
 		public static SingleJarImpl client(Project project, SingleJarMinecraftProvider minecraftProvider) {
-			return new SingleJarImpl(project, minecraftProvider, SingleJarEnvType.CLIENT);
+			return new SingleJarImpl(project, minecraftProvider, SingleJarEnvType.CLIENT, minecraftProvider.canMergeJars() ? MappingsNamespace.OFFICIAL : MappingsNamespace.CLIENT_OFFICIAL);
```
modmuss50: This shouldnt be needed, as you pass the side specific NS's above.

modmuss50: Oh, I see is this for when you want to single jars 🤔 This is going to be a breaking change, as the ns will now change. Not too sure what to do about this.

modmuss50 (commented):
Looking a lot better now, I think the tests still need improving. Do let me know if you want some help with this 👍

SpaceWalkerRS:
> Looking a lot better now, I think the tests still need improving. Do let me know if you want some help with this 👍

I'd appreciate some help, as I'm stuck with both failing tests.

The mapping reader tests fail because of the intermediary source namespace validation. They all fail with 'expected intermediary but found official' which is wrong since all those tests use MC 1.17 which must have official as the source namespace. This most likely means the `canMergeJars` call fails, which I thought was because the test version meta lacked the releaseTime field. But adding that in did not fix it...

As for the legacy project test, while I did add an intermediary file, I do not know how to go about making the test use it.

modmuss50:
> > Looking a lot better now, I think the tests still need improving. Do let me know if you want some help with this 👍
> 
> I'd appreciate some help, as I'm stuck with both failing tests.

No worries at all, I can take a look at this, might be a day or two as im going to be a little busy the 1/2 half of this week.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/LegacyMinecraftProvider.java
SpaceWalkerRS: Yes, but when I was working on it I found that this way makes the `LegacyMergedImpl`s in the intermediary and named minecraft providers easier, so I decided to keep it your way.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/NoOpIntermediateMappingsProvider.java
SpaceWalkerRS: At the moment it's pretty much tied to being before or after 1.3, if we want it to be more flexible this'll have to be changed in several other places too. All the mapped minecraft providers, the intermediate mappings service, the mappings merger. I'm not sure how straightforward it'd be but I'll have a look.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/IntermediaryMinecraftProvider.java
SpaceWalkerRS: Perhaps Loom can provide a built in intermediate mappings provider that uses the 'modern style' namespaces for old versions?

modmuss50: On 2nd thoughts im not too botherd about breaking things like this, this old version support has always been somewhat experimental anyway.

src/test/groovy/net/fabricmc/loom/test/integration/LegacyProjectTest.groovy
```diff
@@ -118,4 +120,35 @@ class LegacyProjectTest extends Specification implements GradleProjectTestTrait
...
+                }
+			"""
+		gradle.buildSrc("legacyMergedIntermediary")
+
+		when:
+		def result = gradle.run(task: "build")
+
+		then:
+		result.task(":build").outcome == SUCCESS
+	}
```
modmuss50: I have added this test, that uses a seperate plugin bellow to supply the 1.2.5 intermediaries.

Its currently failing with `invalid new source ns serverOfficial: not in [intermediary, named] or official`, is this something I have done wrong? Have you got this working outside of the automated tests?

SpaceWalkerRS: Thank you! I'll look into the error when I get home, perhaps I might have messed up with the changes to the mappings service?

SpaceWalkerRS: I have found a few issues that were easily resolved (the mappings were provided in Tiny V1 with the wrong namespaces in the header), but I'm now running into mapping conflicts that do not make sense. I think client and server mappings are leaking into the wrong namespaces. From testing it seemed if I drop one of them right after reading the intermediary mappings, the issue goes away for that jar, but for some reason dropping the not-required namespace in the TinyMappingsService does not have the same effect.

modmuss50: > mapping conflicts that do not make sense.
Do you mean when it tries to remap the jar to intermediary or named?

If you want to push what you have so far I can maybe take a look and try to figure out whats going on.

SpaceWalkerRS: thanks, I pushed the fixed mapping file

SpaceWalkerRS: I also tried enabling the `dropMissingNewSrcName` flag in the SourceNsSwitch in the TinyMappingsService but that did not change anything for some reason.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/TinyMappingsService.java
```diff
@@ -31,22 +31,24 @@
...
 import net.fabricmc.loom.util.service.SharedServiceManager;
 import net.fabricmc.mappingio.MappingReader;
+import net.fabricmc.mappingio.adapter.MappingSourceNsSwitch;
 import net.fabricmc.mappingio.tree.MemoryMappingTree;
 
 public final class TinyMappingsService implements SharedService {
 	private final MemoryMappingTree mappingTree;
 
-	public TinyMappingsService(Path tinyMappings) {
+	public TinyMappingsService(Path tinyMappings, String srcNs) {
```
modmuss50: I have removed (Ill push my changes shortly) the srcNs from here as I dont think its needed or ocrrect. The source NS is controlled by the input mapping file, this service is basically a shared way to read the raw mapping file, if another NS is needed it should be handled by the consumer of this service.

modmuss50:
Out of intrest what named mappings are you using? Id love to do a full manual end to end test with this, some intermediary names and then names I can make a small mod against.

modmuss50:
Im happy with this now, would just like to see it working outside of the automated tests. Make sure the game actually runs (both client and server), and that you can apply named mappings.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/TinyMappingsService.java
SpaceWalkerRS: I added that because of [this error](https://github.com/SpaceWalkerRS/fabric-loom/blob/05fbc2d539012e4d9eea59270e3ba291fbe14d51/src/main/java/net/fabricmc/loom/util/TinyRemapperHelper.java#L72), which otherwise causes a crash. Should there be a SourceNsSwitch there instead of an exception then?

SpaceWalkerRS:
> Out of intrest what named mappings are you using? Id love to do a full manual end to end test with this, some intermediary names and then names I can make a small mod against.

It's still early days but eventually [Ornithe's Feather mappings](https://github.com/OrnitheMC/feather-mappings/tree/main) will be ported to this new merged intermediary, and could be used to test this. At the moment though I do not have any named mappings available for these intermediaries.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/TinyMappingsService.java
SpaceWalkerRS: nvm I see you change the fixRecords flag based on the MC provider - but the LegacyMergedMinecraftProvider uses delegate single jar providers, so it doesn't work

modmuss50: Ill take a look, can maybe improve this to not depend on the provider at all.

modmuss50: Now checks the required java version, should be a bit more robust 👍

SpaceWalkerRS: looking more into this, it's basically every place where you replaced `minecraftProvider.canMergeJars()` with `!(minecraftProvider instanceof LegacyMergedMinecraftProvider)` that is broken now, and that's quite a few places.

SpaceWalkerRS: that might be partially 'cause I was using a server only test project, but those should still use the new namespaces and whatnot in pre-1.3 versions

modmuss50: Yeah, I have just noticed the same, I might revert this 🤔

modmuss50:
Pushed some more fixes, no idea how the test was passing on my machine, once I cleared the test caches it started failing. I have a dev env where I have a working server, but the client stalls on startup. I have a feeling this is due to my arm64 mac.

modmuss50:
Also it doesnt seem to be using a merged jar 🤔

modmuss50:
Right, sorry for the commit spam. I think I have resolved most of my mess. I cannot test legacy clients on my arm mac (this might be something to solve later), but it seems the merged server isnt working:

```
java.lang.ClassFormatError: Duplicate interface name "net/minecraft/unmapped/C_2286002" in class file net/minecraft/unmapped/C_4253713
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at net.fabricmc.loader.impl.launch.knot.KnotClassLoader.defineClassFwd(KnotClassLoader.java:160)
	at net.fabricmc.loader.impl.launch.knot.KnotClassDelegate.tryLoadClass(KnotClassDelegate.java:355)
	at net.fabricmc.loader.impl.launch.knot.KnotClassDelegate.loadClass(KnotClassDelegate.java:218)
	at net.fabricmc.loader.impl.launch.knot.KnotClassLoader.loadClass(KnotClassLoader.java:119)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at net.minecraft.unmapped.C_4507027.m_7189836(C_4507027.java:36)
	at net.minecraft.unmapped.C_7366354.m_7447312(C_7366354.java:53)
	at net.minecraft.unmapped.C_3906126.<init>(C_3906126.java:162)
	at net.minecraft.unmapped.C_7366354.<init>(C_7366354.java:26)
	at net.minecraft.server.MinecraftServer.m_2695328(MinecraftServer.java:253)
	at net.minecraft.server.MinecraftServer.m_0615401(MinecraftServer.java:188)
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:335)
	at net.minecraft.unmapped.C_6018901.run(C_6018901.java:492)
```

Screenshot 2024-03-18 at 22 42 09

This appears it might be an issue with the jar merger?

SpaceWalkerRS:
Woo, project setup works for me locally as well now! Thanks so much for the help!

> it seems the merged server isnt working

I think I know the issue, this'll be due to the changes to the MinecraftJarMerger I made, forgot to do an equals check I suppose.

SpaceWalkerRS:
alright, it's not very pretty, but this should fix the duplicate interfaces

SpaceWalkerRS:
I can confirm both the server and client run from the dev environment, at least for the merged jars. Client-only and server-only is broken because the single jar impls no longer use the appropriate source namespace for pre-1.3 versions

modmuss50 (approved):

modmuss50:
> Client-only and server-only is broken because the single jar impls no longer use the appropriate source namespace for pre-1.3 versions

Ill take a quick look to see if this can be fixed.

SpaceWalkerRS:
imo the changes made when removing `canMoveJars` should be reverted - it's not the jar configuration that determines how the mappings are provided, but the MC version

modmuss50:
> imo the changes made when removing `canMoveJars` should be reverted - it's not the jar configuration that determines how the mappings are provided, but the MC version

My thinking behind this change was to remove the idea of merged jars away from the abstract `MinecraftProvider` class, you should still be able to `SingleJarMinecraftProvider` as before (as far as I can tell this is working as it was before). The minecraft version is only used to decide what jar provider to use.

modmuss50:
> Client-only and server-only is broken because the single jar impls no longer use the appropriate source namespace for pre-1.3 versions

This now works as it did before, using `named` as the only namespace. In an `IntermediateMappingsProvider` you can query `getIsLegacyMerged()` to know if the new namespaces should be used or not.

SpaceWalkerRS:
> My thinking behind this change was to remove the idea of merged jars away from the abstract `MinecraftProvider` class, you should still be able to `SingleJarMinecraftProvider` as before (as far as I can tell this is working as it was before). The minecraft version is only used to decide what jar provider to use.

Perhaps the method should be renamed then? The idea behind it is that it returns whether the client and server have common obfuscation maps, i.e. whether they are 'merge-able'. If they are not, mappings must be provided with separate client and server namespaces, no matter which jar configuration is used. 
At the moment it only expects the `clientOfficial` and `serverOfficial` namespaces to be used if the legacy merged jar configuration is selected for a pre-1.3 version, while they should also be used for the client-only or server-only jar configuration for pre-1.3 versions.

SpaceWalkerRS:
> In an `IntermediateMappingsProvider` you can query `getIsLegacyMerged()` to know if the new namespaces should be used or not.

But `isLegacyMerged` is only set to true if the jar configuration is the legacy merged one.

modmuss50:
> > In an `IntermediateMappingsProvider` you can query `getIsLegacyMerged()` to know if the new namespaces should be used or not.
> 
> But `isLegacyMerged` is only set to true if the jar configuration is the legacy merged one.

Yes, its set with:
`provider.getIsLegacyMerged().set(getProject().provider(() -> getMinecraftProvider() instanceof LegacyMergedMinecraftProvider));`

SpaceWalkerRS:
> > But `isLegacyMerged` is only set to true if the jar configuration is the legacy merged one.
> 
> Yes, its set with: `provider.getIsLegacyMerged().set(getProject().provider(() -> getMinecraftProvider() instanceof LegacyMergedMinecraftProvider));`

It should be true for pre-1.3 versions regardless of the jar configuration.

modmuss50:
> It should be true for pre-1.3 versions regardless of the jar configuration.

Im not sure I agree with that, especially if we want to maintain backwards compatibility. You could still use merged mappings with this, by chaning the namespaces in the intermediary provider.

SpaceWalkerRS:
> > It should be true for pre-1.3 versions regardless of the jar configuration.
> 
> Im not sure I agree with that, especially if we want to maintain backwards compatibility. You could still use merged mappings with this, by chaning the namespaces in the intermediary provider.

Providing mappings differently depending on the jar configuration seems like a bad idea to me. It will lead to the same kind of cache issues we had before Loom 1.3 with different intermediary providers.

modmuss50:
> Providing mappings differently depending on the jar configuration seems like a bad idea to me. It will lead to the same kind of cache issues we had before Loom 1.3 with different intermediary providers.

Thats a good point, ill change it back then 👍

src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
```diff
@@ -44,6 +44,14 @@ public abstract class IntermediateMappingsProvider implements Named {
 
 	public abstract Property<Function<String, DownloadBuilder>> getDownloader();
 
+	/**
+	 * Set to true if the minecraft version is pre 1.3, and the client and server are going to be merged.
+	 * When true the expected src namespace is intermediary, and the expected dst namespaces are clientOfficial and serverOfficial
+	 * When false the expected src namespace is named and the expected dst namespace is intermediary
+	 */
+	@ApiStatus.Experimental
+	public abstract Property<Boolean> getIsLegacyMerged();
```
modmuss50: I see you reverted this as well, I think this is  the wrong name as it can still be used for the single jars.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
```diff
@@ -185,6 +186,10 @@ public MinecraftVersionMeta getVersionInfo() {
 		return Objects.requireNonNull(metadataProvider, "Metadata provider not setup").getVersionMeta();
 	}
 
+	public boolean canMergeJars() {
```
modmuss50: Im not too keen on this name, it should be more about the version, and not if it can be merged or not. Right now with this the the split or single providers will return true for `canMergeJars`, this isnt very clear. I think something along the lines of `isLegacyVersion` may be better?

modmuss50:
Sorry, I didnt realise you had made this change as well, I was working on https://github.com/modmuss50/fabric-loom-1/commit/137d9b5f5a43526cadef6b2fbd788665a029c1a7 I think it makes it a bit clearer, and removes the duplicated logic where it decides on the ns to use. If you feel strongly with what you have im happy to go with it.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
SpaceWalkerRS: `isLegacyVersion` or `isLegacyMinecraft` sounds good, perhaps `isPre1_3` though that doesn't leave any room for it to change I suppose.

SpaceWalkerRS:
Ah oops I misread your message - I think your changes make a lot of sense, so let's go ahead with those 👍

modmuss50:
> Ah oops I misread your message - I think your changes make a lot of sense, so let's go ahead with those 👍

Thanks, I have pushed those here. I will merge this once the tests pass. Can of course fix/improve things in later PRs 👍

modmuss50:
Merged, thanks a lot for persisting with this 🎉

PR #1031, opened 2024-01-28
Improve locking strategy for concurrent loom executions

jpenilla:
This does **not** allow building a project concurrently, and that is not a goal of this change. The goal of this change is to prevent the loom cache from being invalidated when IntelliJ syncs the project on its own during a user-requested Gradle run.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -296,7 +296,7 @@ public abstract class CompileConfiguration implements Runnable {
 				return LockResult.ACQUIRED_ALREADY_OWNED;
 			}
 
-			logger.lifecycle("Lock file \"{}\" is currently held by pid '{}'.", lockFile, lockingProcessId);
+			logger.lifecycle("\"{}\" is currently held by pid '{}'.", lockFile, lockingProcessId);
 
 			if (ProcessHandle.of(lockingProcessId).isEmpty()) {
 				logger.lifecycle("Locking process does not exist, assuming abrupt termination and deleting lock file.");
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -31,6 +31,7 @@ import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.function.Consumer;
 
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -235,20 +240,91 @@ private Path getLockFile() {
...
+					}
+
+					sleptMs += 100;
+
+					if (sleptMs >= 1000 * 60 && sleptMs % (1000 * 60) == 0L) {
+						logger.lifecycle(
+								"""
+										Have been waiting on lock file '{}' held by pid '{}' for {} minute(s).
+										If this persists for an unreasonable length of time, kill this process, run './gradlew --stop' and then try again.
+										If the problem persists, the lock file may need to be deleted manually.""",
```
modmuss50: This looks great overall, I'm not super happy about suggesting to the user to delete the lock file. The issue with doing this is loom will contiune thinking everything is ok. In what situation may you need to delete the lockfile?

modmuss50: Maybe the soltuion is to write -1 to the lockfile here, so any future runs will go down the route of the previous owner missing?

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -235,20 +240,91 @@ private Path getLockFile() {
...
-			throw new UncheckedIOException("Failed to acquire getProject() configuration lock", e);
+			return this.acquireProcessLockWaiting_(lockFile, timeoutMs);
+		} catch (final IOException e) {
+			throw new RuntimeException("Exception acquiring lock " + lockFile, e);
+		}
+	}
+
+	// Returns true if our process already owns the lock
+	@SuppressWarnings("BusyWait")
+	private LockResult acquireProcessLockWaiting_(Path lockFile, long timeoutMs) throws IOException {
```
modmuss50: Nit: Take a Duration for the timeout.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -235,20 +240,91 @@ private Path getLockFile() {
...
+				lockingProcessId = Long.parseLong(Files.readString(lockFile));
+			} catch (final Exception e) {
+				lockingProcessId = -1;
+			}
+
+			if (lockingProcessId == currentPid) {
+				return LockResult.ACQUIRED_ALREADY_OWNED;
+			}
+
+			logger.lifecycle("Lock file '{}' is currently held by pid '{}'.", lockFile, lockingProcessId);
```
modmuss50: As commented bellow im a little hesitant to point the user towards the lock file, I worry if they start tampering with it they may end up in more of a mess. Maybe just print the project name that goes into the lock file?

modmuss50 (commented):
This is a great idea! The goal of the lock file was to try and resolve issues for the user, not create more, suggesting to them to manually delete the lock file doesnt help with this imo.

Could possibly do with a unit test. To do this you would likely need to mock out the ProcessHandle usages.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
jpenilla: I think simply changing the message is enough, if we were to write `-1` while a zombie process holds the lock it could cause other problems...

modmuss50: Ok, thats fine. If it does become an issue we could possibly look into getting more info about the other process (such as its command line, or whatever to try and figure out what it is). Not sure how easy that is to do so dont worry about it for this PR.

modmuss50 (approved):
Looks great, this should be a big improvement. 👍

PR #1033, opened 2024-01-31
Test Gradle 8.6 and latest nightly

modmuss50:
PR to just let the tests run.


modmuss50:
All of the tests passed on 8.6-rc4, however the 8.7 nightly is a diffrent story. The tests that did fail were because of:

- Declaring client module dependencies has been deprecated. This is scheduled to be removed in Gradle 9.0. Please use component metadata rules instead. Consult the upgrading guide for further information: https://docs.gradle.org/8.7-20240131001302+0000/userguide/upgrading_version_8.html#declaring_client_module_dependencies

PR #1038, opened 2024-02-02
Prepare for SelfResolvingDependency's removal & Update to Gradle 8.6

### LINKED ISSUE #1022: SelfResolvingDependency slated for deprecation
This doesn't appear to be a a near term thing by any means, but per https://github.com/gradle/gradle/pull/27420, it looks like the Gradle folks are discussing deprecating `SelfResolvingDependency`, which loom appears to use for some of its mappings resolution.

Hopefully this deprecation is still a ways out, but looking ahead will hopefully make it less of a scramble at the last minute to resolve deprecation warnings :)

modmuss50:
Layred mappings no longer use SelfResolvingDependency, and are now created during afterEvalute and placed into the global loom maven. This matches how remapped mods are handled, there may be other ways to handle this.

Any use of SelfResolvingDependency though the API has been supported via reflection, and will break when updating to Gradle 9. The new Gradle Problem API has been used to log a warning if SelfResolvingDependency is used. 

Fixes #1022

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/gradle/SelfResolvingDependencyUtils.java
+++ b/src/main/java/net/fabricmc/loom/util/gradle/SelfResolvingDependencyUtils.java
@@ -1,0 +1,111 @@
+/*
+ * This file is part of fabric-loom, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2024 FabricMC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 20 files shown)


modmuss50:
This is blocked untill Gradle releases 8.6 docker images for the tests.

src/main/java/net/fabricmc/loom/util/gradle/SelfResolvingDependencyUtils.java
```diff
@@ -0,0 +1,111 @@
...
+	private static final Class<?> SELF_RESOLVING_DEPENDENCY_CLASS = getSelfResolvingDependencyOrNull();
+	@Nullable
+	private static final Method RESOLVE_METHOD = getResolveMethod(SELF_RESOLVING_DEPENDENCY_CLASS);
+
+	/**
+	 * @return true when dependency is a SelfResolvingDependency but NOT a FileCollectionDependency.
+	 */
+	public static boolean isExplicitSRD(Dependency dependency) {
+		// FileCollectionDependency is usually the replacement for SelfResolvingDependency
+		if (dependency instanceof FileCollectionDependency) {
```
modmuss50, line 55: ProjectDependency is also a SelfResolvingDependency, need to handle this here and elsewhere.

modmuss50: Doesnt need to be handled elsewhere.

PR #1040, opened 2024-02-04
Print info about process lock owner

jpenilla:
image


src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -344,6 +351,69 @@ private LockResult acquireProcessLockWaiting_(LockFile lockFile, Duration timeou
 		return abrupt ? LockResult.ACQUIRED_PREVIOUS_OWNER_MISSING : LockResult.ACQUIRED_CLEAN;
 	}
 
+	private String printWithParents(ProcessHandle processHandle) {
+		StringBuilder output = new StringBuilder();
+
+		var chain = getParentChain(null, processHandle);
```
modmuss50: Nit: id use var for `var output = new StringBuilder();` as its clear what the type is, but not when calling a method.

modmuss50 (approved):

PR #1042, opened 2024-02-04
Download intermediary using Gradle

jpenilla:
This gives us free caching and improved usability under offline mode.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
@@ -29,6 +29,7 @@ import java.nio.file.Path;
 import java.util.function.Function;
 
 import org.gradle.api.Named;
+import org.gradle.api.Project;
 import org.gradle.api.provider.Property;
 import org.jetbrains.annotations.ApiStatus;
 
@@ -49,4 +50,21 @@ public abstract class IntermediateMappingsProvider implements Named {
 	 * @throws IOException
 	 */
 	public abstract void provide(Path tinyMappings) throws IOException;
+
+	public final void provideWithContext(Path tinyMappings, Project project) throws IOException {
+		if (this instanceof Contextual ctx) {
+			ctx.provide(tinyMappings, project);
+		} else {
+			this.provide(tinyMappings);
+		}
+	}
+
+	public abstract static class Contextual extends IntermediateMappingsProvider {
+		@Override
+		public final void provide(Path tinyMappings) throws IOException {
+			throw new UnsupportedOperationException();
+		}
+
+		public abstract void provide(Path tinyMappings, Project project) throws IOException;
+	}
 }
--- a/src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
@@ -29,7 +29,6 @@ import java.nio.file.Path;
 import java.util.function.Function;
 
 import org.gradle.api.Named;
-import org.gradle.api.Project;
 import org.gradle.api.provider.Property;
 import org.jetbrains.annotations.ApiStatus;
 
@@ -50,21 +49,4 @@ public abstract class IntermediateMappingsProvider implements Named {
 	 * @throws IOException
 	 */
 	public abstract void provide(Path tinyMappings) throws IOException;
-
-	public final void provideWithContext(Path tinyMappings, Project project) throws IOException {
-		if (this instanceof Contextual ctx) {
-			ctx.provide(tinyMappings, project);
-		} else {
-			this.provide(tinyMappings);
-		}
-	}
-
-	public abstract static class Contextual extends IntermediateMappingsProvider {
-		@Override
-		public final void provide(Path tinyMappings) throws IOException {
-			throw new UnsupportedOperationException();
-		}
-
-		public abstract void provide(Path tinyMappings, Project project) throws IOException;
-	}
 }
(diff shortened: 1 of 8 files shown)


modmuss50:
It seems all the test are broken :)

Whats the reasoning behind this change? Looms own downloader should be able to cache this just fine and handle offline mode. I do want to try and add an intergration test to provide offline works, I just dont have a good way to run loom without a working internet connection in CI.

jpenilla:
The downloader's caching isn't being used as the destination is a temporary directory.
In cases where we can't use Gradle's dependency resolution, the downloader's caching is great, i.e. Mojang's `server/client.txt` mappings. But for things that are available as artifacts on the Maven, I think it's better to use Gradle's built-in resolution. This way users can control the cache times, etc. through the standard Gradle mechanisms, and the cached artifacts can survive a loom cache rebuild.

modmuss50:
> The downloader's caching isn't being used as the destination is a temporary directory.

Ah ok, yes thats the real issue here. Ideally everything that is downloaded via looms downloader wont be reset during a cache rebuild. Ill see if I can add a test for this in another PR.

At somepoint (likely loom 2.0) I want to remove `getIntermediaryUrl` from the extension, and replace it fully with the newer `IntermediateMappingsProvider` API, would save needing to have these two paths. For now what you have is ok.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -45,24 +55,59 @@ public abstract class IntermediaryMappingsProvider extends IntermediateMappingsP
...
 
 		// Download and extract intermediary
 		final Path intermediaryJarPath = Files.createTempFile(getName(), ".jar");
 		final String encodedMcVersion = UrlEscapers.urlFragmentEscaper().escape(getMinecraftVersion().get());
-		final String url = getIntermediaryUrl().get().formatted(encodedMcVersion);
+		final String urlRaw = getIntermediaryUrl().get();
 
-		LOGGER.info("Downloading intermediary from {}", url);
+		if (project != null && urlRaw.equals(LoomGradleExtensionApiImpl.DEFAULT_INTERMEDIARY_URL)) {
+			final Configuration config = project.getConfigurations()
```
modmuss50: A deteached configuration should work here

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -45,24 +55,59 @@ public abstract class IntermediaryMappingsProvider extends IntermediateMappingsP
...
+					.getByName(Constants.Configurations.INTERMEDIARY_MAPPINGS);
+			try {
+				config.defaultDependencies(new Action<DependencySet>() {
+					@Override
+					public void execute(final DependencySet dependencySet) {
+						final ModuleDependency intermediaryDep = (ModuleDependency) project.getDependencies()
+								.create("net.fabricmc:intermediary:" + encodedMcVersion);
```
modmuss50: Nit: move the group:name to a constant

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -45,24 +55,59 @@ public abstract class IntermediaryMappingsProvider extends IntermediateMappingsP
...
+						intermediaryDep.artifact(new Action<DependencyArtifact>() {
```
modmuss50: Would [DependencyFactory::create](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/dsl/DependencyFactory.html#create-java.lang.String-java.lang.String-java.lang.String-java.lang.String-java.lang.String-) make it easier to specify the classifier? I havent used it so im not 100% sure.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -45,24 +55,59 @@ public abstract class IntermediaryMappingsProvider extends IntermediateMappingsP
...
+				// We only care about adding default dependencies once
+				if (!dataException.getMessage().contains("after it has been resolved")) {
+					throw dataException;
+				}
+			}
 
-		Files.deleteIfExists(tinyMappings);
-		Files.deleteIfExists(intermediaryJarPath);
+			Files.copy(
+					config.getIncoming().getArtifacts().getArtifactFiles().getSingleFile().toPath(),
```
modmuss50: Would `config.getSingleFile().toPath()` do the same job?

src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
```diff
@@ -49,4 +50,21 @@ public abstract class IntermediateMappingsProvider implements Named {
...
+	}
+
+	public abstract static class Contextual extends IntermediateMappingsProvider {
+		@Override
+		public final void provide(Path tinyMappings) throws IOException {
+			throw new UnsupportedOperationException();
+		}
+
+		public abstract void provide(Path tinyMappings, Project project) throws IOException;
+	}
```
modmuss50: I think these should be internal and not public API so they can be changed at a later date, custom intermediary providers can just use the downloader.

I have a been a little bit cautious about passing `Project`s around freely like this. Its very easy to break the rules around configuration caching. It also makes it a little harder to unit test (in this case thats not an issue though).

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
jpenilla: The idea behind using `defaultDependencies` and a named configuration was to allow specifying a custom artifact in the build script.

modmuss50: Humm, I see. Im not sure we need a 3rd way to change where intermediary comes from, its not something people do often.

modmuss50 (approved):

PR #1050, opened 2024-02-18
Injected Interfaces Expansion => Generics Support

### LINKED ISSUE #597: Interface injection does not preserve generics
Let's say we have an event for registry additions, and a methods to get that event for a specific registry.
Now let's say this method is in an interface and we make it implement on Registry.

Registries are typed, and we have a context object, which can return the entry value, so a generic. The event would look like this:
```java
@FunctionalInterface
public interface EntryAdded<V> {
	void onAdded(RegistryEntryContext<V> context);
}
```

And the interface to implement on the registry:
```java
@FunctionalInterface
	public interface EntryAdded<V> {
		void onAdded(RegistryEntryContext<V> context);
	}
}
```

And we have a mixin for implementation:
```java
@Mixin(Registry.class)
public class RegistryMixin<V> implements RegistryEventStorage<V> {
	@Override
	public Event<RegistryEvents.EntryAdded<V>> getEntryAddedEvent() {
		return new Event<>(new Identifier("test"));
	}
}
```

So even on the mixin we have a generic type, though mixin does not seem to pay much attention to it.
Now let's reload with loom, and look at the generated Registry class file:

```java
public abstract class Registry<T> implements Keyable, IndexedIterable<T>, RegistryEventStorage {
```

This definitely does not look right, currently interface injection does not preserve any generic types.
This can be quite a problem for such cases.

Here's the stuff to replicate this: https://github.com/LambdAurora/loom_interface_injection_issue/tree/type_erasure

FirstMegaGame4:
Before this pull request, modders were able to add injected interfaces in other classes, but they were not able to use generics for them. This pull request allows to specify the raw generics signature (which is a part of the [ClassSignature](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-ClassSignature)) after the interface name in `fabric.mod.json`, solving the issue.

Examples:

Before:

```json
"net/minecraft/class_x": [
    "com/example/WithoutGenerics",
    "com/example/WithGenerics" // but we can't specify them
]
```

After:

```json
"net/minecraft/class_x": [
    "com/example/WithoutGenerics",
    "com/example/WithGenerics<Ljava/lang/String;>"
]
```

Fixes #597

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
@@ -284,8 +284,8 @@ public abstract class InterfaceInjectionProcessor implements MinecraftJarProcess
 
 			// See JVMS: https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-ClassSignature
 			if (InjectedInterface.containsGenerics(injectedInterfaces) && signature == null) {
-				// Classes that are not using generics don't need signatures, so they are null
-				// So if the class is not using generics but that an injected interface is using them, we are creating the signature
+				// Classes that are not using generics don't need signatures, so their signatures are null
+				// If the class is not using generics but that an injected interface targeting the class is using them, we are creating the class signature
 				StringBuilder baseSignatureBuilder = new StringBuilder("L" + superName + ";");
 
 				for (String baseInterface : baseInterfaces) {
--- a/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
@@ -41,11 +41,17 @@ import javax.inject.Inject;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+
+import net.fabricmc.tinyremapper.api.TrRemapper;
+
 import org.jetbrains.annotations.Nullable;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.ClassWriter;
 import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.signature.SignatureReader;
+import org.objectweb.asm.signature.SignatureVisitor;
+import org.objectweb.asm.util.CheckSignatureAdapter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -113,7 +119,7 @@ public abstract class InterfaceInjectionProcessor implements MinecraftJarProcess
 					.map(injectedInterface -> remap(
 							injectedInterface,
 							s -> mappings.mapClassName(s, intermediaryIndex, namedIndex),
-							s -> tinyRemapper.get().getEnvironment().getRemapper().mapSignature(s, true)
+							tinyRemapper.get().getEnvironment().getRemapper()
 					))
 					.toList();
 
@@ -125,17 +131,13 @@ public abstract class InterfaceInjectionProcessor implements MinecraftJarProcess
 		}
 	}
 
-	private InjectedInterface remap(InjectedInterface in, Function<String, String> remapper, Function<String, String> signatureRemapper) {
-		String className = remapper.apply(in.className());
-		String ifaceName = remapper.apply(in.ifaceName());
-		String generics = null;
-
-		if (in.generics() != null) {
-			String fakeSignature = signatureRemapper.apply("Ljava/lang/Object" + in.generics() + ";"); // Turning the raw generics string into a fake signature
-			generics = fakeSignature.substring("Ljava/lang/Object".length(), fakeSignature.length() - 1); // Retrieving the remapped raw generics string from the remapped fake signature
-		}
-
-		return new InjectedInterface(in.modId(), className, ifaceName, generics);
+	private InjectedInterface remap(InjectedInterface in, Function<String, String> remapper, TrRemapper signatureRemapper) {
+		return new InjectedInterface(
+				in.modId(),
+				remapper.apply(in.className()),
+				remapper.apply(in.ifaceName()),
+				in.generics() != null ? signatureRemapper.mapSignature(in.generics(), false) : null
+		);
 	}
 
 	private List<Pair<String, ZipUtils.UnsafeUnaryOperator<byte[]>>> getTransformers(List<InjectedInterface> injectedInterfaces) {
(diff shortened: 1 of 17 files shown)


src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -105,23 +107,35 @@ public void processJar(Path jar, Spec spec, ProcessorContext context) throws IOE
...
 		}
 	}
 
-	private InjectedInterface remap(InjectedInterface in, Function<String, String> remapper) {
-		return new InjectedInterface(
-				in.modId(),
-				remapper.apply(in.className()),
-				remapper.apply(in.ifaceName())
-		);
+	private InjectedInterface remap(InjectedInterface in, Function<String, String> remapper, Function<String, String> signatureRemapper) {
```
modmuss50: Nit: Id pass the Remapper into this cass, instead of using the two Function's

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -105,23 +107,35 @@ public void processJar(Path jar, Spec spec, ProcessorContext context) throws IOE
...
+		String className = remapper.apply(in.className());
+		String ifaceName = remapper.apply(in.ifaceName());
+		String generics = null;
+
+		if (in.generics() != null) {
+			String fakeSignature = signatureRemapper.apply("Ljava/lang/Object" + in.generics() + ";"); // Turning the raw generics string into a fake signature
+			generics = fakeSignature.substring("Ljava/lang/Object".length(), fakeSignature.length() - 1); // Retrieving the remapped raw generics string from the remapped fake signature
```
modmuss50: Just from a quick look at ASM's code, it seems if you pass `typeSignature` as false when calling mapSignature you might be able to get away without this

src/test/groovy/net/fabricmc/loom/test/unit/processor/InterfaceInjectionProcessorTest.groovy
```diff
@@ -83,10 +86,20 @@ class InterfaceInjectionProcessorTest extends Specification {
...
 			loadedClass.interfaces.first().name == "net/fabricmc/loom/test/unit/processor/classes/SimpleInterface"
 			loadedClass.constructors.first().newInstance().injectedMethod() == 123
 		}
+
+		"class_3" | "net/fabricmc/loom/test/unit/processor/classes/GenericInterface<Ljava/lang/String;>" | GenericTargetClass.class { Class<?> loadedClass ->
+			loadedClass.interfaces.first().name == "net/fabricmc/loom/test/unit/processor/classes/GenericInterface"
+			loadedClass.constructors.first().newInstance().genericInjectedMethod() == null
+		}
+
+		"class_4" | "net/fabricmc/loom/test/unit/processor/classes/GenericInterface<TT;>" | PassingGenericTargetClass.class { Class<?> loadedClass ->
```
modmuss50: These tests look good, however I think there are quite a few more situations that we need to make sure are covered:

- Target a class with two generic interfaces
- Test that signature remapping is working, e.g `GenericInterface<Lclass_3;>`
- More complex edge cases such as https://discord.com/channels/507304429255393322/592749519171616796/1208823754072006747

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -210,10 +224,19 @@ public static List<InjectedInterface> fromMod(FabricModJson fabricModJson) {
...
 			for (String className : addedIfaces.keySet()) {
-				final JsonArray ifaceNames = addedIfaces.getAsJsonArray(className);
+				final JsonArray ifacesInfo = addedIfaces.getAsJsonArray(className);
+
+				for (JsonElement ifaceElement : ifacesInfo) {
+					String ifaceInfo = ifaceElement.getAsString();
+
+					if (ifaceInfo.contains("<") && ifaceInfo.contains(">")) {
+						String ifaceName = ifaceInfo.substring(0, ifaceInfo.indexOf("<"));
+						String generics = ifaceInfo.substring(ifaceInfo.indexOf("<"));
```
modmuss50: Can we do some basic validation of the generics here? ASM has a `CheckSignatureAdapter` that looks somewhat useful.

It may be even possible to validate some additional stuff when applying to the class, such as checking that `T` is present on the target class? Im not too worried about this.

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -210,10 +224,19 @@ public static List<InjectedInterface> fromMod(FabricModJson fabricModJson) {
...
 
-				for (JsonElement ifaceName : ifaceNames) {
-					result.add(new InjectedInterface(modId, className, ifaceName.getAsString()));
+						result.add(new InjectedInterface(modId, className, ifaceName, generics));
+					} else {
+						result.add(new InjectedInterface(modId, className, ifaceElement.getAsString(), null));
+					}
```
modmuss50: Nit: drop the else, just have a single `new InjectedInterface` that passes a nullable generics String

modmuss50 (commented):
This looks like a great start 👍

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
FirstMegaGame4: I tested with but it didn't work

src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -210,10 +231,23 @@ public static List<InjectedInterface> fromMod(FabricModJson fabricModJson) {
...
+					if (ifaceInfo.contains("<") && ifaceInfo.contains(">")) {
+						generics = ifaceInfo.substring(ifaceInfo.indexOf("<"));
+
+						// First Generics Check, if there are generics, are them correctly written?
+						SignatureReader reader = new SignatureReader("Ljava/lang/Object" + generics + ";");
+						CheckSignatureAdapter checker = new CheckSignatureAdapter(CheckSignatureAdapter.CLASS_SIGNATURE, null);
+						reader.accept(checker);
+					}
+
+					result.add(new InjectedInterface(modId, className, ifaceElement.getAsString(), generics));
```
modmuss50: I think this is wrong, as `ifaceElement.getAsString()` will also contain the generics, you just want the raw classname.

modmuss50 (commented):
This is looking good, seems to be struggling to build though.

modmuss50 (approved):

PR #1053, opened 2024-02-19
Initial config caching work

modmuss50:
Trying an incremental approach at supporting config caching.


modmuss50:
The test harness is tottaly broken with configuration caching, in current shipping loom `gradlew --configuration-cache help` works just fine. Right now its exploding on a simple `SourceSet.getResources()` when ran using the test harness.

I dont get how we are supposed to support this, I had other problems when I looked into this last year: https://github.com/FabricMC/fabric-loom/pull/893

PR #1058, opened 2024-02-25
Explore looking up the process holding a file handle.

modmuss50:
Based on: https://devblogs.microsoft.com/oldnewthing/20120217-00/?p=8283

Im a very wary against adding native code to loom for a range of reasons, but getting the bottom/providing better error messages for the locked file issues might make it worth it.


native/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+}
+
+library {
+	linkage = [
+		Linkage.SHARED
+	]
+}
+
+import org.gradle.internal.jvm.Jvm
+def jvmHome = Jvm.current().javaHome
```
Juuxel, line 20: Wouldn't it be better to use something like the `java.home` system property or the `JAVA_HOME` env var to avoid Gradle internals?

modmuss50: Probably yeah, blame my lazyness and the Jetbrains docs lol: https://www.jetbrains.com/help/idea/setting-up-jni-development-in-gradle-project.html#adding_jni_support

modmuss50:
Closing this PR as having this code build as part of the loom build will be a nightmare. I will move it into a sperate lib, similar to how the native code is built for the installer.

PR #1061, opened 2024-03-03
Refactor/better kotlin metadata

LlamaLad7:
The library provides extension properties so we don't need to rely on the internal APIs. This makes the code much cleaner and should be more stable.
Also bumps the library and adds 1 missing piece of remapping.


modmuss50 (approved):
This is great thanks, no idea why I over complicated it so much 👍

Issue #929, opened 2023-07-13
Concurrent gradle reload causes fails


Flemmli97:
I have a multi module project in intellij and when i reload all the gradle projects at the same time it fails due to some library redownload issue.
If i reload the projects one after another it works fine.
This is due to loom creating a lock during a reload and after finishing deleting it but when reloading the lock is present it will try to redownload all minecraft jars etc. again (see https://github.com/FabricMC/fabric-loom/blob/dev/1.3/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java#L94). Since all projects now are downloading and writing to the same file this obviously causes issue.

One underlying problem is that the lock file is the same for all the project since its using the first 16 string of the projects path which if you have something like `../Modding/Project1`   `../Modding/Project2` will be the same. Im not too sure why its using the first 16 when the last 16 would be more defining (see https://github.com/FabricMC/fabric-loom/blob/dev/1.3/src/main/java/net/fabricmc/loom/util/Checksum.java#L89).

I am aware that this will not fix concurrent downloads when the libraries are missing in the first place (or the user forces a refresh dependencies for all project at once) but when the libraries are already present this issue shouldn't happen.

Note:

To solve concurrently downloading the same file we could instead:
1. Download the file to the path + append the project name to it so it is a different file.
2. If the download is finished rename the file to what it originally was

This way 2 instances downloading the same file wont overwrite the same file when downloading and cause issues writing different data to it.

modmuss50:
We now have better locking in place. If there are still issues with the latest loom please open a new issue. Thanks.

PR #1064, opened 2024-03-04
Use last 16 chars for checksum

Flemmli97:
This makes it to use the last 16 chars for the checksum used to create a lock.
Currently it uses the first 16 chars which makes structures like
```
Foo/bar/project_1
            .../project_2
```

have the same lock file.

This is of interest due to #929 (and the advantages are outlined there too).

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/Checksum.java
+++ b/src/main/java/net/fabricmc/loom/util/Checksum.java
@@ -96,6 +96,7 @@ public class Checksum {
 
 	public static String projectHash(Project project) {
 		String str = project.getProjectDir().getAbsolutePath() + ":" + project.getPath();
-		return toHex(str.getBytes(StandardCharsets.UTF_8)).substring(0, 16);
+		String hex = toHex(str.getBytes(StandardCharsets.UTF_8));
+		return hex.substring(hex.length() - 16);
 	}
 }
--- a/src/main/java/net/fabricmc/loom/util/Checksum.java
+++ b/src/main/java/net/fabricmc/loom/util/Checksum.java
@@ -96,7 +96,7 @@ public class Checksum {
 
 	public static String projectHash(Project project) {
 		String str = project.getProjectDir().getAbsolutePath() + ":" + project.getPath();
-		String hex = toHex(str.getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/util/Checksum.java
```diff
@@ -96,6 +96,7 @@ public static String toHex(byte[] bytes) {
 
 	public static String projectHash(Project project) {
 		String str = project.getProjectDir().getAbsolutePath() + ":" + project.getPath();
-		return toHex(str.getBytes(StandardCharsets.UTF_8)).substring(0, 16);
+		String hex = toHex(str.getBytes(StandardCharsets.UTF_8));
```
modmuss50: I think the intended code whould be to hash the string, and then substring, eg. use `sha1Hex` instead of `hex`

modmuss50:
https://github.com/FabricMC/fabric-loom/pull/1031 is also going to help a lot with the original issue.

src/main/java/net/fabricmc/loom/util/Checksum.java
Flemmli97: i was just using the original code. but i can make it use the hash version instead in this pr if thats fine

modmuss50: If you could that would be great. I think it better fits what this code was intended to to, especially considering the name of the method.

Flemmli97:
> #1031 is also going to help a lot with the original issue.

ye this also wouldnt really solve all the concurrent problems.

think i outlined it in the issue but:
this solves reloading all project at once assuming all needed files are already downloaded which are most of the cases. if you e.g. add a dependency and manually reload this problem doesnt happen either way.
does not solve the problem with initializing all projects at once.

for that we could implement following:

Download the file to the path + append the project name to it so it is a different file.
E.g. if we wanted to download `mc-1.20.1.jar` to `./gradle/caches/whatever/`
it would download it as `./gradle/caches/whatever/mc-1.20.1_projectname.jar`.
And when its done we rename it back to `mc-1.20.1.jar`.

This way the download will not conflict and even if we rename (+ replace) the existing file its still basically the same file

modmuss50 (approved):

PR #1068, opened 2024-03-09
Optimise fabric.mod.json files

### LINKED ISSUE #1067: Ensure that schemaVersion is the first entry in a FMJ when remapping
Ensure that schemaVersion is the first entry in a FMJ when remapping, as a minor optimistation.

Once done the loader warning here: https://github.com/FabricMC/fabric-loader/blob/d69cb72d26497e3f387cf46f9b24340b402a4644/src/main/java/net/fabricmc/loader/impl/metadata/ModMetadataParser.java#L131 can likely be dropped to debug level, or even removed.

modmuss50:


zOnlyKroks (approved):
LGTM

modmuss50:
I have decided to make this opt-in as it will change the fomatting of the json file, and no one is ever going to agree if its should be minifed or not.

zOnlyKroks:
Sounds reasonable to me

PR #1071, opened 2024-03-10
Publish kotlin remapper to maven

modmuss50:
No guarantees about stability, or it even exissting. Just a little bit better than people copy/pasting code.


modmuss50:
Going to pass on this I think, long term with Kotlin 2.0 this can likely be driven by FLK.

PR #1077, opened 2024-03-19
Use different cache files for custom manifest and metadata jsons

SpaceWalkerRS:
This PR serves to replace #1069 and resolve the issues discussed within it a different way.

- A new API is added where plugins can provide custom versions manifests. All versions manifests (including the built in ones) are sorted in a priority queue so mods can choose let their custom manifests act either as a fallback or to override the built in manifests.
- The manifest cache file name gets a hash suffix if a custom versions manifest URL is given through this new API. The hash is computed by calling `hashCode` on the manifest URL string and converting that to hex.
- The Minecraft metadata cache file name gets a hash suffix if a custom metadata URL is given (through the `customMinecraftManifest` property in the Loom extension API) or if the versions manifest is custom. The hash is computed by calling `hashCode` on the custom metadata URL string if it is present, or else by calling `hashCode` on the manifest URL string, and converting that to hex.

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -142,7 +142,15 @@ public interface LoomGradleExtensionAPI {
 	@ApiStatus.Experimental
 	ManifestLocations getVersionsManifests();
 
-	Property<String> getCustomMinecraftManifest();
+	/**
+	 * @deprecated use {@linkplain #getCustomMinecraftMetadata} instead
+	 */
+	@Deprecated
+	default Property<String> getCustomMinecraftManifest() {
+		return getCustomMinecraftMetadata();
+	}
+
+	Property<String> getCustomMinecraftMetadata();
 
 	SetProperty<String> getKnownIndyBsms();
 
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -134,16 +134,20 @@ public interface LoomGradleExtensionAPI {
 
 	InterfaceInjectionExtensionAPI getInterfaceInjection();
 
+	@ApiStatus.Experimental
 	default void versionsManifests(Action<VersionsManifestsAPI> action) {
 		action.execute(getVersionsManifests());
 	}
 
+	@ApiStatus.Experimental
 	ManifestLocations getVersionsManifests();
 
+	@ApiStatus.Experimental
 	default void experimentalVersionsManifests(Action<VersionsManifestsAPI> action) {
 		action.execute(getExperimentalVersionsManifests());
 	}
 
+	@ApiStatus.Experimental
 	ManifestLocations getExperimentalVersionsManifests();
 
 	Property<String> getCustomMinecraftManifest();
--- a/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
+++ b/src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
@@ -43,6 +43,7 @@ import org.gradle.api.tasks.SourceSet;
 import org.jetbrains.annotations.ApiStatus;
 
 import net.fabricmc.loom.api.decompilers.DecompilerOptions;
+import net.fabricmc.loom.api.manifest.VersionsManifestsAPI;
 import net.fabricmc.loom.api.mappings.intermediate.IntermediateMappingsProvider;
 import net.fabricmc.loom.api.mappings.layered.spec.LayeredMappingSpecBuilder;
 import net.fabricmc.loom.api.processor.MinecraftJarProcessor;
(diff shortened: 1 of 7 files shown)


modmuss50:
I dont think this does replace #1069 as it doesnt really fullfil `an API allowing third party plugins to provide their own versions manifest and version metadata` imo.

Just to clarfiy/surmise what the requirements are, it would be something like:

> A 3rd party plugin should be able to provide its own metadata for a custom minecraft version, that is not present in the default metadata.

Im not sure we want 3rd party plugins hijacking and replacing the mirrors, it wasnt designed with this in mind.

SpaceWalkerRS:
> Im not sure we want 3rd party plugins hijacking and replacing the mirrors, it wasnt designed with this in mind.

I see. Perhaps we can add custom properties to the Loom extension API, the same way there is already a custom metadata url property there?

modmuss50:
Maybe, I think this is a bit better, but could possibly just be a single `ListProperty<String> getAdditionalVersionsManifests();`, allowing plugins to register their own manifests alonside the existing vanilla ones? I imagine the vanilla and fabric one should take priority?

SpaceWalkerRS:
Lists could work for that for sure, I like that idea. But I think the custom manifests should take priority, to make the point of these properties to allow plugins to override their own manifests, but have the default manifests (i.e. the Mojang and Fabric ones) to fall back on.

modmuss50:
Ok sure, not tottaly keen on them being able to override, as it will mess up the caches but other than that its not a major issue.

SpaceWalkerRS:
The lists will require some shuffling around of the logic, 'cause you no longer know which URL is gonna be used, and I used that to change the cache file names.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -132,6 +134,18 @@ default void interfaceInjection(Action<InterfaceInjectionExtensionAPI> action) {
...
+		action.execute(getVersionsManifests());
+	}
+
+	ManifestLocations getVersionsManifests();
+
+	default void experimentalVersionsManifests(Action<VersionsManifestsAPI> action) {
+		action.execute(getExperimentalVersionsManifests());
+	}
+
+	ManifestLocations getExperimentalVersionsManifests();
```
modmuss50: NIT: these should all be @ApiStatus.Experimental

src/main/java/net/fabricmc/loom/api/manifest/VersionsManifestsAPI.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package net.fabricmc.loom.api.manifest;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Experimental
+public interface VersionsManifestsAPI {
+	/**
+	 * Adds a URL to a versions manifest json with the default priority of {@code 0}.
```
modmuss50, line 32: NIT: might be nice to mention how the priority work, is 0 first or last?

src/main/java/net/fabricmc/loom/api/manifest/VersionsManifestsAPI.java
```diff
@@ -0,0 +1,45 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package net.fabricmc.loom.api.manifest;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Experimental
+public interface VersionsManifestsAPI {
```
modmuss50, line 30: Nit drop API here, and use Impl on the other class.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
modmuss50: Why is the experimental one needed? Couldnt these just be one, with the experimental one coming after the default manifest?

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftMetadataProvider.java
```diff
@@ -123,65 +134,77 @@ private ManifestVersion.Versions getVersionEntry() throws IOException {
...
 		} else {
 			builder.defaultCache();
 		}
 
-		final String json = builder.downloadString(options.minecraftMetadataPath());
+		final String fileName = versionEntry.manifest == null
+				? "minecraft-info-" + versionEntry.entry.url.hashCode() + ".json"
+				: !versionEntry.manifest.isBuiltIn()
+						? "minecraft-info-" + versionEntry.manifest.url().hashCode() + ".json"
+						: "minecraft-info.json";
```
modmuss50: Hard to read this many ternary's split it out into a function.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
SpaceWalkerRS: I made some changes to combine them, let me know if this is what you had in mind

modmuss50 (approved):

PR #1080, opened 2024-03-23
Work on better `include` configuration

lukebemish:
Currently, the `include` configuration uses dependency resolution information at runtime; additionally, it works based off of `Dependency` objects, which is generally speaking a bad way to pass dependency information into a task - we're meant to use the various dependency resolution result types instead. All of this combined means that the system is not very future-proof. This PR is an attempt to modernize the system, while keeping the old behavior.


src/main/java/net/fabricmc/loom/build/nesting/NestableJarGenerationTask.java
```diff
@@ -0,0 +1,252 @@
...
+		getJarIds().get().forEach((fileName, metadata) -> {
+			fabricModJsons.put(fileName, generateModForDependency(metadata));
+		});
+
+		try {
+			File targetDir = getOutputDirectory().get().getAsFile();
+			FileUtils.deleteDirectory(targetDir);
+			targetDir.mkdirs();
+		} catch (IOException e) {
+			throw new org.gradle.api.UncheckedIOException(e);
```
modmuss50:
```suggestion
			throw new UncheckedIOException(e);
```

src/main/java/net/fabricmc/loom/build/nesting/NestableJarGenerationTask.java
```diff
@@ -0,0 +1,252 @@
...
+import net.fabricmc.loom.util.ZipReprocessorUtil;
+import net.fabricmc.loom.util.fmj.FabricModJsonFactory;
+
+public abstract class NestableJarGenerationTask extends DefaultTask {
+	private static final Logger LOGGER = LoggerFactory.getLogger(NestableJarGenerationTask.class);
+	private static final String SEMVER_REGEX = "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$";
+	private static final Pattern SEMVER_PATTERN = Pattern.compile(SEMVER_REGEX);
+
+	@InputFiles
+	@PathSensitive(PathSensitivity.NAME_ONLY)
```
modmuss50, line 73: Question: Why name only?

src/main/java/net/fabricmc/loom/build/nesting/NestableJarGenerationTask.java
```diff
@@ -0,0 +1,252 @@
...
+				} else if (moduleLocation.version == null) {
+					throw new RuntimeException("Attempted to nest artifact " + id + " which has no version");
+				}
+
+				String group = moduleLocation.group;
+				String name = moduleLocation.name;
+				String version = moduleLocation.version;
+				String classifier = null;
+
+				if (artifact.getFile().getName().startsWith(name + "-" + version + "-")) {
```
modmuss50, line 149: Shame there is no better way to get the classifier, this seems quite fragile.

modmuss50 (approved):
This looks a lot better, thanks. Just a few minor fixes + questions.

Sorry it took me so long to look at this.

src/main/java/net/fabricmc/loom/build/nesting/NestableJarGenerationTask.java
lukebemish: The only part of the path to the jar to nest that affects the output is it's name and contents; the relative/absolute path is irrelevant

src/main/java/net/fabricmc/loom/build/nesting/NestableJarGenerationTask.java
lukebemish: Yeah unfortunately this is the only way to do this (without doing the questionable approach it did before); classifiers are a maven concept that doesn't translate super well into gradle. I'd encourage people to use capability requests and feature variants instead.

All that said -- this was the workaround I found on a relevant gradle issue about the lack of a way to query artifact classifier (I can try to track that down later), and it should actually be pretty stable given that the classifier and extension provided are used only while gradle is resolving which artifact of a module to use; so if something was selected by classifier it should be in the name. At any rate, probably less fragile than the old system still.

lukebemish: See https://github.com/gradle/gradle/issues/23702 -- basically, at the level we're operating at we're in variant aware resolution, and gradle doesn't really care about classifiers at that level because they're a very maven concept, so they're not there

modmuss50 (approved):

PR #1107, opened 2024-04-25
Sandbox support

modmuss50:
No promises the otherside of this ever sees the light of day, might just remove it at somepoint if stuff doesnt work out.

### DIFF
--- a/src/test/groovy/net/fabricmc/loom/test/integration/SandboxTest.groovy
+++ b/src/test/groovy/net/fabricmc/loom/test/integration/SandboxTest.groovy
@@ -67,7 +67,9 @@ class SandboxTest extends Specification implements GradleProjectTestTrait {
 
 		then:
 		result.task(":runClientSandbox").outcome == SUCCESS
-		result.output.contains("Running real main: net.fabricmc.loader.impl.launch.knot.KnotClient")
+		result.output.contains("Running real main: net.fabricmc.devlaunchinjector.Main")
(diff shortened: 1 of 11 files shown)


src/test/groovy/net/fabricmc/loom/test/integration/SandboxTest.groovy
```diff
@@ -0,0 +1,110 @@
...
+		def mavenHelper = new LocalMavenHelper("net.fabricmc.loom.test", "sandbox", "1.0.0", null, gradle.getMavenLocalDir().toPath())
+		mavenHelper.copyToMaven(createDummySandboxJar(), null)
+		mavenHelper.savePom()
+
+		when:
+		def result = gradle.run(task: "runClientSandbox")
+
+		then:
+		result.task(":runClientSandbox").outcome == SUCCESS
+		result.output.contains("Running real main: net.fabricmc.loader.impl.launch.knot.KnotClient")
```
modmuss50, line 70: Note to self: This is wrong, it should be DLI, and we need to test that DLI has this class. It should work as follows:
- Privaliged JVM
- Sandbox entrypoint
- Sandboxed JVM
- DLI entrypoint
- Game entrypoint

modmuss50: Now changed so the sandbox tottaly replaces DLI.

PR #1112, opened 2024-04-29
Always remap changing deps

modmuss50:
Is there a better way to know if a ResolvedArtifact is changing?

TODO:
- Check snapshots, specially specific versions
- Maybe check the file hash, so we dont always remap?

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/mods/ArtifactRef.java
+++ b/src/main/java/net/fabricmc/loom/configuration/mods/ArtifactRef.java
@@ -52,9 +52,11 @@ public interface ArtifactRef {
 
 	@Nullable String classifier();
 
+	boolean changing();
+
 	void applyToConfiguration(Project project, Configuration configuration);
 
-	record ResolvedArtifactRef(ResolvedArtifact artifact, @Nullable Path sources) implements ArtifactRef {
+	record ResolvedArtifactRef(ResolvedArtifact artifact, @Nullable Path sources, boolean changing) implements ArtifactRef {
 		@Override
 		public Path path() {
 			return artifact.getFile().toPath();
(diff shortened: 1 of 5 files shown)


Axionize:
How large is the performance hit here? Is that the only reason this hasn't been merged? If so, then could you consider use something like https://github.com/Cyan4973/xxHash to quickly determine if a `changing` jar has actually changed and if to remap it?

modmuss50:
I dont know tbh, I would need to find out, I might get the PR updated to the latest loom version. Hashing the file also means reading it from disk, could maybe get away with just checking the last updated time on the file or something. 🤔

Axionize:
> I dont know tbh, I would need to find out, I might get the PR updated to the latest loom version. Hashing the file also means reading it from disk, could maybe get away with just checking the last updated time on the file or something. 🤔

Yes that could work, you could also have an easy fail chain. Store the last modified dates of each dep and then

```
for (dep : depList) {
   if (!dep.isChanging()) return;
   if (dep.getStoredLastModifiedTimestamp() == dep.getFile().getLastModifiedTimestamp()) return;
   if (dep.getHash() == hash(dep.getFile()) return;
   rempDep(dep)
}
```

So 99% of the time basically nothing will change, you have very fast checking and early exit of processing most jars, and you will only have to remap precisely when needed

Player3324:
size and mtime should be a good default

I wouldn't allow newer but only identical unlike the previous comment suggests

Axionize:
> size and mtime should be a good default
> 
> I wouldn't allow newer but only identical unlike the previous comment suggests

Typo, agreed.

Axionize:
Is this not happening then @modmuss50 ?

modmuss50:
No, I dont think so. Its a pain to do and tottaly solved with the unobf versions.

PR #1116, opened 2024-05-05
Config cache support for remap jar.

modmuss50:
This is roughly based on my first attempt


modmuss50:
Superseeded by all the work gone into 1.8

PR #1118, opened 2024-05-10
Add super classes to decompiler cache ClassEntry

### LINKED ISSUE #1117: Decompiler cache causes suboptimal results when reused
When the decompiler cache gets reused weird changes occur, which aren't present when generating the sources without using the cache.

This seems to happen when a big amount of cache hits are present, as is the case between two versions of the game that are adjacent in time, and when adding/removing an access widener entry. In the first case it pollutes the diff quite a lot, and reduces readability in both cases.

## Steps to reproduce
0. Delete your decompiler cache.
1. Clone the [reproduction repo](https://github.com/VelizarBG/decompilercachetest).
2. Checkout the 1.20.6 branch.
3. Run `gradlew genSourcesWithVineflower --use-cache`
4. Checkout the 24w18a branch.
5. Repeat step 3.
6. Store a copy of the generated sources somewhere.
7. Run `gradlew genSourcesWithVineflower --no-use-cache`
8. Diff the sources generated with cache and the ones without.
9. Notice the suboptimal differences(e.g. missing `@Override` annotations, unnecessary casts, missing generics, and so on).

PS: This also applies to the CFR decompiler. Haven't tested FernFlower.

modmuss50:
Fixes #1117


modmuss50:
This does not fix the issue 😢 

It might be something more complex, possibly a decompiler issue.

modmuss50:
Turns out it wasnt that complex, now fixed :)

PR #1134, opened 2024-06-24
Test against Gradle 8.9

modmuss50:
PR to just run the tests


modmuss50:
Failed tests are due to test code in the 8.10 nightlies, so not a concern.

PR #1141, opened 2024-07-06
Add SHA1 suffix to intermediaries

Wyvest:
Having two or more different intermediaries for one version in a project breaks. Although this isn't possible in vanilla Loom, one example would be loading Legacy Forge 1.8.9 and Legacy Fabric 1.8.9 in one project (via something like preprocessor).
Attaching a differentiating name to each intermediary we download, we use this to differentiate between intermediaries.

To be honest I made this around a year ago and never PRed it here so this might be a really terrible approach. Please feel free to give feedback.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
@@ -25,9 +25,12 @@
 package net.fabricmc.loom.configuration.providers.mappings;
 
 import java.io.IOException;
+import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
 
 import javax.inject.Inject;
 
@@ -108,8 +111,12 @@ public abstract class IntermediaryMappingsProvider extends IntermediateMappingsP
 	@Override
 	public @NotNull String getName() {
 		final String encodedMcVersion = UrlEscapers.urlFragmentEscaper().escape(getMinecraftVersion().get());
-		final String url = getIntermediaryUrl().get().formatted(encodedMcVersion);
-		return NAME + "-" + getSha1(url);
+		final String urlRaw = getIntermediaryUrl().get();
+		if (!LoomGradleExtensionApiImpl.DEFAULT_INTERMEDIARY_URL.equals(urlRaw)) {
+			final String url = getIntermediaryUrl().get().formatted(encodedMcVersion);
+			return NAME + "-" + getSha1(url);
+		}
+		return NAME;
 	}
 
 	private String getSha1(String value) {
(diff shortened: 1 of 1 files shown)


modmuss50 (commented):
I dont think we need this for official intermediary names.

SpaceWalkerRS:
Isnt this what the different intermediary providers are for? You can create a new provider and change the name and it'll end up in a different cache file.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -107,6 +110,24 @@ public void execute(final DependencyArtifact dependencyArtifact) {
...
+		final String encodedMcVersion = UrlEscapers.urlFragmentEscaper().escape(getMinecraftVersion().get());
+		final String urlRaw = getIntermediaryUrl().get();
+		if (!LoomGradleExtensionApiImpl.DEFAULT_INTERMEDIARY_URL.equals(urlRaw)) {
+			final String url = getIntermediaryUrl().get().formatted(encodedMcVersion);
+			return NAME + "-" + getSha1(url);
+		}
 		return NAME;
 	}
+
+	private String getSha1(String value) {
```
modmuss50: Loom already has a helper for this, see Checksum.sha1Hex

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -107,6 +110,24 @@ public void execute(final DependencyArtifact dependencyArtifact) {
 
 	@Override
 	public @NotNull String getName() {
+		final String encodedMcVersion = UrlEscapers.urlFragmentEscaper().escape(getMinecraftVersion().get());
+		final String urlRaw = getIntermediaryUrl().get();
+		if (!LoomGradleExtensionApiImpl.DEFAULT_INTERMEDIARY_URL.equals(urlRaw)) {
```
modmuss50, line 115: This wont build due to checkstyle, please make sure it builds and actually works by testing it.

I dont see an automated test for this (likely becuase its quite hard to test), so it definitely should be manually tested.

modmuss50 (changes requested):

modmuss50 (approved):

Wyvest:
> Isnt this what the different intermediary providers are for? You can create a new provider and change the name and it'll end up in a different cache file.

Using the loom extension in your buildscript you can only set the intermediary URL. Besides, it isn't specified that you have to change the name and URL for it to work.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/IntermediaryMappingsProvider.java
```diff
@@ -107,6 +109,15 @@ public void execute(final DependencyArtifact dependencyArtifact) {
 
 	@Override
 	public @NotNull String getName() {
+		final String encodedMcVersion = UrlEscapers.urlFragmentEscaper().escape(getMinecraftVersion().get());
```
Octol1ttle, line 112: Wouldn't it be better to have this inside the `if` block (since that is the only place where it is used)?

PR #1148, opened 2024-07-18
Fix inner class detection in `JarWalker`

### LINKED ISSUE #1147: `genSources` fails with Loom 1.6 and 1.7 for old Minecraft versions
From the stacktrace it looks like it might be a decompiler cache issue:

```
Caused by: java.lang.RuntimeException: Failed to decompile, java.lang.IllegalStateException: Unexpected output: /paulscode/sound/SoundSystemConfig$Codec.java
        at net.fabricmc.loom.util.ExceptionUtil.createDescriptiveWrapper(ExceptionUtil.java:60)
        at net.fabricmc.loom.task.GenerateSourcesTask.run(GenerateSourcesTask.java:237)
Caused by: java.lang.IllegalStateException: Unexpected output: /paulscode/sound/SoundSystemConfig$Codec.java
        at net.fabricmc.loom.decompilers.cache.CachedJarProcessor.completeJob(CachedJarProcessor.java:185)
        at net.fabricmc.loom.task.GenerateSourcesTask.runWithCache(GenerateSourcesTask.java:288)
        at net.fabricmc.loom.task.GenerateSourcesTask.run(GenerateSourcesTask.java:233)
```

reproduction case (slightly modified example mod): https://github.com/SpaceWalkerRS/loom-1.7-decompile-issue-reproduction-case

SpaceWalkerRS:
fixes #1147 

This changes the inner class detection in the `JarWalker` to read the inner class and outer method attributes in the class file. Currently it checks for the presence of a `$` character in the class name. This is problematic for older versions of the game, where all the inner class and outer method attributes are stripped but the names of shaded libraries are not obfuscated and thus still contain the `$` character. This causes issues as the decompiler cache still treats these classes as inner classes while the decompilers do not.

thanks to @Juuxel for looking into this and realizing what the problem is

### DIFF
--- a/src/main/java/net/fabricmc/loom/decompilers/cache/JarWalker.java
+++ b/src/main/java/net/fabricmc/loom/decompilers/cache/JarWalker.java
@@ -49,6 +49,8 @@ import java.util.stream.Stream;
 
 import org.gradle.api.JavaVersion;
 import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.InnerClassNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -88,13 +90,12 @@ public final class JarWalker {
 					continue;
 				}
 
-				boolean isInnerClass = fileName.contains("$");
+				String outerClass = findOuterClass(fs, fileName);
 
-				if (isInnerClass) {
-					String outerClassName = fileName.substring(0, fileName.indexOf('$')) + ".class";
-					innerClasses.computeIfAbsent(outerClassName, k -> new ArrayList<>()).add(fileName);
-				} else {
+				if (outerClass == null) {
 					outerClasses.add(fileName);
+				} else {
+					innerClasses.computeIfAbsent(outerClass + ".class", k -> new ArrayList<>()).add(fileName);
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/decompilers/cache/JarWalker.java
```diff
@@ -127,6 +128,40 @@ public static List<ClassEntry> findClasses(FileSystemUtil.Delegate fs) throws IO
 		}
 	}
 
+	/**
+	 * Check if the given class file denotes and inner class and find the corresponding outer class name.
+	 */
+	private static String findOuterClass(FileSystemUtil.Delegate fs, String classFile) throws IOException {
```
Juuxel, line 134: This could be sped up by directly returning null if `$` is not in the path.

SpaceWalkerRS: that would break on any class with the attributes present but with the `$` character obfuscated away - though I can't think of any Minecraft versions where that is the case 🤔

Juuxel: Isn't this run on the named jar though? So if the names are broken, then it's just a sign of broken mappings (or missing mappings, but like you said, it shouldn't happen with obf names).

SpaceWalkerRS: I suppose, and as opposed to having it the other way around (with `$` without attributes), this wouldn't lead to a crash, right? I'd just mean that class isn't cached.

modmuss50 (approved):

PR #1152, opened 2024-07-31
Make no loader warning an info log

modmuss50:


Juuxel (approved):
Nice change, this is also annoying in multiloader projects where the shared code uses Loom without a loader dep.

PR #1154, opened 2024-08-04
Linux RISC-V support

modmuss50:


src/test/groovy/net/fabricmc/loom/test/unit/library/LibraryProcessorManagerTest.groovy
```diff
@@ -78,6 +78,29 @@ class LibraryProcessorManagerTest extends LibraryProcessorTest {
 		processed.find { it.is("org.lwjgl:lwjgl-glfw") && it.target() == Library.Target.NATIVES }.classifier() == "natives-macos"
 	}
 
+	// A test to ensure that we can add linux ARM64 support on an unsupported version
```
thecatcore: RISCV instead of ARM64.

PR #1156, opened 2024-08-08
Test 8.10-rc-1

modmuss50:
PR just to run the tests, will close if all goes well.


modmuss50:
The 2 test failuires are only in fragile test code, so all is good 👍

PR #1159, opened 2024-08-12
Remove multi project optimisation

### LINKED ISSUE #1044: Some field references are not remapped when building with Loom 1.5
I am having great difficulty finding a simple reproduction case for this, so I'll just link the [project](https://github.com/OrnitheMC/ornithe-standard-libraries) where I'm experiencing this bug instead. The issue occurs [here](https://github.com/OrnitheMC/ornithe-standard-libraries/blob/b7603010251dfb218346206ab212178bb23bca3a/libraries/resource-loader/resource-loader-client-mca1.2.2-1624-mc11w48a/src/main/java/net/ornithemc/osl/resource/loader/impl/BuiltInModTexturePack.java#L26), a value is assigned to a field from a super class. While references to this super class and methods from it are remapped to intermediary, these two field references are not.

I tried creating a project from the Fabric example mod, but could not reproduce it. But bringing the OSL project back to Loom 1.3 immediately fixed it again (could not use Loom 1.4 'cause in that version client-only and server-only projects were broken).

modmuss50:
This is the start of a major refactor to the remap jar tasks. Going to split it up over many PRs in an attmept to not break too much 😆


Axionize:
Is there any chance of bringing this back? This would be really useful for C2ME and also pretty useful for my many multi-version projects (which have a core and version-specific module built against the same version)

The based on the historic build time improvement it would be significant

modmuss50:
> The based on the historic build time improvement it would be significant

Please test again, in fabric API it only made it a few seconds slower on my machine. Its fundamentally not compatible with how Gradle wants things done.

PR #1162, opened 2024-08-20
Enable configuration cache for most tests by default

modmuss50:
# TODO:
- Client only entries in jars
- run game tasks
- ~~gen sources~~ A pain, skip for now
- ~~migrate mappings~~ Will do later, low priority
- vscode
- idea sync task


modmuss50:
I will enable/fix the remaining issues in follow up PRs.

PR #1168, opened 2024-09-07
Only validate loom version when mixins are to be remapped with TinyRemapper

mezz:
I think there is an inversion in the logic introduced in https://github.com/FabricMC/fabric-loom/pull/980

The comments for `MixinRemapType.STATIC` say "Jar does not use refmaps, so will be remapped by tiny-remapper"
And the comments for `validateLoomVersion` say "This is only done for jars with tiny-remapper remapped mixins."

So I think the logic should be checking `== STATIC` instead of `!= STATIC`

### DIFF
--- a/src/test/groovy/net/fabricmc/loom/test/unit/ArtifactMetadataTest.groovy
+++ b/src/test/groovy/net/fabricmc/loom/test/unit/ArtifactMetadataTest.groovy
@@ -117,8 +117,8 @@ class ArtifactMetadataTest extends Specification {
 		MIXIN       | ["hello.json": "{}"]       // None Mod jar
 		MIXIN       | ["fabric.mod.json": "{}"]  // Fabric mod without manfiest file
 		MIXIN       | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Remap", "true")]              // Fabric mod without remap type entry
-		MIXIN       | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "mixin")] 	// Fabric mod with remap type entry "mixin"
-		STATIC      | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "static")]	// Fabric mod with remap type entry "static"
+		MIXIN       | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "mixin")]  // Fabric mod with remap type entry "mixin"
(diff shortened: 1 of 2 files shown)


mezz:
I was able to add new failing tests, and confirm that now they pass with the PR changes

I also had to change two tests in order to match the new code changes,
`"Invalid loom version"` and `"Valid loom version"`
used to check for remap type "mixin" and now they test for type "static"

src/test/groovy/net/fabricmc/loom/test/unit/ArtifactMetadataTest.groovy
```diff
@@ -158,9 +159,35 @@ class ArtifactMetadataTest extends Specification {
...
+		// Valid
+		"1.4"       | "1.0.1"
+		"1.4"       | "1.0.99"
+		"1.4"       | "1.4"
+		"1.4"       | "1.4.0"
+		"1.4"       | "1.4.1"
+		"1.4"       | "1.4.99"
+		"1.4"       | "1.4.local"
+		"1.5"		| "1.4.99"
+		"2.0"		| "1.4.99"
```
Juuxel: This should be all spaces, no tabs since it's for alignment:

src/test/groovy/net/fabricmc/loom/test/unit/ArtifactMetadataTest.groovy
```diff
@@ -116,14 +116,15 @@ class ArtifactMetadataTest extends Specification {
 		type | entries
 		MIXIN       | ["hello.json": "{}"] 												// None Mod jar
 		MIXIN       | ["fabric.mod.json": "{}"] 										// Fabric mod without manfiest file
-		MIXIN       | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "mixin")] 	// Fabric mod without remap type entry
-		STATIC  	| ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "static")]	// Fabric mod opt-in
+		MIXIN       | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Remap", "true")]              // Fabric mod without remap type entry
+		MIXIN       | ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "mixin")] 	// Fabric mod with remap type entry "mixin"
+		STATIC  	| ["fabric.mod.json": "{}", "META-INF/MANIFEST.MF": manifest("Fabric-Loom-Mixin-Remap-Type", "static")]	// Fabric mod with remap type entry "static"
```
Juuxel: Same here

modmuss50 (approved):

PR #1169, opened 2024-09-08
Gen sources config caching support

modmuss50:
I think I left the best till last 😓


src/main/java/net/fabricmc/loom/configuration/providers/minecraft/mapped/AbstractMappedMinecraftProvider.java
```diff
@@ -99,9 +106,29 @@ public List<MinecraftJar> provide(ProvideContext context) throws Exception {
...
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	protected void createBackupJars(List<MinecraftJar> minecraftJars) throws IOException {
+		for (MinecraftJar minecraftJar : minecraftJars) {
+			Files.copy(minecraftJar.getPath(), getBackupJarPath(minecraftJar));
```
Juuxel: Needs `REPLACE_EXISTING`, otherwise it'll crash if for example some jars exist (so `hasBackupJars` returns false) and others don't.

PR #1171, opened 2024-09-12
Fix conflicts caused by `MappingsMerger`

SpaceWalkerRS:
The mappings merger uses a `MappingNsCompleter` visitor to fill any empty spots in the official namespace. According to [this commit](https://github.com/FabricMC/fabric-loom/commit/b2228e3175653beb3d5c3c2c382140e29797bd5c) this was done to fix unobfuscated methods not having parameters or javadocs. However, this solution can lead to conflicts, in particular in pre-1.3 mod environments, where the client and server have separate official namespaces.

To resolve this I implemented a modified version of the `MappingNsCompleter` visitor, that only fills empty spots for un-changed mappings. This will catch any unobfuscated methods as they have the same name in both the intermediary and named namespaces, but it's not perfect as it will also catch any members that are not mapped.

I also had to modify `TinyRemapperHelper` because it did not take empty mapping src names into account. Is there any reason Loom is not using [Tiny Remapper's `TinyUtils.createMappingProvider`](https://github.com/FabricMC/tiny-remapper/blob/6736d5f3fd3daadc560b0a25b683ba8851837c22/src/main/java/net/fabricmc/tinyremapper/TinyUtils.java#L79)?

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
@@ -40,7 +40,6 @@ import org.slf4j.LoggerFactory;
 import net.fabricmc.loom.api.mappings.layered.MappingsNamespace;
 import net.fabricmc.loom.configuration.providers.mappings.IntermediateMappingsService;
 import net.fabricmc.loom.configuration.providers.minecraft.MinecraftProvider;
-import net.fabricmc.mappingio.adapter.MappingNsCompleter;
 import net.fabricmc.mappingio.adapter.MappingSourceNsSwitch;
 import net.fabricmc.mappingio.format.tiny.Tiny2FileReader;
 import net.fabricmc.mappingio.format.tiny.Tiny2FileWriter;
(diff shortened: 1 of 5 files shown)


modmuss50:
> is there any reason Loom is not using [Tiny Remapper's TinyUtils.createMappingProvider](https://github.com/FabricMC/tiny-remapper/blob/6736d5f3fd3daadc560b0a25b683ba8851837c22/src/main/java/net/fabricmc/tinyremapper/TinyUtils.java#L79)?

No, there is no reason other than it hasnt been moved over to it. I will handle this in a seperate PR.

modmuss50 (approved):
Change looks fine, have you done any testing for this? Some automated tests would be great, especially to ensure that it doesnt break again.

SpaceWalkerRS:
> > is there any reason Loom is not using [Tiny Remapper's TinyUtils.createMappingProvider](https://github.com/FabricMC/tiny-remapper/blob/6736d5f3fd3daadc560b0a25b683ba8851837c22/src/main/java/net/fabricmc/tinyremapper/TinyUtils.java#L79)?
> 
> No, there is no reason other than it hasnt been moved over to it. I will handle this in a seperate PR.

Alrighty. I think for this to work https://github.com/FabricMC/tiny-remapper/pull/133 needs to be merged as well.

SpaceWalkerRS:
> Change looks fine, have you done any testing for this? Some automated tests would be great, especially to ensure that it doesnt break again.

I did test a local build to make sure it actually fixed the conflicts, but I didn't test any modern version mod project to make sure that wasn't broken.

As for adding tests, I think it should be fine to make some tests with mock mappings and test the `MappingsMerger` directly, right? As opposed to a full integration test running a gradle build. I'll work on that and see what I can come up with.

SpaceWalkerRS:
I had a go at adding a test for the `MappingsMerger`. Let me know if you have any issues with it.

PR #1181, opened 2024-09-30
Fix `transformJson` mis-encoding.

qyl27:
This is a mis-encoding bug, the InputStreamReader uses the system encoding by default.  
In most of Windows system installations, it was not UTF-8, for Chinese users, it may be GB18030.  
Reading and writing use different encodings, which will cause garbled when injecting access wideners, especially for CJK characters, etc. (Users may write any string in the `description` field and any of `authors` fields in `fabric.mod.json`.)


modmuss50 (approved):
Thanks for spotting this. 👍

PR #1182, opened 2024-09-30
Select jar configuration based on which sides are present

SpaceWalkerRS:
If the user does not specifically set the client-only or server-only configuration, then automatically configure this for versions where only one side exists. This does not happen at all for new versions anymore, but is common for Minecraft versions prior to Release 1.3, and the norm for versions prior to Beta 1.0.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -157,8 +157,16 @@ public abstract class CompileConfiguration implements Runnable {
 
 		var jarConfiguration = extension.getMinecraftJarConfiguration().get();
 
-		if (jarConfiguration == MinecraftJarConfiguration.MERGED && !metadataProvider.getVersionMeta().isVersionOrNewer(Constants.RELEASE_TIME_1_3)) {
-			jarConfiguration = MinecraftJarConfiguration.LEGACY_MERGED;
+		if (jarConfiguration == MinecraftJarConfiguration.MERGED) {
+			// if no configuration is selected by the user, attempt to select one
+			// based on the mc version and which sides are present for it
+			if (!metadataProvider.getVersionMeta().downloads().containsKey("server")) {
+				jarConfiguration = MinecraftJarConfiguration.CLIENT_ONLY;
+			} else if (!metadataProvider.getVersionMeta().downloads().containsKey("client")) {
+				jarConfiguration = MinecraftJarConfiguration.SERVER_ONLY;
+			} else if (!metadataProvider.getVersionMeta().isVersionOrNewer(Constants.RELEASE_TIME_1_3)) {
+				jarConfiguration = MinecraftJarConfiguration.LEGACY_MERGED;
+			}
 		}
 
 		// Provide the vanilla mc jars
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -60,7 +60,6 @@ import net.fabricmc.loom.configuration.processors.MinecraftJarProcessorManager;
 import net.fabricmc.loom.configuration.processors.ModJavadocProcessor;
 import net.fabricmc.loom.configuration.providers.mappings.LayeredMappingsFactory;
 import net.fabricmc.loom.configuration.providers.mappings.MappingConfiguration;
-import net.fabricmc.loom.configuration.providers.minecraft.MinecraftJarConfiguration;
 import net.fabricmc.loom.configuration.providers.minecraft.MinecraftMetadataProvider;
 import net.fabricmc.loom.configuration.providers.minecraft.MinecraftProvider;
 import net.fabricmc.loom.configuration.providers.minecraft.MinecraftSourceSets;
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -157,8 +157,16 @@ private synchronized void setupMinecraft(ConfigContext configContext) throws Exc
 
 		var jarConfiguration = extension.getMinecraftJarConfiguration().get();
 
-		if (jarConfiguration == MinecraftJarConfiguration.MERGED && !metadataProvider.getVersionMeta().isVersionOrNewer(Constants.RELEASE_TIME_1_3)) {
-			jarConfiguration = MinecraftJarConfiguration.LEGACY_MERGED;
+		if (jarConfiguration == MinecraftJarConfiguration.MERGED) {
+			// if no configuration is selected by the user, attempt to select one
+			// based on the mc version and which sides are present for it
+			if (!metadataProvider.getVersionMeta().downloads().containsKey("server")) {
+				jarConfiguration = MinecraftJarConfiguration.CLIENT_ONLY;
```
modmuss50: No, this is wrong. It should set the value in the extension, otherwise all the other places like use this configuration will not be in agreement.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -157,8 +157,16 @@ private synchronized void setupMinecraft(ConfigContext configContext) throws Exc
 
 		var jarConfiguration = extension.getMinecraftJarConfiguration().get();
 
-		if (jarConfiguration == MinecraftJarConfiguration.MERGED && !metadataProvider.getVersionMeta().isVersionOrNewer(Constants.RELEASE_TIME_1_3)) {
-			jarConfiguration = MinecraftJarConfiguration.LEGACY_MERGED;
+		if (jarConfiguration == MinecraftJarConfiguration.MERGED) {
+			// if no configuration is selected by the user, attempt to select one
```
modmuss50: This isnt always stricly true, as a user could have requested a merged jar. I know its unlikely and changing this would be a much bigger change. Overall im not sure how much benefit this provides.

SpaceWalkerRS: How would the user select a merged jar? It's the default and there are API methods only for changing it to client-only or server-only. Regardless, if one side does not exist then the build will fail on the merger configuration anyway.

modmuss50: The user could set the value of LoomGradleExtensionAPI.getMinecraftJarConfiguration(), I think its unlikely so I wouldnt worry too much about it.

> if one side does not exist then the build will fail on the merger configuration anyway.
Yes, thats expected. I dont think its a lot to ask for someone to set this.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -157,8 +157,16 @@ private synchronized void setupMinecraft(ConfigContext configContext) throws Exc
...
+			// based on the mc version and which sides are present for it
+			if (!metadataProvider.getVersionMeta().downloads().containsKey("server")) {
+				jarConfiguration = MinecraftJarConfiguration.CLIENT_ONLY;
+			} else if (!metadataProvider.getVersionMeta().downloads().containsKey("client")) {
+				jarConfiguration = MinecraftJarConfiguration.SERVER_ONLY;
+			} else if (!metadataProvider.getVersionMeta().isVersionOrNewer(Constants.RELEASE_TIME_1_3)) {
+				jarConfiguration = MinecraftJarConfiguration.LEGACY_MERGED;
```
modmuss50: I dont think we should fall back to this, as this assumes that you have 3rd party intermediary mappings.

modmuss50:
Overall im not too keen on this, its impossible to know for sure what the user wants, and its not a lot to ask them to set it manually. If the error messages arent good enough it might be best to just improve thoes.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
SpaceWalkerRS: it is the current behavior already so I kept it

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
SpaceWalkerRS: I see, that's a fair point. It would be possible to leave the value empty instead of setting the Convention to merged, that way you'd know if the user selected a specific configuration or not. 

I agree it's not a big deal to set this. It's a minor inconvenience but this is also a minor change to alleviate that so I thought it was fair.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
modmuss50: Oh, I see yeah that makes sense, as its already requested to be merged. thats fine.

src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
modmuss50: It would likely be best to move all of this logic into the convension as a Provider. I do worry that the metadata might not have been populated if someone requests the value too soon though.

SpaceWalkerRS: currently the MetadataProvider is only available through the MinecraftProvider, it'd require adding that to the extension to use it in the configuration provider

modmuss50: It already is part of the internal extenstion, you have the project, so you can use `LoomGradleExtension.get` to get the internal one.

SpaceWalkerRS: Alright I moved it into the convention. It throws an error if the metadata provider is accessed to soon as well.

modmuss50 (approved):
Great, I like this much more. Thanks.

PR #1197, opened 2024-10-20
fix: make JarNester deterministic

ashhhleyyy:
Sorts the list of jars to nest before adding them to fabric.mod.json to ensure the ordering is deterministic.

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/nesting/JarNester.java
+++ b/src/main/java/net/fabricmc/loom/build/nesting/JarNester.java
@@ -28,6 +28,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


solonovamax:
~~sorting it only if the `isReproducibleFileOrder` may be a desired solution as it avoides sorting when determinism isn't desired~~
from the fabric discord:
> Everything should be reproducible by default, I think this one would be a regression

also, the sort order *can* be different on windows vs unix, as-per the javadocs for `File.compareTo`:
> Compares two abstract pathnames lexicographically. The ordering defined by this method depends upon the underlying system. On UNIX systems, alphabetic case is significant in comparing pathnames; on Microsoft Windows systems it is not.

it may be desireable to `.toLowerCase()` the pathnames first before sorting (also possibly to ensure the input files are absolute and/or only sort using the base file name?)

ashhhleyyy:
I think just using the filename would be okay as using the absolute path makes it dependent on the build environment, and only the filename makes it into the jar. I don't think it's a good idea to lower-case them first, since then two files with different cases will get sorted the same, and the ordering between the two becomes platform dependent again

solonovamax:
> I think just using the filename would be okay as using the absolute path makes it dependent on the build environment, and only the filename makes it into the jar. I don't think it's a good idea to lower-case them first, since then two files with different cases will get sorted the same, and the ordering between the two becomes platform dependent again

if you don't want to lowercase them, then make sure to sort using the `String` representation of the file/filename rather than the `File` representation

ashhhleyyy:
yeah, i meant to push the change that went along with that but forgor, will push when i get home

solonovamax (approved):
LGTM

LambdAurora (approved):

modmuss50 (approved):

PR #1202, opened 2024-10-26
Allow changing the project Directory

Stein-N:
This is more of a personal issue, but it could be helpful for anyone using Fabric-Loom in a multi-loader setup.

With this change, it would be possible to use a centralized folder in the common project as the run/working directory within the RunConfiguration. This way, a separate folder doesn’t need to be used for each mod loader.


lukebemish:
I'm a bit confused on the use for this -- can't you just set the `runDir` to, you know, some run directory in the root project's directory? Via, say, `../run` or whatever? If you _can't_, it seems like the better bet would be to back `runDir` with some sort of `RegularFileProperty` or the like that supports that.

PR #1210, opened 2024-11-08
fix: ignore some very very old kotlin classes to prevent them breaking remapping process

zly2006:
The current kotlin toolchain does not support pre-1.4 compiled classes, so I ignored them.

https://github.com/JetBrains/kotlin/blob/7c539b01f522d7ad458a3c774879277eccf03ffd/libraries/kotlinx-metadata/jvm/src/kotlin/metadata/jvm/KotlinClassMetadata.kt#L427

### DIFF
--- a/src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
+++ b/src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
@@ -62,8 +62,13 @@ class KotlinClassMetadataRemappingAnnotationVisitor(
                     "is using (${KotlinVersion.CURRENT}).",
             )
         }
+        val metadata = KotlinClassMetadata.readLenient(header)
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/fabricmc/loom/kotlin/remapping/KotlinClassMetadataRemappingAnnotationVisitor.kt
```diff
@@ -62,8 +62,13 @@ class KotlinClassMetadataRemappingAnnotationVisitor(
                     "is using (${KotlinVersion.CURRENT}).",
             )
         }
+        val metadata = KotlinClassMetadata.readLenient(header)
+        if (metadata.version.major < 1 || (metadata.version.major == 1 && metadata.version.minor < 4)) {
+            logger.warn("$className is not supported by kotlin metadata remapping (version: ${metadata.version})")
+            return
```
modmuss50, line 69: I think this is going to remove the metadata for these classes, is that what you expected to do? Or did you just want to leave it untouched? If so add an `accept(next)` before the return.

modmuss50 (approved):

PR #1215, opened 2024-11-14
Initial project isolation support

modmuss50:
This PR disables some none essential features to allow Loom to run with project isolation.

What doesnt work with project isolation:
- Applying AWs or TAWs from project dependencies that are mdos.
- Reading any Gradle property.
  - https://github.com/gradle/gradle/issues/28857
  - Or a custom property parser!? surely it cannot be the only solution.
- Sandbox support (doesnt matter, was mostly to help me test stuff)
- Remap jar with other projects mixin AP mappings. (An edge case that I might disable by default anyway)
  - Using TR to remap mixins solves this problem.

The multi mc version test has been rewritten to run with project isolation and configure on demand.


src/test/resources/projects/multi-mc-versions/settings.gradle
```diff
@@ -1,3 +1,14 @@
+plugins {
+	id 'fabric-loom' apply false
+}
+
+gradle.lifecycle.beforeProject {
+	if (it.rootProject == it) return
+
+		apply plugin: 'fabric-loom'
```
Juuxel:
```suggestion
	apply plugin: 'fabric-loom'
```
(to match the next line)

Juuxel (approved):

PR #1225, opened 2024-11-28
Fix getKnownIndyBsms storing a set in a list without sorting

jpenilla:
This caused remapJar to re-run unnecessarily


modmuss50 (approved):
Good spot, Ill merge this into 1.9 (will be released shortly)

modmuss50:
Idk that this build error is about 😟 `java.io.IOException: Failed to load eclipse groovy formatter: java.lang.RuntimeException: org.tukaani.xz.XZFormatException: Input is not in the XZ format` (Clearly not your fault)

PR #1229, opened 2024-12-10
Fix lazily provided layered mappings causing crash

isXander:
## Problem

```groovy
mappings project.provider {
	loom.layered() {
		officialMojangMappings()
	}
}
```

The above is the configuration that caused the following crash.
```
> Failed to setup Minecraft, java.lang.IllegalStateException: Layered mappings have already been evaluated
```

## Why is this behaviour wanted?

This allows people to use providers to define the mappings used by a project, for example, a gradle plugin that wraps Loom might define parchment via a extension property.

```groovy
mappings myExtension.parchmentArtifact.orElse("").map { parchmentArtifact ->
	loom.layered() {
		officialMojangMappings()

		if (!parchmentArtifact.isEmpty()) {
			parchment(parchmentArtifact)
		}
	}
}
```
*excuse the possible invalid groovy syntax, I am used to Kotlin DSL*

## Code analysis
During `setupMinecraft` in `CompileConfiguration`:
```java
// 1
LayeredMappingsFactory.afterEvaluate(configContext);

// 2
final DependencyInfo mappingsDep = DependencyInfo.create(getProject(), Configurations.MAPPINGS);

// 3
final MappingConfiguration mappingConfiguration = MappingConfiguration.create(getProject(), configContext.serviceFactory(), mappingsDep, minecraftProvider);
```
1. LayeredMappings are finalised - `afterEvaluate` method prevents any further configuration being made to the layered spec via `loom.layered() {}`.
2. The dependency for mappings is fetched, this is where the provider is called upon and `loom.layered()` in the buildscript is now evaluated.
3. Loom then applies the mapping that was fetched in step 2

Can you see the problem?

### How did this work before?

Previously, Loom relied on the fact that the buildscript itself is evaluated before `setupMinecraft`. Which meant that if a `loom.layered() {}` was called, it would happen before it was finalised.

Using a provider exposed the problem: the layered spec is finalised before actually *retrieving* the dependency for mappings. The layered spec would finalise and THEN the provider would trigger, trying to call `loom.layered() {}` .

### The fix

The fix is simple, finalise the layered mappings factory AFTER retrieving the dependency. Just swap lines 1 and 2.

```java
// 2
final DependencyInfo mappingsDep = DependencyInfo.create(getProject(), Configurations.MAPPINGS);

// 1
LayeredMappingsFactory.afterEvaluate(configContext);

// 3
final MappingConfiguration mappingConfiguration = MappingConfiguration.create(getProject(), configContext.serviceFactory(), mappingsDep, minecraftProvider);
```

An integration test has been added to test for this.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
@@ -166,10 +166,10 @@ public abstract class CompileConfiguration implements Runnable {
 		extension.setMinecraftProvider(minecraftProvider);
 		minecraftProvider.provide();
 
-		// Created any layered mapping files.
+		final DependencyInfo mappingsDep = DependencyInfo.create(getProject(), Configurations.MAPPINGS);
+		// Created any layered mapping files. (After resolving the mappings dependency)
 		LayeredMappingsFactory.afterEvaluate(configContext);
 
-		final DependencyInfo mappingsDep = DependencyInfo.create(getProject(), Configurations.MAPPINGS);
 		final MappingConfiguration mappingConfiguration = MappingConfiguration.create(getProject(), configContext.serviceFactory(), mappingsDep, minecraftProvider);
 		extension.setMappingConfiguration(mappingConfiguration);
 		mappingConfiguration.applyToProject(getProject(), mappingsDep);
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/configuration/CompileConfiguration.java
```diff
@@ -166,10 +166,10 @@ private synchronized void setupMinecraft(ConfigContext configContext) throws Exc
 		extension.setMinecraftProvider(minecraftProvider);
 		minecraftProvider.provide();
 
-		// Created any layered mapping files.
+		final DependencyInfo mappingsDep = DependencyInfo.create(getProject(), Configurations.MAPPINGS);
+		// Created any layered mapping files. (After resolving the mappings dependency)
 		LayeredMappingsFactory.afterEvaluate(configContext);
```
modmuss50, line 171: This doesnt seem right to me, LayeredMappingsFactory.afterEvaluate is what generates the layered mapping, but calling DependencyInfo before this has the chance of resolving the mappings before they have been created. I will have a look at this, the real fix is likely a bit more complex.

isXander:
Closed in favour of #1237

PR #1230, opened 2024-12-11
create backup jars for legacy merged mapped minecraft providers

SpaceWalkerRS:
this fixes crashes when e.g. running `genSources` for legacy mc versions


modmuss50:
Ive pushed some changes that to try and clean up this some more, all of this code could really do with a good refactor to better support all of the current use cases. There wasnt a lot wrong with your changes, I just worry we are layering special case ontop of special case here, with duplciated code all over the place.

PR #1231, opened 2024-12-12
Run library processor for server-only projects for Minecraft versions without  bundle metadata

SpaceWalkerRS:
This allows custom library processors to be run on server-only projects for older versions.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftLibraryProvider.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftLibraryProvider.java
@@ -37,7 +37,6 @@ import org.gradle.api.artifacts.ModuleDependency;
 import org.gradle.api.provider.Provider;
 
 import net.fabricmc.loom.LoomGradleExtension;
-import net.fabricmc.loom.configuration.providers.BundleMetadata;
 import net.fabricmc.loom.configuration.providers.minecraft.library.Library;
 import net.fabricmc.loom.configuration.providers.minecraft.library.LibraryContext;
 import net.fabricmc.loom.configuration.providers.minecraft.library.LibraryProcessorManager;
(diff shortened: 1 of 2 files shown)


src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftLibraryProvider.java
```diff
@@ -108,13 +108,7 @@ private void provideClientLibraries() {
...
 
 	private void provideServerLibraries() {
-		final BundleMetadata serverBundleMetadata = minecraftProvider.getServerBundleMetadata();
-
-		if (serverBundleMetadata == null) {
-			return;
-		}
-
-		final List<Library> libraries = MinecraftLibraryHelper.getServerLibraries(serverBundleMetadata);
+		final List<Library> libraries = MinecraftLibraryHelper.getServerLibraries(minecraftProvider.getServerBundleMetadata());
```
modmuss50: Is this not always going to be empty for the older versions without the bundle metadata, or am I missing something?

SpaceWalkerRS: It is, but custom library processors can add their own libraries.

modmuss50: I dont think this is the correct fix, and im not sure it actually fixes the issue. The current library processor API doesnt really expect you to be able to add libraries, its very much for changing the vanilla ones. What is your usecase? Can you not just add the library in the gradle way?

SpaceWalkerRS: I have a plugin that upgrades or adds libraries depending on the MC version. For example, MC didn't depend on Log4j until some 1.7 snapshot, so for earlier versions it adds that dependencies, for later versions it upgrades it to a newer version.

modmuss50: Ok, I see yeah. Sorry yes that is possible.

This change is correct, but im not super happy with how it changes getServerLibraries. I think you should change it all in `provideServerLibraries`. Something as simple as:

```java
private void provideServerLibraries() {
	final BundleMetadata serverBundleMetadata = minecraftProvider.getServerBundleMetadata();
	final List<Library> libraries = serverBundleMetadata != null ? MinecraftLibraryHelper.getServerLibraries(serverBundleMetadata) : Collections.emptyList();
	final List<Library> processLibraries = processLibraries(libraries);
	processLibraries.forEach(this::applyServerLibrary);
}
```

SpaceWalkerRS: alrighty, done

modmuss50 (approved):

PR #1235, opened 2024-12-18
Make LineNumberRemapper stop trying to guess line numbers

tranquillity-codes:
Usually the guesses were wrong, so instead let's not generate a LineNumberTable for lines that lack a mapping in the linemap. This makes the behavior on decompiler bugs/issues more predictable.


modmuss50 (approved):
This makes much more sense. 👍

PR #1244, opened 2025-01-02
Add support for running the production client with the tracy profiler.

modmuss50:
There is no default tracy-capture path, as there is no standard way of building/installing it.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/prod/ClientProductionRunTask.java
+++ b/src/main/java/net/fabricmc/loom/task/prod/ClientProductionRunTask.java
@@ -25,13 +25,17 @@
 package net.fabricmc.loom.task.prod;
 
 import java.io.File;
+import java.io.IOException;
 
 import javax.inject.Inject;
 
+import org.gradle.api.Action;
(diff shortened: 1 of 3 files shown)


modmuss50:
On MacOS the changes in this PR work as expected, but on Linux where I want to be able to use this the most the game stalls when shutting down. I believe this is caused by: https://github.com/wolfpld/tracy/issues/943 or at least something very similar, as I dont have symbols for Mojang's own natives its hard to tell. I also don't know if Mojang's own natives are built with `TRACY_HAS_CALLSTACK=0` or not, I expect they are. The ones I created my self stall when launching here:

https://github.com/wolfpld/tracy/blob/5d542dc09f3d9378d005092a4ad446bd405f819a/public/client/TracyProfiler.cpp#L2047-L2049

 so I can see why Mojang might have disabled it, leading to the linked issue. I dont want to add a feature to loom that requires custom tracy natives, so until this is fixed in tracy, released and Mojang updates I dont think we can do this.

modmuss50:
I realised the workaround in in the tracy issue only needs to be applied to the capture server, so this can be fixed without modifying the game :)

PR #1245, opened 2025-01-05
Update to Mapping-IO 0.7.0 & Tiny Remapper 0.11

### LINKED ISSUE #1239: Mixin count wildcard * not working because of remapping issue
using the `*` wildcard in mixin method targets doesn't work, because the remapping step always adds some method args to the method name, even when not specifying them

an example with mc 1.21.4 and yarn:
having this mixin, I would expect it to inject into all 3 available constructors
```
@Mixin(PersistentProjectileEntity.class)
public class ArrowMixin {
    @Inject(method = "<init>*", at = @At("TAIL"))
    private void atEnd(CallbackInfo ci) {}
}
```
instead it only injects into one of them, because the generated refmap removed the `*` and added method args
```
{
  "mappings": {
    "com/fayer3/test/mixin/client/ProjectileMixin": {
      "<init>*": "Lnet/minecraft/class_1665;<init>(Lnet/minecraft/class_1299;Lnet/minecraft/class_1937;)V"
    }
  },
  "data": {
    "named:intermediary": {
      "com/fayer3/test/mixin/client/ProjectileMixin": {
        "<init>*": "Lnet/minecraft/class_1665;<init>(Lnet/minecraft/class_1299;Lnet/minecraft/class_1937;)V"
      }
    }
  }
}
```

modmuss50:


NebelNidas:
You'll need to revert the outer class name propagation changes until MIO 0.8, when the tree-API alternative is ready, where you can specify the source namespace to use.

PR #1252, opened 2025-01-22
Improvements to test and IDE runs

### LINKED ISSUE #1233: Loom 1.8+ does not generate a `build\datagen` directory, causing the 'Run Datagen' run configuration to be broken
Loom 1.8+ no longer generates a `build\datagen` directory. This leaves 'Run Datagen' run configuration broken till you manually create that directory.

This behavior is not seen in Loom 1.7.

Steps to reproduce:
* Generate a fresh template project from https://fabricmc.net/develop/template/ with 'Data Generation' enabled.
* Open the project, wait for Gradle sync to finish, then restart the IDE (in my case IDEA) to load the run configurations.
* The 'Run Datagen' configuration will appear broken.

image

### LINKED ISSUE #1250: Set system property fabric.client.gametest.testModResourcesPath by default in client gametest configuration

modmuss50:
Closes #1250
Fixes #1233

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiTesting.java
+++ b/src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiTesting.java
@@ -65,7 +65,6 @@ public abstract class FabricApiTesting extends FabricApiAbstractSourceSet {
 	void configureTests(Action<GameTestSettings> action) {
 		final LoomGradleExtension extension = LoomGradleExtension.get(getProject());
 		final TaskContainer tasks = getProject().getTasks();
-		final SourceSet mainSourceSet = SourceSetHelper.getMainSourceSet(getProject());
 
 		GameTestSettings settings = getProject().getObjects().newInstance(GameTestSettings.class);
 		settings.getCreateSourceSet().convention(false);
@@ -77,8 +76,12 @@ public abstract class FabricApiTesting extends FabricApiAbstractSourceSet {
 
 		action.execute(settings);
 
+		final SourceSet testSourceSet;
+
 		if (settings.getCreateSourceSet().get()) {
-			configureSourceSet(settings.getModId(), true);
+			testSourceSet = configureSourceSet(settings.getModId(), true);
+		} else {
+			testSourceSet = SourceSetHelper.getMainSourceSet(getProject());
 		}
 
 		Consumer<RunConfigSettings> configureBase = run -> {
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiTesting.java
```diff
@@ -94,10 +99,23 @@ void configureTests(Action<GameTestSettings> action) {
 		}
 
 		if (settings.getEnableClientGameTests().get()) {
+			// Not ideal as there may be multiple resources directories, if this isnt correct the mod will need to override this.
```
Earthcomputer, line 105: Hmm I think this is more likely to be wrong than right. I imagine most people will put their client gametest as a separate mod in a separate source set.

src/main/java/net/fabricmc/loom/configuration/ide/idea/IdeaSyncTask.java
```diff
@@ -110,6 +110,8 @@ private List<IntelijRunConfig> getRunConfigs() throws IOException {
 			irc.getExcludedLibraryPaths().set(excludedLibraryPaths);
 			irc.getLaunchFile().set(runConfigFile);
 			configs.add(irc);
+
+			settings.makeRunDir();
```
Earthcomputer, line 114: Does this need applying to the other IDE sync tasks?

Also, it doesn't have to be in this PR but it would be good to support pre-run Gradle tasks. I know it's not ideal to run a gradle task before an IDE run but I think it is the least bad of all the ways to clean the run directory for client gametests.

Earthcomputer (changes requested):

src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiTesting.java
modmuss50: Fixed, thanks it now uses the generated sourceset.

src/main/java/net/fabricmc/loom/configuration/ide/idea/IdeaSyncTask.java
modmuss50: No, this was something I removed/broke in https://github.com/FabricMC/fabric-loom/pull/1166

I strongly oppose running Gradle before IDE runs, its not easy to in intelij and likely impossible in vscode/eclipse, and takes ages. We already clear the run dir before using one of the Gradle tasks, if your tests depend on that disable the IDE runs and solely rely on Gradle, it will be quicke, better supported and doesnt require any gross hacks to make it work.

modmuss50: As dissussed we will do something about this, likely with a new entrypoint but I will do it as its own thing.

PR #1257, opened 2025-01-26
Add await() calls to tasks using WorkQueues

Roundaround:
In particular, added it to AbstractRemapJarTask and ValidateMixinNameTask.

Without the call to await(), the tasks declare to Gradle that they're complete immediately, not waiting for the work to actually be done. This means that even after declaring a dependency on the task, any following tasks have the potential to start executing before the output file is ready. _Most of the time_ this is probably fine because they happen quickly but in my case I kept running into issues with file locks on the JARs output by the remap tasks when building my own custom plugin that sits on top of Loom.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/AbstractRemapJarTask.java
+++ b/src/main/java/net/fabricmc/loom/task/AbstractRemapJarTask.java
@@ -168,6 +168,8 @@ public abstract class AbstractRemapJarTask extends Jar {
 
 			action.execute(params);
 		});
+
+		workQueue.await();
 	}
 
 	protected abstract Provider<? extends ClientEntriesService.Options> getClientOnlyEntriesOptionsProvider(SourceSet clientSourceSet);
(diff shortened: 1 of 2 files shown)


jpenilla:
> Without the call to await(), the tasks declare to Gradle that they're complete immediately, not waiting for the work to actually be done. This means that even after declaring a dependency on the task, any following tasks have the potential to start executing before the output file is ready.

This is not the case, the work queue is created from a worker executor injected to the task, so Gradle will implicitly wait on any submitted work before the task is counted as complete, including propagating failures. An explicit call to await is only needed if you have further logic in your task that needs to run after the submitted work is done, which is not the case in either of the tasks you're editing.

modmuss50:
Yes, what @jpenilla said is correct. Adding the .await will make them run synchronously removing the whole benefit of using the work queue in the first place (as these tasks all have just 1 job).

If you are having issues with locked files there is either something not releasing the lock or a more complex issue at play, I'd be happy discuss this with you to help you debug it. Either reach out to me on discord or in an issue. Ill close this PR as its not the correct fix.

PR #1262, opened 2025-02-09
Add general purpose download task

modmuss50:
Loom has a robust download system, might as well expose it as a public task for people to use.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/DownloadTask.java
+++ b/src/main/java/net/fabricmc/loom/task/DownloadTask.java
@@ -1,0 +1,145 @@
+/*
+ * This file is part of fabric-loom, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2025 FabricMC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


src/main/java/net/fabricmc/loom/task/DownloadTask.java
```diff
@@ -0,0 +1,145 @@
...
+	@Optional
+	@Input
+	public abstract Property<String> getSha1();
+
+	/**
+	 * The maximum age of the downloaded file in days. When not provided the downloaded file will never be considered stale.
+	 */
+	@Optional
+	@Input
+	public abstract Property<Integer> getMaxAge();
```
Juuxel: Duration implements Serializable so you could probably expose them too (would allow for finer control)

modmuss50: Ah nice, ill change that. I should have checked before assuming.

modmuss50: Looks good, also added a basic test for it. It doesnt check it works (the unit tests do that), just makes sure gradle is happy with it.

PR #1271, opened 2025-03-04
RemapJarTask: Delete output if it exists to prevent ghost files

### LINKED ISSUE #1270: RemapJarTask leaves deleted files in the jar
To reproduce from a template:
1. Create `src/main/resources/test.txt`
2. Run `gradlew assemble` (or `build` or `remapJar`)
3. Delete `src/main/resources/test.txt`
4. Run the same Gradle command again
5. The deleted file is still in the jar!

This bug is caused by [this change](https://github.com/FabricMC/fabric-loom/commit/e387514ff70d9f6b6951e60f43bf30e130be460d#diff-347aa7ef5e0cb5171e40d8af7ec1594d858985461500858da624d62148f3e6f7R130) in e387514, which disabled the default `Jar` task behaviour for `RemapJarTask` due to changed task action ordering.

Unfortunately, the `copy` method is also responsible for deleting the old jar, which means that now TR just adds *new and changed* files into the jar but leaves all deleted files intact.

Juuxel:
Fixes #1270. This changed isn't necessary for RemapSourcesJarTask as SourceRemapperService already deletes the output.


modmuss50 (approved):
Nice spot.

Issue #1164, opened 2024-09-02
Shorter/Improved remapped names


solonovamax:
Currently, when mods are remapped by loom, they result in names that look like the following:
```
loom_mappings_1_19_4_layered_hash_40359_v2.net.fabricmc.fabric-api:fabric-data-generation-api-v1:11.5.2+26f09d40f4
```
This name is quite long and unwieldy, and is annoying to look at in your IDE when attempting to find the sources for a particular library.

I would like to propose that instead when remapping, the `loom_mappings_1_19_4_layered_hash_40359_v2` bit can be replaced with, for example, a 6 character MD5 hash of the minecraft version & the mappings used (ofc properly accounting for layered mappings). eg. minecraft 1.21.1 using yarn mappings version `1.21.1+build.3` might result in the following remapped mod name:
```
remapped_cd6b1a.net.fabricmc.fabric-api:fabric-data-generation-api-v1:11.5.2+26f09d40f4
```
`cd6b1a` is derived from the first 6 characters of the MD5 hash of `minecraft-1.21.1:net.fabricmc-yarn-1.21.1+build.3`. `minecraft-[version]` is used to indicate the minecraft version, and `[mappings group]-[mappings module]-[mappings version]` is used for the mappings used. They are concatenated with a `:` to separate them.

another alternative to prefixing the hash in front of the group id would be to prefix it in front of the version:
```
net.fabricmc.fabric-api:fabric-data-generation-api-v1:remapped-cd6b1a.11.5.2+26f09d40f4
```
this would keep some of the more important information at the front, such as the group id and module name.

modmuss50:
Yes, I I do think we can improve this. We do need to make sure that the group is diffrent, it could maybe even be something as small as.

`abc123.net.fabricmc.fabric-api:fabric-data-generation-api-v1:11.5.2+26f09d40f4` 🤔

solonovamax:
> @modmuss50
> Yes, I I do think we can improve this. We do need to make sure that the group is diffrent, it could maybe even be something as small as.
> 
> `abc123.net.fabricmc.fabric-api:fabric-data-generation-api-v1:11.5.2+26f09d40f4` 🤔

why specifically the group id?

solonovamax:
also tbh I was originally gonna put smth like that but I wasn't sure if it was clear enough where the hash was coming from

solonovamax:
[...]

PR #1273, opened 2025-03-04
Refmap inlining

modmuss50:


modmuss50:
I think I want to fix: https://github.com/FabricMC/fabric-loom/issues/1164 before going further with this, as right now it wont regenerate the remapped deps if we enabled this. The remapped deps need a hash that includes the layered mappings as well as the settings used to remap them (this being the only setting we will have to start with)

modmuss50:
Not 100% done, but good enough to have behind an option.

PR #1276, opened 2025-03-14
Support putting run configs in folders

SquidDev:
- Remove unused `RunConfig.genRuns`.
 - Rewrite templating to do a single find-and-replace, rather than a chain of them.
 - Add a new `ideConfigFolder` option, which sets the `folderName` attribute on the IDEA config file.

IntellJ Idea's run config drop down. There is a folder called "Fabric", where which contains a list of runs (e.g. client, server).

I wasn't quite sure the best way to expose settings from `RunConfigSettings`, as it's currently a bit ad-hoc. For now, I've gone with the simplest option (a `Property<String>` getter), but happy to change that if preferred!

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
@@ -38,10 +38,12 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import groovy.xml.XmlUtil;
 import org.gradle.api.JavaVersion;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.ModuleVersionIdentifier;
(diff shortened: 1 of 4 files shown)


modmuss50:
Looks like https://github.com/FabricMC/fabric-loom/blob/dev/1.10/src/test/groovy/net/fabricmc/loom/test/unit/IdeaClasspathModificationsTest.groovy is now failing

SquidDev:
I did have a think about replacing `IdeaClasspathModificationsTest.fromDummy` with a call to `RunConfig.fromDummy`. *Technically* we can get away passing a null `Project` to it, but it felt a bit risky, so have left as-is for now.

modmuss50:
> I did have a think about replacing `IdeaClasspathModificationsTest.fromDummy` with a call to `RunConfig.fromDummy`. _Technically_ we can get away passing a null `Project` to it, but it felt a bit risky, so have left as-is for now.

I'd leave it for now, the test coverage (and testability) of all of this could be improved vastly, but there is no point doing it as part of this PR. Ill take a closer look at the PR shortly and get it merged into the 1.11 branch.

modmuss50 (approved):

PR #1283, opened 2025-04-07
Add exclusive maven repos for "net.fabricmc" and "com.mojang"

modmuss50:
Force all "net.fabricmc" maven artifacts to come from the Fabric maven, and likewise for the Mojang repo. This should be a little bit more secure, and also faster.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
@@ -68,6 +68,18 @@ public class LoomRepositoryPlugin implements Plugin<PluginAware> {
 			repo.setUrl(MirrorUtil.getFabricRepository(target));
 		});
 
+		repositories.exclusiveContent(exclusiveContent -> {
+			exclusiveContent.forRepositories(repositories.maven(repo -> {
+				repo.setName("FabricExclusive");
+				repo.setUrl(MirrorUtil.getFabricRepository(target));
+			}));
+
+			exclusiveContent.filter(filter -> {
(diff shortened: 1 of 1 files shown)


IMB11:
Option to disable this? Eg; using experimental versions of fapi published to maven local?

src/main/java/net/fabricmc/loom/LoomRepositoryPlugin.java
```diff
@@ -68,6 +68,18 @@ private void declareRepositories(RepositoryHandler repositories, LoomFiles files
...
 
+		repositories.exclusiveContent(exclusiveContent -> {
+			exclusiveContent.forRepositories(repositories.maven(repo -> {
+				repo.setName("FabricExclusive");
+				repo.setUrl(MirrorUtil.getFabricRepository(target));
+			}));
+
+			exclusiveContent.filter(filter -> {
+				filter.includeGroup("net.fabricmc");
+				filter.includeGroupByRegex("net.fabricmc.*");
```
Juuxel, line 79: Is the first `includeGroup` even needed if you have this? (`net.fabricmc` definitely matches the regex `net.fabricmc.*`)

modmuss50: Ill check.

modmuss50:
Humm, this also has the silly behaviour of breaking `mavenLocal()` some of the tests depend on this, and is very useful for testing. Im not so sure this is something we want.

modmuss50:
Yeah, going to close this as it breaks more than it solves.

PR #1288, opened 2025-04-08
Don't apply interface injection to the common jar when a mod dep is client only.

### LINKED ISSUE #1285: Mods providing interface injections on client-only sourceset prevent building
When using split sources, depending on a mod in the client sourceset, like this
```kt
modClientImplementation("mod.with.injected.interfaces")
```
these interface injections apply to the main sourceset. Because the interfaces themselves are not available on the main sourceset, this prevents compilation.

modmuss50:
Fixes #1285

Needs proper testing to make sure I havent broken anything.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
@@ -99,10 +99,14 @@ public abstract class InterfaceInjectionProcessor implements MinecraftJarProcess
 			return null;
 		}
 
-		return new Spec(injectedInterfaces);
+		Set<String> clientOnlyModIds = context.modDependenciesCompileRuntimeClient().stream()
+				.map(FabricModJson::getId)
+				.collect(Collectors.toSet());
+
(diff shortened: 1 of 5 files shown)


src/main/java/net/fabricmc/loom/configuration/ifaceinject/InterfaceInjectionProcessor.java
```diff
@@ -114,6 +118,11 @@ public void processJar(Path jar, Spec spec, ProcessorContext context) throws IOE
 
 		try (LazyCloseable<TinyRemapper> tinyRemapper = context.createRemapper(MappingsNamespace.INTERMEDIARY, MappingsNamespace.NAMED)) {
 			final List<InjectedInterface> remappedInjectedInterfaces = spec.injectedInterfaces().stream()
+					.filter(injectedInterface -> {
+						return context.includesClient() // The client jar depends on the server, so always apply all to it
+								|| context.isMerged() // Always apply all to merged jars
```
modmuss50: Note to self, this line isnt needed as the merged jar always includes the client.

isXander (approved):
Can't see any immediate problems with this implementation!

PR #1304, opened 2025-05-06
Rewrite Checksum util class as the old one was annoying me.

modmuss50:


src/main/java/net/fabricmc/loom/configuration/providers/minecraft/verify/MinecraftJarVerification.java
```diff
@@ -86,9 +85,9 @@ private boolean isValidKnownVersion(Path path, String version, KnownJarType type
 		}
 
 		LOGGER.info("Found executed hash ({}) for known version: {}", expectedHash, version);
-		String hash = Checksum.sha256Hex(Files.readAllBytes(path));
+		Checksum.Result hash = Checksum.of(path).sha256();
```
Toffikk, line 88: shouldn’t this be `.sha256().hex()`?

modmuss50: No, the Checksum.Result has a matchesStr method that will match against the hex string.

PR #1310, opened 2025-05-16
Unpick V3

modmuss50:


src/main/java/net/fabricmc/loom/task/service/UnpickService.java
```diff
@@ -108,73 +132,101 @@ public UnpickService(Options options, ServiceFactory serviceFactory) {
...
+			ClassWriter writer = new ClassWriter(ClassWriter.COMPUTE_MAXS);
+			classNode.accept(writer);
+
+			Files.write(output, writer.toByteArray());
+		}
 	}
 
-	public abstract static class UnpickAction implements WorkAction<UnpickParams> {
-		private static final Logger LOGGER = LoggerFactory.getLogger(UnpickAction.class);
+	private static class ZipFsClassResolver extends ChainClassResolver implements Closeable {
```
Earthcomputer: `ChainClassResolver` is an impl class, I think it's better to have this class implement `IClassResolver` directly and handle one zip file system, then `create` can return an `IClassResolver` by using `IClassResolver.chain`.

modmuss50: Good spot, now fixed. Thanks.

Earthcomputer (approved):

PR #1323, opened 2025-06-18
fix: Allow Parchment to map classes marked as `@DontObfuscate`

### LINKED ISSUE #1086: with layered parchment mappings, some classes don't get their method parameters mapped
not sure if that is the same issue as #559 , but when using layered mappings with mojang + parchment the methods in some classes don't get their mapped parameters.

some I noticed are the following
- com.mojang.blaze3d.platform.GlStatemanager 
- com.mojang.blaze3d.systems.RenderSystem
- com.mojang.blaze3d.platform.TextureUtil

from that it seems to happen with all classes that have the `@DontObfuscate` annotation

with just yarn, or layered with yarn they are mapped correctly with the yarn mappings

in the parchment mappings they should be mapped https://github.com/ParchmentMC/Parchment/blob/versions/1.20.x/data/com/mojang/blaze3d/platform/GlStateManager.mapping

mapping setup: 
```
mappings loom.layered() {
	officialMojangMappings()
	parchment("org.parchmentmc.data:parchment-1.20.4:2024.02.25@zip")
}
```

using loom 1.6-SNAPSHOT

sylv256:
Fixes #1086

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentMappingLayer.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentMappingLayer.java
@@ -30,57 +30,66 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
-import net.fabricmc.loom.api.mappings.layered.MappingLayer;
-import net.fabricmc.loom.api.mappings.layered.MappingsNamespace;
-import net.fabricmc.loom.util.ZipUtils;
-import net.fabricmc.mappingio.MappingVisitor;
-import net.fabricmc.mappingio.tree.MappingTree;
-
-import net.fabricmc.mappingio.tree.VisitableMappingTree;
-
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public record ParchmentMappingLayer(Path parchmentFile, boolean removePrefix) implements MappingLayer {
+import net.fabricmc.loom.api.mappings.layered.MappingLayer;
+import net.fabricmc.loom.api.mappings.layered.MappingsNamespace;
+import net.fabricmc.loom.util.ZipUtils;
+import net.fabricmc.mappingio.MappingVisitor;
+import net.fabricmc.mappingio.tree.MappingTree;
+import net.fabricmc.mappingio.tree.VisitableMappingTree;
+
+public record ParchmentMappingLayer(
+		Path parchmentFile,
+		boolean removePrefix
+) implements MappingLayer {
 	private static final String PARCHMENT_DATA_FILE_NAME = "parchment.json";
-	private static final Logger LOGGER = LoggerFactory.getLogger(ParchmentMappingLayer.class);
+	private static final Logger LOGGER = LoggerFactory.getLogger(
+			ParchmentMappingLayer.class);
 
 	@Override
 	public void visit(VisitableMappingTree mappingTree) throws IOException {
 		ParchmentTreeV1 parchmentData = getParchmentData();
-		
+
 		// Hack to allow classes marked @DontObfuscate to be mapped
 		Stats stats = new Stats();
-		assert parchmentData.classes() != null;
-		for (ParchmentTreeV1.Class clazz : parchmentData.classes()) {
-			if (mappingTree.getClass(clazz.name()) == null) {
-				mappingTree.addClass(new ClassEntry(mappingTree, clazz));
-				stats.unobfuscatedCount++;
+
+		if (parchmentData.classes() != null) {
+			for (ParchmentTreeV1.Class clazz : parchmentData.classes()) {
+				if (mappingTree.getClass(clazz.name()) == null) {
+					mappingTree.addClass(new ClassEntry(mappingTree, clazz));
+					stats.unobfuscatedCount++;
+				}
 			}
 		}
-		LOGGER.info(
-				"Remapped {} unobfuscated classes with Parchment",
-				stats.unobfuscatedCount
-		);
+
+		LOGGER.info("Remapped {} unobfuscated classes with Parchment", stats.unobfuscatedCount);
 
 		MappingVisitor mappingVisitor = mappingTree;
+
 		if (removePrefix()) {
-			mappingVisitor = new ParchmentPrefixStripingMappingVisitor(mappingTree);
+			mappingVisitor = new ParchmentPrefixStripingMappingVisitor(
+					mappingTree);
 		}
 
 		parchmentData.visit(mappingVisitor, MappingsNamespace.NAMED.toString());
 	}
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentMappingLayer.java
```diff
@@ -42,11 +54,405 @@ public void visit(MappingVisitor mappingVisitor) throws IOException {
...
+			for (ParchmentTreeV1.Class clazz : parchmentData.classes()) {
+				if (tree.getClass(clazz.name()) == null) {
+					tree.addClass(new ClassEntry(tree, clazz));
+					stats.unobfuscatedCount++;
+				}
+			}
+		} else if (mappingVisitor instanceof ForwardingMappingVisitor visitor) {
+			try {
+				Field field = ForwardingMappingVisitor.class.getDeclaredField("next");
+				field.setAccessible(true); // spooky!
```
sylv256: I'd rather not do this, but that would be a PR to Tony Remapper.

edit: I think I meant MappingIO. I don't think Loom really uses Tiny Remapper anymore.

sylv256:
The main reason this isn't generalized to all layered mappings is because Parchment is the only mapping guaranteed to only map parameters and javadoc. It's still considered illegal to remap anything else in unobfuscated classes.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentMappingLayer.java
modmuss50: Hey, this looks like a good attempt however this really shouldnt be needed, mapping IO should have everything needed. What exactally is the issue you are trying to solve there?

Loom does very much use tiny remapper, just not for this kind of stuff.

sylv256: I need to go up the chain of MappingVisitors, but ForwardingMappingVisitor doesn't allow me to see the next MappingVisitor. I could go and make a quick MappingIO PR if that's necessary.

modmuss50: Would casing the provided mapping visitor a `MemoryMappingTree` help? Is still a hack but much easier than this?

modmuss50: Oh, I see maybe not. Ill need to take a proper look at the problem here.

sylv256: Sometimes Parchment uses its `ParchmentPrefixRemoverMappingVisitor` forwarded to the original `MemoryMappingTree`, so I need to recursively go up the forwarding visitor chain.

sylv256:
Before this gets merged, I believe I made an error. Empty Javadocs aren't excluded.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/parchment/ParchmentMappingLayer.java
sylv256: I'm going to make a PR to MappingIO so I don't have to do reflection.

sylv256: I removed the reflection and cursedness at the cost of a small refactor.

sylv256:
oops forgot to run checkstyle

sylv256:
I wasn't able to run the Gradle tests on my computer because it takes more than 32 GiB to run all of them (thanks Gradle), so I guess I'm relying on GH Actions instead.

modmuss50:
> I wasn't able to run the Gradle tests on my computer because it takes more than 32 GiB to run all of them (thanks Gradle), so I guess I'm relying on GH Actions instead.

Yes, dont try to run them locally, just run the single tests that you care about via the IDE and let GHA do the rest.

modmuss50:
Can you expand the unit tests to cover this in ParchmentMappingLayerTest? Im struggling to reproduce it.

I added the following to the "Read parchment mappings" test and it seems to work fine, am I missing something or not testing the right method/class?

```java
reorderedMappings.getClass("net/minecraft/class_4493")
		.getMethod("_activeTexture", "(I)V", 0)
		.getArg(0, 0, null)
		.getDstName(0) == "pTextureIn"
```

sylv256:
> I added the following to the "Read parchment mappings" test and it seems to work fine, am I missing something or not testing the right method/class?

As long as the class is marked @DontObfuscate. You should probably also include a Blaze3D class just in case.

sylv256:
I realize what's happening. We need to be testing 1.17 mappings or higher, otherwise it's obfuscated.

modmuss50:
Ah, thanks of course. Ill look at adding some unit tests for this as I want to make sure I understand the problem first.

sylv256:
This worked for me. It doesn't work with `reorderedMappings` because for whatever reason that strips `GlStateManager` of its method mappings except for its initializers.
```groovy
mappings.getClass("com/mojang/blaze3d/platform/GlStateManager")
	.getMethod("_activeTexture", "(I)V")
	.getArg(0, 0, null)
	.getSrcName() == "pTextureIn"
```

I posted a basic patch for it in the Fabric Discord.

modmuss50:
Ok, I think I understand the problem, ive been through what we do to the mappings step by step and simplified it down to a trivial example:

We first read in the intermediary so the mappings look like:
```
net/minecraft/class_123
```

We then read in the mojang mappings, so the mappings look as follows

```
intermediary				named
net/minecraft/class_123			net/minecraft/ExampleObfClass
					net/minecraft/NoneObfClass
```

To read parchment we switch the source NS:
```
named					intermediary
net/minecraft/ExampleObfClass		net/minecraft/class_123		
net/minecraft/NoneObfClass
```

At this point all is good, we have read intermediary and all of the method/field mappings. Untill loom reorders the mappings back to having `intermediary` as the source:
https://github.com/FabricMC/fabric-loom/blob/b37c4d3474fccd30f69beb25a20cc84da94f0574/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java#L122-L125

The imporant detail seems to be that `MappingSourceNsSwitch` has true passed for `dropMissingNewSrcName`, meaning that becuase we dont have an intermediary name for the none obf classes, their named mappings (along with params/comments/whatever) are dropped, giving us just:

```
intermediary				named
net/minecraft/class_123			net/minecraft/ExampleObfClass
```

When we actaully want to go back to:
```
intermediary				named
net/minecraft/class_123			net/minecraft/ExampleObfClass
					net/minecraft/NoneObfClass
```

I hate to say it, but unless im missing something this PR can just be a single line change from `true` -> `false` (ingoring test changes). Ill create a seperate PR so we can compare.

sylv256:
Unnecessary after deobfuscation, plus this is jank.

PR #1331, opened 2025-07-10
Add ability to add and remove annotations via mappings

Earthcomputer:
Adds the ability for mappings to add and remove annotations to vanilla code, by adding an `extras/annotations.json` file.


src/main/java/net/fabricmc/loom/configuration/providers/mappings/extras/annotations/AnnotationsData.java
```diff
@@ -0,0 +1,87 @@
...
+		JsonObject json = GSON.fromJson(reader, JsonObject.class);
+
+		if (!json.has("version")) {
+			throw new JsonSyntaxException("Missing annotations version");
+		}
+
+		int version = json.getAsJsonPrimitive("version").getAsInt();
+
+		if (version != 1) {
+			throw new JsonSyntaxException("Invalid annotations version " + version);
```
modmuss50: Maybe suggest to update loom as clueless users may see this in the future.

src/test/groovy/net/fabricmc/loom/test/unit/processor/AnnotationsApplyTest.groovy
```diff
@@ -0,0 +1,283 @@
...
+		def classReader1 = new ClassReader(getClassBytes(ExampleClass1))
+		classReader1.accept(classVisitor1, ClassReader.SKIP_CODE)
+
+		def text1 = textifyImportantPartsOfClass(annotatedNode1)
+		def field1Text = textify(annotatedNode1.fields.find { it.name == "field1" })
+		def field2Text = textify(annotatedNode1.fields.find { it.name == "field2" })
+		def method1Text = textify(annotatedNode1.methods.find { it.name == "method1" })
+		def method2Text = textify(annotatedNode1.methods.find { it.name == "method2" })
+
+		//noinspection GrDeprecatedAPIUsage
```
modmuss50, line 63: Out of interest whats deprecated here? Ideally new code shouldnt use deprecated API?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/extras/annotations/GenericAnnotationData.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AnnotationNode;
+import org.objectweb.asm.tree.TypeAnnotationNode;
+
+public record GenericAnnotationData(
+		Set<String> annotationsToRemove,
```
modmuss50, line 39: Thought/question/nit pick: These end up a little verbose in the json as `annotations_to_remove` it doesnt matter but wonder if it would be better as `to_remove`? You kinda know they are related to annotations. Again im looking for fault here.

modmuss50 (approved):
Looks really good, just had a quick look over. As I said on discord I hope to get an exp/1.12 branch going once im back from my holiday.

The json serialisation was quite a bit more involved that I had in mind when I suggested JSON, I still think its easier than a tottaly custom parser though.

I dont think there is any concern with using this from Yarn's build script (it depends on Loom as a lib), I dont think you touched anything directly related to Gradle.

The unit tests look really good, you could maybe add a integration test to test it end to end but its likely quite a lot of effort for little gain. Maybe once we have a yarn build with this and unpick V3 I can handle both at once.

This has been something we have been talking about for many years 👍

src/main/java/net/fabricmc/loom/configuration/providers/mappings/extras/annotations/GenericAnnotationData.java
cputnam-a11y: potentially even just `remove`?

src/test/groovy/net/fabricmc/loom/test/unit/processor/AnnotationsApplyTest.groovy
Earthcomputer: The test class, that I'm using to test the removal of deprecated

modmuss50: Ah sneaky 😂

src/main/java/net/fabricmc/loom/configuration/providers/mappings/extras/annotations/GenericAnnotationData.java
Earthcomputer: On the topic of making the format less verbose, I suppose it would make more sense to nest field and method data inside class data

PR #1332, opened 2025-07-10
Disable jar verification by default

modmuss50:
I dont have a good solution for old server jars that arent on Mojang's launcher meta.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
@@ -122,7 +122,7 @@ public abstract class MinecraftProvider {
 	}
 
 	private void verifyJars() throws IOException, SignatureVerificationFailure {
-		if (GradleUtils.getBooleanProperty(getProject(), Constants.Properties.DISABLE_MINECRAFT_VERIFICATION)) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftProvider.java
```diff
@@ -122,7 +122,7 @@ protected void initFiles() {
 	}
 
 	private void verifyJars() throws IOException, SignatureVerificationFailure {
-		if (GradleUtils.getBooleanProperty(getProject(), Constants.Properties.DISABLE_MINECRAFT_VERIFICATION)) {
+		if (GradleUtils.getBooleanProperty(getProject(), Constants.Properties.DISABLE_MINECRAFT_VERIFICATION, true)) {
 			LOGGER.info("Skipping Minecraft jar verification!");
```
LostLuma, line 126: This is not fully related to the PR, but this block also needs to exit the method otherwise setting the property does nothing.

modmuss50: Opps, yeah thats not good. Ill fix that in another PR. Im still unsure if I want to disable this by default.

modmuss50: Fixed directly in 1.11.

modmuss50:
Going to make a new PR for this.

PR #1337, opened 2025-07-26
Misc peformance improvements

modmuss50:


src/main/java/net/fabricmc/loom/util/AsyncCache.java
```diff
@@ -0,0 +1,80 @@
...
+		try {
+			return future.join();
+		} catch (CompletionException e) {
+			sneakyThrow(e.getCause() != null ? e.getCause() : e);
+			throw new IllegalStateException();
+		}
+	}
+
+	private static <E extends Throwable> void sneakyThrow(Throwable e) throws E {
+		//noinspection unchecked
```
Juuxel: Should probably be a `@SuppressWarnings` annotation instead

Issue #1284, opened 2025-04-07
Allow definition of Interface Injection in Gradle DSL


isXander:
Currently you can only specify interface injection by putting it in `fabric.mod.json` (FMJ). Loom reads from the FMJ at sync time.

I propose extending the Loom Gradle extension to allow you define this in the buildscript, much like access wideners are already.

*examples in Kotlin DSL*
```kt
// A crude but easier way of implementing this
loom.interfaceInjection = mapOf(
    "net/minecraft/TargetClass" to listOf("my/mod/interface")
)

// A more complete DSL
loom.interfaceInjection {
    target("net/minecraft/TargetClass") {
        inject("my/mod/interface")
    }
}
```

## Why?

My specific use case for this is to aid in creating my own gradle plugin that wraps both Fabric Loom and NeoForge into a single API.

This way, a single declaration (a bit like shown above), can configure both Loom and NeoForge, as well as transforming their respective mod manifests to include in prod.

PR #1345, opened 2025-08-06
fabric.mod.json generation task

### LINKED ISSUE #1344: fabric.mod.json generatation
The idea is to add a generic task for generating a fabric.mod.json file from within Gradle, something along the lines of:

```gradle
tasks.register("generateFmj", GenerateFabricModJson) {
    output = file("src/main/resources/fabric.mod.json")

    modid = "example"
    name = "Example Mod"
    dependencies {
        depends("fabric-api-base", ">=0.0.1")
        breaks("example", "1.0.0")
    }
    // And everything else FMJs support
}
```
The task should not dictate where and how the generated file is used, this is to allow developers to have full control over how their build is configured. We must ensure that the build still supports other IDEs and Intelij that dont build with Gradle.

I dont expect this would be used in the example mod either. If you have any thoughts/ideas on this please let them be known.

modmuss50:
Minimal example:

```groovy
tasks.register("generateModJson", net.fabricmc.loom.task.FabricModJsonV1Task) {
	outputFile = file("fabric.mod.json")

	json {
		modId = "examplemod"
		version = "1.0.0"
	}
}
```

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/fmj/gen/FabricModJsonV1Generator.java
+++ b/src/main/java/net/fabricmc/loom/util/fmj/gen/FabricModJsonV1Generator.java
@@ -29,8 +29,12 @@ import static net.fabricmc.loom.util.fmj.gen.GeneratorUtils.addArray;
 import static net.fabricmc.loom.util.fmj.gen.GeneratorUtils.addRequired;
 import static net.fabricmc.loom.util.fmj.gen.GeneratorUtils.addStringOrArray;
 
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
+import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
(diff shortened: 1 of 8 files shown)


maityyy:
Closes #1284 I guess

src/main/java/net/fabricmc/loom/util/fmj/gen/FabricModJsonV1Generator.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+
+import net.fabricmc.loom.LoomGradlePlugin;
+import net.fabricmc.loom.api.fmj.FabricModJsonV1Spec;
+
+// Opposite of https://github.com/FabricMC/fabric-loader/blob/master/src/main/java/net/fabricmc/loader/impl/metadata/V1ModMetadataParser.java
+public final class FabricModJsonV1Generator implements FabricModJsonGenerator<FabricModJsonV1Spec> {
```
maityyy, line 46: I just thought it was a little strange that there isn't some kind of common API for serializing FMJ

modmuss50: Could possibly do this with that in mind, I worry it would be overcomplex to provide a none Gradle based API for it, and then build the Gradle DSL ontop of that.

Gamebuster19901: I feel like the actual generation should be `api` in loader. But iirc loader intentionally removed gson, and just copied classes it needed from it into its codebase.

So that might not actually be feasible. This implementation looks okay so far though. I’ll try testing it out when I get off work today.

modmuss50: We can maybe look at an API in loader later on, im not too sure how useful it would be though. There is still a lot I need to do with this PR, its quite tedious/boring work so im going to do it bit by bit.

maityyy: Btw, JDK are discussing [an official API for JSON](https://github.com/openjdk/jdk-sandbox/tree/json/src/java.base/share/classes/java/util/json)

modmuss50:
> Closes #1284 I guess

No, thats tottaly unrelated.

maityyy:
> > Closes #1284 I guess
> 
> No, thats tottaly unrelated.

so no interface injection support? for now?

modmuss50:
> > > Closes #1284 I guess
> > 
> > 
> > No, thats tottaly unrelated.
> 
> so no interface injection support? for now?

This PR is purely for generating a fmj in a task. You can use it to generate an FMJ with an injected interface. That issue is asking to define injected interfaces without using an fmj.

maityyy:
yeah I understand, that really makes sense

PR #1347, opened 2025-08-10
Unpick v3 fixes

Earthcomputer:
- Update unpick
- Pass slashy class names into tiny remapper (as it expects them) and convert the return value back to dotty
- Use reflection as a fallback to get the field type of a field in a JDK class
- Throw early when unable to find the descriptor of a field (returning null will create NPEs down the road anyway)
- Fixed remapper test to properly mock TR to handle slashy class names. Also added missing mock of relevant `trEnvironment` methods used for getting the descriptor of a field.


modmuss50 (approved):
Looks great, thanks. After yarn has been published with V3 an intergration test should be added to make sure I dont break this going forward. 😄

PR #1351, opened 2025-08-12
Add a new Property based RunConfiguration API.

modmuss50:
This should be backwards compatible (mostly), but has required new names for the properties to allow backwards comapt.

TODO:
Fix eclipse and idea launch not using correct relative path. E.g move `${workspace_loc:%ECLIPSE_PROJECT%}` from the xml to code.


src/main/java/net/fabricmc/loom/api/RunConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+	 *
+	 * <p>Note: unless the project is the root project (or {@link #getAppendProjectPathToConfigName()} is disabled),
+	 * the project path will be appended automatically, e.g. 'Minecraft Client (:some:project)'.
+	 */
+	Property<String> getConfigurationName();
+
+	/**
+	 * Arguments for the JVM, such as system properties.
+	 */
+	ListProperty<String> getJVMArguments();
```
modmuss50: Rename to `getJvmArguments`

src/main/java/net/fabricmc/loom/api/RunConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+		getJVMArguments().add("-D%s=%s".formatted(name, value));
+	}
+
+	default void property(String name) {
+		getJVMArguments().add("-D%s".formatted(name));
+	}
+
+	default void properties(Map<String, String> props) {
+		props.forEach(this::property);
+	}
```
modmuss50: A map for the properties is likely better, then they can be passed via the DLI config.

src/main/java/net/fabricmc/loom/api/RunConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+public interface RunConfiguration extends Named {
+	/**
+	 * The full name of the run configuration, i.e. 'Minecraft Client'.
+	 *
+	 * <p>By default this is determined from the base name.
+	 *
+	 * <p>Note: unless the project is the root project (or {@link #getAppendProjectPathToConfigName()} is disabled),
+	 * the project path will be appended automatically, e.g. 'Minecraft Client (:some:project)'.
+	 */
+	Property<String> getConfigurationName();
```
Juuxel: `getDisplayName` might be clearer (`RunConfiguration.getName()` from `Named` is already a "configuration name")

Technici4n: Or getIdeName?

src/main/java/net/fabricmc/loom/api/RunConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Arguments for the program, usually Minecraft specific arguments.
+	 */
+	ListProperty<String> getProgramArguments();
+
+	/**
+	 * Environment variables to set when running the configuration.
+	 */
+	MapProperty<String, Object> getEnvironmentVars();
```
Technici4n, line 65: Why vars instead of variables?

modmuss50: The old deprecated method is `getEnvironmentVariables` so a diffrent name is needed. Maybe just `getEnvironment`?

Technici4n: getEnvVariables? Still not great

ljfa-ag:
> TODO:
Fix eclipse and idea launch not using correct relative path. E.g move ${workspace_loc:%ECLIPSE_PROJECT%} from the xml to code.

Any updates on resolving this? Currently, when used in a Gradle subproject, the relative path of the subproject is added twice to the Eclipse working directory, for example "rootProject/fabric/fabric/run" instead of "rootProject/fabric/run". The "${workspace_loc:...}" value in the run config apparently already points to the subproject location.

PR #1357, opened 2025-08-26
remove use of Task.getProject at exec time in ValidateAccessWidenerTask

derklaro:
Seems like this was missed - this is unsupported with configuration cache enabled.
Not sure if this should target `dev/1.11` or `exp/1.12` - issue exists in both branches.


modmuss50 (approved):

modmuss50:
Good spot, likely missed as its an error case that has no test coverage.

PR #1361, opened 2025-09-02
Fix split official namespaces existing on versions where they don't make sense

### LINKED ISSUE #1360: Mappings on single-jar legacy versions broken due to force-enabling split official namespaces
(Loom 1.11.7 on Gradle 9.0.0)

When targeting MC 0.30, Loom incorrectly uses the `clientOfficial` and `serverOfficial` namespaces of legacy merged versions despite 0.30 having no server jar. This means that mappings that use the correct `official` namespace no longer work, despite working in previous 1.x versions such as 1.5.

The `clientOfficial` and `serverOfficial` namespaces should not exist when there's only one environment available to begin with.

The specific issue that caused mappings to stop working was not adding those namespaces since namespace completion filled in the blanks. The bug was "fully" added in #1171 which stopped unwanted ns completion on split official ns versions to prevent mapping conflicts.

Reproducer (see the two branches on that repo): https://github.com/Juuxel/classic-mappings-reproducer

I can confirm from local testing that applying [this patch](https://gist.github.com/Juuxel/9c5f90dd667876b63676a68fc093d735) to completely disable the merged namespace system single-jar envs with no intermediate mappings works.

Juuxel:
Fixes #1360.

Renames an experimental API in `IntermediateMappingsProvider`: `getIsLegacyMinecraft` -> `getUseSplitOfficialNamespaces` to make it more clear what the method does since it doesn't apply to all legacy versions anymore.

TODO:
- [x] Should the `isLegacyClientAndServer` check do a specific version range instead to account for individual versions in the beta/pre-1.3 version range that don't have a client/server even if they have the other component? See https://github.com/FabricMC/fabric-loom/issues/1360#issuecomment-3245898067

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
+++ b/src/main/java/net/fabricmc/loom/api/mappings/intermediate/IntermediateMappingsProvider.java
@@ -1,7 +1,7 @@
 /*
  * This file is part of fabric-loom, licensed under the MIT License (MIT).
  *
- * Copyright (c) 2022 FabricMC
+ * Copyright (c) 2022-2025 FabricMC
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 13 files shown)


modmuss50 (approved):
Im happy with this for now, we can look at the mappings stuff later (I expect its harder than it sounds).

@SpaceWalkerRS please could you also take a look at this PR.

SpaceWalkerRS (approved):

PR #1364, opened 2025-09-05
Make `fabric.mod.json` path configurable at configuration stage.

kikugie:
## Motivation
Multi-version mods often have a shared source directory, which is then processed to generated sources for each subproject. Currently Loom expects `fabric.mod.json` to be in `src/main/resources` *(or client)*` of the project's folder.
In multi-version projects, this file is located in the shared source, and due to the hardcoded path, Loom doesn't recognize it, leading to issues with using injected interfaces.

## Implementation
This PR adds `fabricModJsonPath` property to the Loom extension. If specified that file is used, otherwise falling back to the existing behaviour.


src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonFactory.java
```diff
@@ -109,7 +111,10 @@ public static Optional<FabricModJson> createFromZipOptional(Path zipPath) {
 
 	@Nullable
 	public static FabricModJson createFromSourceSetsNullable(Project project, SourceSet... sourceSets) throws IOException {
-		final File file = SourceSetHelper.findFirstFileInResource(FABRIC_MOD_JSON, project, sourceSets);
+		final RegularFileProperty fmjPath = project.getExtensions().getByType(LoomGradleExtensionAPI.class).getFabricModJsonPath();
```
modmuss50: I dont think this right place to do this, `FabricModJsonHelpers.getModsInProject` is the place for it.

modmuss50 (commented):
If you can an intergration test to cover this would be very useful. Let me know if you need help with it.

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonFactory.java
kikugie: What exactly? My implementation was based on the fact that the `createFromSourceSetsNullable` method is called from the Loom and Fabric API extensions (`LoomGradleExtensionApiImpl.getModVersion` and `FabricApiAbstractSourceSet.configureSourceSet`), so using the property here would provide the desired file to those too.

kikugie:
A thing to consider in a next commit is the way to deal with fmj used in the datagen source set. Currently it still uses the old method, but it would also be helpful to be able to specify its path.

src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiAbstractSourceSet.java
```diff
@@ -46,6 +47,8 @@ abstract class FabricApiAbstractSourceSet {
 
 	protected abstract String getSourceSetName();
 
+	protected abstract RegularFileProperty getFabricModJsonPath();
```
kikugie: I still need to test if Gradle adds a backing property for this, because sometimes it just doesn't.

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonHelpers.java
```diff
@@ -24,39 +24,54 @@
...
-		try {
-			final FabricModJson fabricModJson = FabricModJsonFactory.createFromSourceSetsNullable(project, sourceSets.toArray(SourceSet[]::new));
+		return getModsInProject(project, overrideFile, sourceSets.toArray(SourceSet[]::new));
+	}
 
+	/**
+	 * Returns the list of mods provided by either {@code overrideFile} property
+	 * or {@code fabric.mod.json} in the {@code sourceSets} array.
+	 */
+	public static List<FabricModJson> getModsInProject(Project project, Provider<File> overrideFile, SourceSet... sourceSets) {
```
modmuss50: Why is this needed? Can you just not always get it from the extension? I dont get why this is needed in the fabricapi extension?

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonFactory.java
```diff
@@ -107,10 +107,29 @@ public static Optional<FabricModJson> createFromZipOptional(Path zipPath) {
 		return Optional.ofNullable(createFromZipNullable(zipPath));
 	}
 
+	@Nullable
+	public static FabricModJson createFromOverrideNullable(@Nullable File file) throws IOException {
```
modmuss50: createFromFileNullable

src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiAbstractSourceSet.java
```diff
@@ -64,17 +64,14 @@ protected SourceSet configureSourceSet(Property<String> modId, boolean isClient)
 		});
 
 		modId.convention(getProject().provider(() -> {
-			try {
-				final FabricModJson fabricModJson = FabricModJsonFactory.createFromSourceSetsNullable(getProject(), sourceSet);
+			List<FabricModJson> fabricModJsons = FabricModJsonHelpers
+				.getModsInProject(getProject(), null, sourceSet);
 
-				if (fabricModJson == null) {
-					throw new RuntimeException("Could not find a fabric.mod.json file in the data source set or a value for DataGenerationSettings.getModId()");
```
modmuss50, line 71: Add this exception back.

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonFactory.java
```diff
@@ -107,10 +107,29 @@ public static Optional<FabricModJson> createFromZipOptional(Path zipPath) {
...
+		var modJson = readFmjJsonObject(file);
+		if (modJson == null) {
```
modmuss50: Please make sure to run checkstyle, this wont pass.

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonFactory.java
modmuss50: Why would the passed in file ever be null?

src/main/java/net/fabricmc/loom/api/fabricapi/DataGenerationSettings.java
```diff
@@ -67,4 +72,15 @@ public interface DataGenerationSettings {
 	 * Contains a boolean property indicating whether data generation will be compiled and ran with the client.
 	 */
 	Property<Boolean> getClient();
+
+	/**
+	 * Sets {@link #getModId()} property based on the {@code id} field defined in the provided file.
+	 */
+	default void modId(File fabricModJsonFile) {
+        try {
+            getModId().set(Objects.requireNonNull(FabricModJsonFactory.createFromOverrideNullable(fabricModJsonFile)).getId());
```
modmuss50: This should fail if its an invalid FMJ.

src/main/java/net/fabricmc/loom/api/fabricapi/DataGenerationSettings.java
```diff
@@ -67,4 +72,15 @@ public interface DataGenerationSettings {
...
+        } catch (IOException | NullPointerException e) {
```
modmuss50: Caching an NullPointerException is an anti pattern imo this should never happen, see the other comments.

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonHelpers.java
```diff
@@ -24,39 +24,54 @@
...
-		try {
-			final FabricModJson fabricModJson = FabricModJsonFactory.createFromSourceSetsNullable(project, sourceSets.toArray(SourceSet[]::new));
+		return getModsInProject(project, overrideFile.getOrNull(), sourceSets.toArray(SourceSet[]::new));
+	}
 
+	/**
+	 * Returns the list of mods provided by either {@code overrideFile} property
+	 * or {@code fabric.mod.json} in the {@code sourceSets} array.
+	 */
+	public static List<FabricModJson> getModsInProject(Project project, @Nullable File overrideFile, SourceSet... sourceSets) {
```
modmuss50: Is this method still needed?

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonHelpers.java
```diff
@@ -24,39 +24,54 @@
...
+	 */
 	public static List<FabricModJson> getModsInProject(Project project) {
-		final LoomGradleExtension extension = LoomGradleExtension.get(project);
+		LoomGradleExtension extension = LoomGradleExtension.get(project);
+		var overrideFile = extension.getFabricModJsonPath().getAsFile();
 		var sourceSets = new ArrayList<SourceSet>();
-		sourceSets.add(SourceSetHelper.getMainSourceSet(project));
 
+		sourceSets.add(SourceSetHelper.getMainSourceSet(project));
 		if (extension.areEnvironmentSourceSetsSplit()) {
```
modmuss50, line 56: Again this wont pass checkstyle, try not change existing code.

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonHelpers.java
```diff
@@ -24,39 +24,54 @@
...
 public class FabricModJsonHelpers {
-	// Returns a list of Mods found in the provided project's main or client sourcesets
+	/**
+	 * Returns the list of mods provided by either {@link LoomGradleExtensionAPI#getFabricModJsonPath()}
+	 * or {@code fabric.mod.json} in main or client resources.
+	 */
 	public static List<FabricModJson> getModsInProject(Project project) {
-		final LoomGradleExtension extension = LoomGradleExtension.get(project);
+		LoomGradleExtension extension = LoomGradleExtension.get(project);
+		var overrideFile = extension.getFabricModJsonPath().getAsFile();
```
modmuss50: I think this can just return early if the override file is present. If it is it should try to read the FMJ and throw if it fails to read the FMJ. We shouldnt sliently hide errors from the user like this.

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -66,6 +66,8 @@ public interface LoomGradleExtensionAPI {
 
 	RegularFileProperty getAccessWidenerPath();
 
+	RegularFileProperty getFabricModJsonPath();
```
modmuss50, line 83: Some basic javadoc wouldn't harm. I know the existing functions dont have much but lets try to set a good example going forward.

modmuss50 (commented):
Im happy with the API design and rough implementation. However it really needs some proper thought about what is a fatal error and what isnt. The null handling is a bit of a mess here, its not clear what can be null and what isnt.

I think what we really want is that if a user specifies a custom FMJ path, it should be a non null file that exists, and is a valid FMJ. Aything other than that should cause a hard crash.

Also please make sure to run checkstyle / spotless (there is an intelij plugin that makes it easy to see what is wrong). If you run `gradlew build -x test` locally you can verify that it will build on CI. (Dont try to run all the tests locally otherwise you will be waiting all day)

kikugie:
image
Not my issue :shrug:

modmuss50:
> Not my issue 🤷

It was, I fixed it by running `spotlessApply`

src/main/java/net/fabricmc/loom/util/fmj/FabricModJsonHelpers.java
```diff
@@ -24,37 +24,41 @@
...
 
 public class FabricModJsonHelpers {
-	// Returns a list of Mods found in the provided project's main or client sourcesets
+	/**
+	 * Returns the list of mods provided by either {@link LoomGradleExtensionAPI#getFabricModJsonPath()}
+	 * or {@code fabric.mod.json} in main or client resources.
+	 */
 	public static List<FabricModJson> getModsInProject(Project project) {
 		final LoomGradleExtension extension = LoomGradleExtension.get(project);
+		var overrideFile = extension.getFabricModJsonPath().getAsFile();
```
modmuss50: FYI: I see you like to use `var` wherever possible, this isnt something we tend to do. In other fabric projects we disallow var other than for new instance creataion. e.g `var a = new Object()` is fine as you can clearly see the type. Loom doesnt enforce this as there are some places where gradle forces you to use an obtuse type, but for simple types such a `File` in this case I wouldnt choose to use var as it makes the code harder to read.

Dont worry about it for this PR its just a minor nit pick.

kikugie: Thanks for pointing out. I used var for consistency with the other cases I've seen in the area, but I can replace it if needed.

modmuss50: Ive done it :) I also moved the tests to one of the existing files, this is becuase each test file gets a new github actions job, there was nothing wrong with having its own file but it will be a tiny bit quicker. Im happy with this PR now :) Thanks.

modmuss50 (approved):

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionApiImpl.java
```diff
@@ -74,6 +72,7 @@
 import net.fabricmc.loom.task.GenerateSourcesTask;
 import net.fabricmc.loom.util.DeprecationHelper;
 import net.fabricmc.loom.util.MirrorUtil;
+import net.fabricmc.loom.util.fmj.FabricModJson;
```
kikugie, line 75: Linters - my greatest nemesis

Mat0u5:
You two are the best <3

PR #1368, opened 2025-09-18
Add source/classPaths to generated vscode launch.json

### LINKED ISSUE #1367: Generated Run Configurations for VS Code Include `modCompileOnly` dependencies
When running a project that uses `modCompileOnly` dependencies, `./gradlew runClient` correctly does not include those dependencies at runtime. Using the run configurations generated by `./gradlew vscode` however will include those dependencies when running.

For an example, my mod [Custom Fog](https://github.com/Setadokalo/custom-fog) has a compile-only dependency on Sodium, which shows up at runtime only when using the VSCode run profile.

Setadokalo:
The run configurations produced by fabric-loom in the current version are mostly functional, but they include mods that are `modCompileOnly` dependencies (and presumable `modCompileOnlyApi`, though I've never used that myself). This PR fixes that by explicitly providing the `sourcePaths` and `classPaths` to the vscode java debugger, instead of allowing it to automatically detect them.

Fixes #1367.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/GenVsCodeProjectTask.java
+++ b/src/main/java/net/fabricmc/loom/task/GenVsCodeProjectTask.java
@@ -24,6 +24,7 @@
 
 package net.fabricmc.loom.task;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


Setadokalo:
A problem with this method is that it means the `vscode` setup command has to be rerun when the build configuration changes. I'll try to find a better way to keep the classpath up to date.

modmuss50:
As you said im not too keen on this as it requires keeping the json uptodate, did you manage to find a better way? I think the default behaviour is acceptable for most people, maybe this could be behind an opt-in option?

Setadokalo:
I've been looking through the code for the java debug plugin for vscode trying to figure out how exactly it determines the class path, if I'm understanding correctly it uses code from eclipse's buildship for that, and it looks like there's open issues about [exactly this problem on the eclipse side](https://github.com/eclipse-buildship/buildship/issues/1313). So... I think solving this properly is actually something to handle on the buildship side, but I think having an opt-in flag to generate the classPath in the run config is a good stopgap.

PR #1387, opened 2025-10-05
Only warn when attempting to configure the disabled mixin AP

modmuss50:


Juuxel (approved):
This could probably use the problem reporting API too btw, but might be too susceptible to potential Gradle breakages since it's still incubating and this is an important code path

modmuss50:
> This could probably use the problem reporting API too btw, but might be too susceptible to potential Gradle breakages since it's still incubating and this is an important code path

Yes, I did think about using that, and chose not to for that reason. When its stable we can move over to that and centralise all of the deprecations somewhere so its easy to remove them when we finally get around to a 2.0

Juuxel:
Yeah, I think expanding the LoomProblems class might be a good starting point for it. (There can be a problem group for just deprecations for example)

PR #1392, opened 2025-10-08
Support classpath groups when using configure on demand.

modmuss50:
TODO:

- [ ] Actually test it
- [ ] Maybe choose a better name for the new plugin
- [ ] Check versioning of new plugin

### DIFF
--- a/src/main/java/net/fabricmc/loom/api/ModSettings.java
+++ b/src/main/java/net/fabricmc/loom/api/ModSettings.java
@@ -26,6 +26,7 @@ package net.fabricmc.loom.api;
 
 import java.io.File;
 import java.util.List;
+import java.util.Map;
 
 import javax.inject.Inject;
 
@@ -38,6 +39,7 @@ import org.gradle.api.provider.ListProperty;
 import org.gradle.api.tasks.SourceSet;
 import org.jetbrains.annotations.ApiStatus;
 
+import net.fabricmc.loom.LoomCompanionGradlePlugin;
 import net.fabricmc.loom.configuration.classpathgroups.ExternalClasspathGroup;
 import net.fabricmc.loom.util.gradle.SourceSetHelper;
 import net.fabricmc.loom.util.gradle.SourceSetReference;
@@ -90,6 +92,8 @@ public abstract class ModSettings implements Named {
 	 */
 	@Deprecated
 	public void sourceSet(SourceSet sourceSet, Project project) {
+		ensureCompanion(project);
+
 		sourceSet(sourceSet.getName(), project.getPath());
 	}
 
@@ -101,6 +105,8 @@ public abstract class ModSettings implements Named {
 	 */
 	@Deprecated
 	public void sourceSet(String name, Project project) {
+		ensureCompanion(project);
+
 		sourceSet(name, project.getPath());
 	}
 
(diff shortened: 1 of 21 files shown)


src/main/java/net/fabricmc/loom/task/launch/ExportClassPathTask.java
```diff
@@ -0,0 +1,93 @@
...
+import org.gradle.workers.WorkAction;
+import org.gradle.workers.WorkParameters;
+import org.gradle.workers.WorkQueue;
+import org.gradle.workers.WorkerExecutor;
+
+import net.fabricmc.loom.LoomGradlePlugin;
+import net.fabricmc.loom.configuration.classpathgroups.ExternalClasspathGroupDTO;
+import net.fabricmc.loom.task.AbstractLoomTask;
+
+public abstract class ExportClassPathTask extends AbstractLoomTask {
```
Juuxel: Nit: probably should be `ExportClasspathTask` for consistency

src/main/java/net/fabricmc/loom/configuration/classpathgroups/plugin/LoomClasspathGroupPlugin.java
```diff
@@ -0,0 +1,53 @@
...
+	private static final String NAME = "net.fabricmc.fabric-loom-classpath-group";
+
+	@Override
+	public void apply(@NotNull Project target) {
+		applyToProject(target);
+	}
+
+	public static void applyToProject(Project project) {
+		if (project.getPlugins().hasPlugin(LoomGradlePlugin.NAME) && project.getPlugins().hasPlugin(NAME)) {
+			throw new IllegalStateException("fabric-loom-classpath-group must not be applied alongside fabric-loom");
```
Juuxel: Why not just have Loom apply this plugin instead of erroring?

modmuss50: Good idea 👍

src/main/java/net/fabricmc/loom/configuration/classpathgroups/ClasspathGroup.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.File;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+import net.fabricmc.loom.api.ModSettings;
+
+public record ClasspathGroup(List<String> paths, List<ExternalClasspathGroup> externalGroups) implements Serializable {
```
modmuss50, line 35: This class fails to serialise when being ran with Gradle 8.14 for some reason that I cannot figure out. Same happens when its not a record.

src/main/java/net/fabricmc/loom/api/ModSettings.java
```diff
@@ -82,18 +86,42 @@ public void sourceSet(String name) {
 
 	/**
 	 * Add {@link SourceSet}'s output directories from the supplied project to be grouped with the named mod.
+	 * @deprecated Replaced with {@link #sourceSet(String, String)} to avoid passing a project reference.
 	 */
+	@Deprecated
 	public void sourceSet(SourceSet sourceSet, Project project) {
-		getModSourceSets().add(new SourceSetReference(sourceSet, project));
```
modmuss50, line 87: Just realised this wont be backwards compatible, as it assumes the companion plugin was added the other project. Unlesss we do it for them 🤔

modmuss50:
Im going to merge this, I expect something will break, do let me know :D

PR #1394, opened 2025-10-15
Apply record component annotations to getter methods and canonical constructor parameters

Earthcomputer:
I also moved the anonymous classes in the apply visitor to named inner classes to reduce nesting.

When the class is a record, it now reads all methods into `MethodNode`s and applies them inside `visitEnd()`, because record components as per the ASM docs are not guaranteed to be visited before methods (even though in practice they are by `ClassReader.accept` and `ClassNode.accept`).


modmuss50 (approved):

modmuss50:
Tests might be nice, but ill leave that up to you.

PR #1396, opened 2025-10-16
Add annotations apply integration test

Earthcomputer:
Adds an integration test for applying annotations in mappings, based on the annotation that is published in yarn at the moment. We can add more test cases when more annotations are added in yarn.

### DIFF
--- a/src/test/resources/projects/annotationsApply/build.gradle
+++ b/src/test/resources/projects/annotationsApply/build.gradle
@@ -1,0 +1,9 @@
+plugins {
+	id 'fabric-loom'
+}
+
+dependencies {
+	minecraft "com.mojang:minecraft:25w42a"
+	mappings "net.fabricmc:yarn:25w42a+build.9:v2"
+	modImplementation "net.fabricmc:fabric-loader:0.17.3"
+}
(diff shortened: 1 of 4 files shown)


src/test/resources/projects/annotationsApply/build.gradle
```diff
@@ -0,0 +1,9 @@
+plugins {
+	id 'fabric-loom'
+}
+
+dependencies {
+	minecraft "com.mojang:minecraft:25w42a"
+	mappings "net.fabricmc:yarn:25w42a+build.9:v2"
+	modImplementation "net.fabricmc:fabric-loader:0.17.3"
+}
```
modmuss50, line 9: For a test as simple as this just use the `minimalBase` project, and have a look at how the other tests use it, you can just move this dependencies block into the groovy test code. It doesnt look like you have a reason to have actual mod code for this test.

modmuss50 (approved):

src/test/resources/projects/annotationsApply/build.gradle
Earthcomputer: Without mod code, `:compileJava` would be skipped, and I noticed it wasn't generating the Minecraft jar

modmuss50: Uhh, that shouldnt be the case (I wish it was) 🤔 Happy to go with this its not a big deal.

PR #1417, opened 2025-11-01
Create fabric-loom-no-remap plugin

modmuss50:


isXander:
Can't this just be a property in the loom extension?

modmuss50:
> Can't this just be a property in the loom extension?

No, as it must be set before evaluation to skip creating the configurations and tasks.

PR #1419, opened 2025-11-03
Update to Mapping-IO 0.8.0

NebelNidas:


src/main/java/net/fabricmc/loom/configuration/providers/mappings/tiny/MappingsMerger.java
```diff
@@ -110,32 +109,11 @@ public static void legacyMergedMergeAndSaveMappings(Path from, Path out, Interme
...
-					if (!namedParentClass.equals(currentPath)) {
-						classEntry.setDstName(namedParentClass
-										+ "$" + String.join("$", Arrays.copyOfRange(path, i + 1, path.length)),
-								namedIdx);
-						break;
-					}
-				}
-			}
-		}
+		tree.propagateOuterClassNames("intermediary", List.of("named"), false);
```
NebelNidas, line 117: Passing `false` here for `processRemappedDstNames` matches the old code's behavior, but `true` would probably be preferable. Not sure if that actually changes anything for Yarn though.

modmuss50: Lets keep it as-is, unless there is a good reason to change it.

modmuss50 (approved):

PR #1420, opened 2025-11-03
Enable the usage of JSR annotations

opekope2:
Add option to loom to use JSR annotations. This helps IDEA with Kotlin nullability analysis ([if configured properly](https://kotlinlang.org/docs/java-interop.html#compiler-configuration)). Sparked by [this Discord conversation](https://discord.com/channels/507304429255393322/592749519171616796/1434664001995931819).

```kotlin
// build.gradle.kts
loom {
  remapJsrAnnotationsToJetBrains = true
}
```

TODO:

* [x] Re-generate Minecraft JAR after changing this option
* [x] Remap JSR annotations using a JAR processor
* [ ] Add JSR to classpath when remapping is disabled (instead of/in addition to JetBrains)


zOnlyKroks:
> Add option to loom to use JSR annotations. This helps IDEA with Kotlin nullability analysis. Sparked by [this Discord conversation](https://discord.com/channels/507304429255393322/592749519171616796/1434664001995931819).
> 
> ```kotlin
> // build.gradle.kts
> loom {
>   useJsrAnnotations = true
> }
> ```
> 
> TODO:
> 
>     * [x]  Re-generate Minecraft JAR after changing this option
> 
>     * [x]  Remap JSR annotations using a JAR processor
> 
>     * [ ]  Add JSR to classpath when remapping is disabled (instead of/in addition to JetBrains)

The processor wraps exceptions but doesn't provide much context:

> throw new RuntimeException("Failed to remap JSR annotations", e);

Suggestion: Add more diagnostic information (which JAR failed, which annotation caused issues, etc.)

Next Suggestion: Add integration tests -> Annotations are correctly remapped

opekope2:
The actual log message is

> throw new RuntimeException("Failed to remap annotations to JSR in " + jar, e);

where I append the JAR path to the end, just like `InterfaceInjectionProcessor` does.

Tests are a good idea, I'll write some later.

Some notes about this PR I mentioned on discord:

* Remapping JSR to JetBrains is the default behavior of Loom. When `remapJsrAnnotationsToJetBrains` is set to false, `JsrAnnotationRemapperProcessor` will map them back to JSR.
Not remapping JSR to JetBrains would cause breaking changes, so I kept this remapping
Minecraft 1.21.10 mostly uses JSR annotations, and sporadically JetBrains annotations, so these would get mapped to JSR if `remapJsrAnnotationsToJetBrains` is set to false
* Minecraft 1.21.11 uses JSpecify annotations, so the `remapJsrAnnotationsToJetBrains` will do nothing. This should be noted somewhere.

opekope2:
The target branch probably needs to be changed to 1.13

src/main/java/net/fabricmc/loom/api/LoomGradleExtensionAPI.java
```diff
@@ -257,6 +257,8 @@ default void splitMinecraftJar() {
 
 	boolean areEnvironmentSourceSetsSplit();
 
+	Property<Boolean> getRemapJsrAnnotationsToJetBrains();
```
modmuss50, line 280: A bit of documentation wouldnt hurt

modmuss50 (commented):
I think this is good, there is no point over enginnering something like this. Please could you retarget to `exp/1.14`?

modmuss50 (approved):

Issue #1422, opened 2025-11-04
Support none obfucasted versions


modmuss50:
This issue is basically just a TODO list for Loom 1.14, most of these tasks can be done in no particular order.

- [x] Add `net.fabricmc.fabric-loom-no-remap-experimental` plugin - https://github.com/FabricMC/fabric-loom/pull/1417
- [x] Support using the experimental debof mc jars - https://github.com/FabricMC/fabric-loom/pull/1423
- [x] Add a replacement for finding/using the `fabric-installer.json` file, it should come from the main compile classpath I think. https://github.com/FabricMC/fabric-loom/pull/1428
- [x] Create a none remapped impl of SpecContext (used to find TAW, injected interfaces from deps) - https://github.com/FabricMC/fabric-loom/pull/1424
- [x] Get the game running, dont pass/generate the remap classpath. https://github.com/FabricMC/fabric-loom/pull/1426
- [x] Support JIJ with the regular `jar` task. - https://github.com/FabricMC/fabric-loom/pull/1425
- [x] ~~Re-split split sources mods into two jars.~~
- [x] Add loom's manifest attributes to the regular `jar` task. https://github.com/FabricMC/fabric-loom/pull/1429
- [x] ~~Maybe? Strip nested JIJs from mod depedencies? What will happen if we dont?~~
- [x] Support XVFB on the regular run tasks?
- [x] Build a deobf version of Fabric API, maybe publish it somewhere for testing?
- [ ] Explore supporting dev only remapping for these versions.

PR #1425, opened 2025-11-06
[Non-Obf]: Support JIJ with the regular jar task

zOnlyKroks:
As per #1422

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
+++ b/src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
@@ -34,6 +34,7 @@ import org.gradle.api.Action;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.ConfigurationContainer;
 import org.gradle.api.artifacts.dsl.ArtifactHandler;
 import org.gradle.api.file.Directory;
 import org.gradle.api.plugins.BasePlugin;
@@ -48,12 +49,15 @@ import net.fabricmc.loom.LoomGradleExtension;
 import net.fabricmc.loom.build.nesting.JarNester;
 import net.fabricmc.loom.build.nesting.NestableJarGenerationTask;
 import net.fabricmc.loom.util.Constants;
+import net.fabricmc.loom.util.gradle.GradleUtils;
+import net.fabricmc.loom.util.gradle.SourceSetHelper;
 import net.fabricmc.loom.util.gradle.SyncTaskBuildService;
 
 import org.jetbrains.annotations.NotNull;
 
 public abstract class RemapTaskConfiguration implements Runnable {
 	public static final String REMAP_JAR_TASK_NAME = "remapJar";
+	public static final String REMAP_SOURCES_JAR_TASK_NAME = "remapSourcesJar";
 
 	@Inject
 	protected abstract Project getProject();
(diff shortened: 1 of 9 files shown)


src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -190,4 +206,48 @@ private void trySetupSourceRemapping() {
...
+		NestJarsAction(Provider<Directory> nestedJarsDir) {
+			this.nestedJarsDir = nestedJarsDir;
+		}
+
+		@Override
+		public void execute(@NotNull Task t) {
+			final Jar jarTask = (Jar) t;
+
+			if (!jarTask.getArchiveFile().isPresent()) {
+				jarTask.getLogger().warn("Jar task archive file not available, skipping JIJ");
```
Juuxel: I don't think this is reachable in practice as the task shouldn't be running without its output file configured.

zOnlyKroks: true, will remove that unnecessary check

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -67,63 +75,71 @@ public void run() {
...
+				task.dependsOn(processIncludeJarsTask);
+				// Use JarNester to properly add jars and update fabric.mod.json
+				task.doLast(new NestJarsAction(processIncludeJarsTask.flatMap(NestableJarGenerationTask::getOutputDirectory)));
+			});
 
-		// must not be lazy to ensure that the prepare tasks get setup for other projects to depend on.
-		// Being lazy also breaks maven publishing, see: https://github.com/FabricMC/fabric-loom/issues/1023
-		getTasks().create(REMAP_JAR_TASK_NAME, RemapJarTask.class, remapJarTaskAction);
+			// Add jar task to unmapped collection
+			extension.getUnmappedModCollection().from(getTasks().getByName(JavaPlugin.JAR_TASK_NAME));
```
modmuss50: Move into a new class.

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -67,63 +75,71 @@ public void run() {
...
+		} else {
```
modmuss50: Prefer early returns instead of else if possible.

src/main/java/net/fabricmc/loom/task/RemapTaskConfiguration.java
```diff
@@ -190,4 +206,42 @@ private void trySetupSourceRemapping() {
 			}
 		});
 	}
+
+	/**
+	 * Configuration-cache-compatible action for nesting jars.
+	 * Uses a provider to avoid capturing task references at configuration time.
+	 * Do NOT turn me into a record!
+	 */
+	private static class NestJarsAction implements Action<Task>, Serializable {
```
modmuss50: Make it a seperate class, classes are free!

modmuss50 (approved):
Changes look good, its not everything we need but a lot better than nothing, I'd be happy to merge this and cleanup later or wait for you to do it.

zOnlyKroks:
> Changes look good, its not everything we need but a lot better than nothing, I'd be happy to merge this and cleanup later or wait for you to do it.

will clean this up right now :)

modmuss50 (approved):

PR #1429, opened 2025-11-07
Add manifest attributes and jar filtering to non-remapped jar task

zOnlyKroks:
Adds proper manifest modification to the non-remapped jar task to bring it to feature parity with RemapJarTask.

  Changes:
  - Set `Fabric-Mapping-Namespace` to `official` for non-remapped jars
  - Add all Loom metadata attributes (versions, etc.) via JarManifestService
  - Support client-only entries manifest attributes when using split source sets
  - Filter only `.jar` files from nested jars directory
  - Add test coverage to verify manifest attributes are correctly written

  This ensures non-remapped jars have the same metadata as remapped jars, making them properly identifiable by fabric-loader and other tools.


src/main/java/net/fabricmc/loom/task/ManifestModificationAction.java
```diff
@@ -0,0 +1,103 @@
...
+import net.fabricmc.loom.task.service.JarManifestService;
+import net.fabricmc.loom.util.Check;
+import net.fabricmc.loom.util.Constants;
+import net.fabricmc.loom.util.ZipUtils;
+
+/**
+ * Action that modifies the manifest of a jar file to add Loom metadata.
+ * Configuration-cache-compatible implementation using providers.
+ */
+public class ManifestModificationAction implements Action<Task>, Serializable {
```
modmuss50, line 53: Can this be `Action<Jar>` to remove that cast?

src/main/java/net/fabricmc/loom/task/ManifestModificationAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	public void execute(@NotNull Task t) {
+		final Jar jarTask = (Jar) t;
+		final File jarFile = jarTask.getArchiveFile().get().getAsFile();
+
+		try {
+			modifyManifest(jarFile);
+		} catch (IOException e) {
+			throw new RuntimeException("Failed to modify jar manifest for " + jarFile.getName(), e);
```
modmuss50: UncheckedIOException

src/main/java/net/fabricmc/loom/task/ManifestModificationAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+	private void modifyManifest(File jarFile) throws IOException {
+		Map<String, String> manifestAttributes = new HashMap<>();
+
+		// Set the mapping namespace to "official" for non-remapped jars
+		manifestAttributes.put(Constants.Manifest.MAPPING_NAMESPACE, targetNamespace);
+
+		// Add client-only entries attributes if present
+		if (clientOnlyEntries != null && !clientOnlyEntries.isEmpty()) {
+			manifestAttributes.put(Constants.Manifest.SPLIT_ENV, "true");
```
modmuss50, line 90: This should still be true if and there are no client only entries `areEnvironmentSourceSetsSplit`. You can have a common only "split" jar.

src/main/java/net/fabricmc/loom/task/ManifestModificationAction.java
zOnlyKroks: The Task.doLast() method signature is:
 void doLast(Action<? super Task> action)

 This requires an Action<Task> or an action on a supertype of Task (like Action<Object>), not a subtype like Action<Jar>.

PR #1433, opened 2025-11-08
Optimize github actions for subsequently faster runs

zOnlyKroks:
Rework github actions, to use caching / other strategies for faster subsequent runs

### DIFF
--- a/.github/workflows/publish-exp.yml
+++ b/.github/workflows/publish-exp.yml
@@ -11,13 +11,14 @@ jobs:
   build:
     runs-on: ubuntu-24.04
     steps:
+      - uses: actions/checkout@v4
+      - uses: gradle/actions/wrapper-validation@v4
       - name: setup jdk
         uses: actions/setup-java@v4
         with:
           java-version: 21
           distribution: 'temurin'
-      - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v4
+      - uses: gradle/actions/setup-gradle@v4
 
       # Generate the build number based on tags to allow per branch build numbers, not something github provides by default.
       - name: Generate build number
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,13 +11,14 @@ jobs:
   build:
     runs-on: ubuntu-24.04
     steps:
+      - uses: actions/checkout@v4
+      - uses: gradle/actions/wrapper-validation@v4
       - name: setup jdk
         uses: actions/setup-java@v4
         with:
           java-version: 21
           distribution: 'temurin'
-      - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v4
+      - uses: gradle/actions/setup-gradle@v4
 
       # Generate the build number based on tags to allow per branch build numbers, not something github provides by default.
       - name: Generate build number
--- a/.github/workflows/test-push.yml
+++ b/.github/workflows/test-push.yml
@@ -21,14 +21,26 @@ jobs:
   build:
     runs-on: ubuntu-24.04
     steps:
+      - uses: actions/checkout@v4
+      - uses: gradle/actions/wrapper-validation@v4
       - name: setup jdk
         uses: actions/setup-java@v4
         with:
           java-version: 21
           distribution: 'temurin'
-      - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v4
+      - uses: gradle/actions/setup-gradle@v4
+        with:
+          cache-read-only: ${{ github.ref != 'refs/heads/dev/1.13' }}
       - run: ./gradlew build check -x test --stacktrace --warning-mode fail
+      - uses: actions/upload-artifact@v4
+        with:
+          name: build-artifacts
+          path: |
+            build/
+            .gradle/
+            */build/
+            */.gradle/
+          retention-days: 1
 
   build_windows:
     if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.extended_tests == 'true' }}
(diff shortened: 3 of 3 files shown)


.github/workflows/publish-exp.yml
```diff
@@ -11,13 +11,14 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: gradle/actions/wrapper-validation@v4
       - name: setup jdk
         uses: actions/setup-java@v4
         with:
           java-version: 21
           distribution: 'temurin'
-      - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v4
+      - uses: gradle/actions/setup-gradle@v4
```
modmuss50, line 21: Never use caching for publish runs that contain the secrets, its not clear where it ends up.

.github/workflows/publish.yml
```diff
@@ -11,13 +11,14 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: gradle/actions/wrapper-validation@v4
       - name: setup jdk
         uses: actions/setup-java@v4
         with:
           java-version: 21
           distribution: 'temurin'
-      - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v4
+      - uses: gradle/actions/setup-gradle@v4
```
modmuss50, line 21: Same as above.

.github/workflows/test-push.yml
```diff
@@ -20,29 +20,58 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-24.04
+    env:
+      GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx4g -XX:+UseParallelGC
```
modmuss50, line 24: Why would you want to use the parallel GC!? no args best args imo (other than memory)

.github/workflows/test-push.yml
```diff
@@ -20,29 +20,58 @@ concurrency:
...
+      - uses: gradle/actions/wrapper-validation@v4
       - name: setup jdk
         uses: actions/setup-java@v4
         with:
           java-version: 21
-          distribution: 'temurin'
-      - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v4
-      - run: ./gradlew build check -x test --stacktrace --warning-mode fail
+          distribution: 'graalvm'
```
modmuss50, line 34: No, graal is going to cause all sorts of issues and make things harder to debug when they only fail on CI, lets test with what our users should be using.

.github/workflows/test-push.yml
```diff
@@ -20,29 +20,58 @@ concurrency:
...
+      - run: ./gradlew build check -x test --stacktrace --warning-mode fail --parallel --configuration-cache-problems=warn --build-cache --no-watch-fs
+      - uses: actions/upload-artifact@v4
+        with:
+          name: build-artifacts
+          path: |
+            build/libs/
+            build/classes/
+            */build/libs/
+            */build/classes/
+            .gradle/configuration-cache/
```
modmuss50, line 49: No!!! Even if loom was using the configuration cache it should never be shared, it likely includes all secrects on the machine (even the short lived github token used for the build). Gradle encrypts this by default to prevent misuse like this.

.github/workflows/test-push.yml
```diff
@@ -20,29 +20,58 @@ concurrency:
...
+          gradle-home-cache-cleanup: true
+      - run: ./gradlew build check -x test --stacktrace --warning-mode fail --parallel --configuration-cache-problems=warn --build-cache --no-watch-fs
+      - uses: actions/upload-artifact@v4
+        with:
+          name: build-artifacts
+          path: |
+            build/libs/
+            build/classes/
+            */build/libs/
+            */build/classes/
```
modmuss50, line 48: I dont think this applies to any files?

.github/workflows/test-push.yml
```diff
@@ -20,29 +20,58 @@ concurrency:
...
           java-version: 21
-          distribution: 'temurin'
-      - uses: gradle/actions/wrapper-validation@v4
-      - run: ./gradlew build check -x test --stacktrace --warning-mode fail
+          distribution: 'graalvm'
+          cache: 'gradle'
+      - uses: gradle/actions/setup-gradle@v4
+        with:
+          cache-read-only: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository }}
+          gradle-home-cache-cleanup: true
```
modmuss50, line 73: I cannot find any documentation for what this does on the action docs page: https://github.com/gradle/actions/blob/main/docs/setup-gradle.md

modmuss50 (changes requested):
I am strongly against using any form of caching, its a recipe for total disaster in a public repo. You have to be 100% sure you know what is being cached, when and how its used. There are too many parts for us to know that for sure.

In the loom tests we are testing Gradle/Loom, so we want things to re-run, it wouldnt be good if a change is made but the code is not re-ran as its using a cached result from a previous run. It also makes it hard to track down issues that only happen on a fresh run or due to a previous cache run.

This PR only caches the actual loom build, and not the loom test runs. I do think there is room for improvement (such as a caching proxy that the test runs can use) but it needs proper consideration.

Maybe its just becuase I'm used to long buids 1h-4h is normal for me, 20 mins doesnt seem to bad at all. This used to be a lot longer!

.github/workflows/test-push.yml
modmuss50: See: https://nvd.nist.gov/vuln/detail/CVE-2023-30853 you are opt-ing back into this vulnerability. This is fatal... if you do this on any other project consider it compromised.

I dont know where gradle caches the encryption key for the config cache, but assuming you have it working the key must be in the cache.

PR #1439, opened 2025-11-09
Optimize DeobfSpecContext by scanning configurations individually

zOnlyKroks:
## Summary
  Optimizes DeobfSpecContext.create() by scanning configurations individually instead of using the expensive getFullClasspath().getFiles() call.

  ## Changes
  - Remove getDependentMods() method that called getFullClasspath().getFiles()
  - Scan each configuration (runtime/compile, main/client) individually
  - Leverage existing AsyncCache to deduplicate artifact parsing across configurations
 
  ## Performance (this was tested in WSL, so maybe someone can verify this)
  - **Before:** ~9.8s with FAPI (97% spent in getFullClasspath().getFiles())
  - **After:** ~0.35s (27x faster)
  - Cache efficiency: 47 hits, 48 misses across 2 configs (98% hit rate on second scan)

  ## Possible Issues
  - Behavior is identical when all configs resolve to same artifacts
  - Edge case: If project deps are in one config but not another, they may not be included in transforming mods (but this matches the intended intersection logic)
  - AsyncCache uses file path as key - should work correctly for all cases

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/processors/speccontext/ProjectView.java
+++ b/src/main/java/net/fabricmc/loom/configuration/processors/speccontext/ProjectView.java
@@ -40,7 +40,7 @@ import net.fabricmc.loom.util.gradle.GradleUtils;
 
 // Used to abstract out the Gradle API usage to ease unit testing.
 public interface ProjectView {
-	// Returns a list of Loom Projects found in the specified Configuration
+	//Returns a list of Loom Projects found in the specified Configuration
 	Stream<Project> getLoomProjectDependencies(String name);
 
 	// Returns the mods defined in the current project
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/configuration/processors/speccontext/ProjectView.java
```diff
@@ -50,6 +50,8 @@ public interface ProjectView {
 
 	boolean areEnvironmentSourceSetsSplit();
 
+	Project getProject();
```
Juuxel: Is this used?

zOnlyKroks: ah no, that was a left over from when i was debugging

modmuss50 (approved):

PR #1467, opened 2025-12-06
Improve error message when reading AW with incorrect namespace.

modmuss50:


src/main/java/net/fabricmc/loom/configuration/processors/MinecraftJarProcessorManager.java
```diff
@@ -145,7 +145,7 @@ public void processJar(Path jar, ProcessorContext context) throws IOException {
 			try {
 				entry.processJar(jar, context);
 			} catch (IOException e) {
-				throw new IOException("Failed to process jar when running jar processor: %s".formatted(entry.name()), e);
+				throw new IOException("Failed to process jar when running jar processor: %s - %s".formatted(entry.name(), e.getMessage()), e);
```
Juuxel, line 148: This could also technically use ExceptionUtil but I feel a manual solution is better here. (It won't duplicate the IOException class name unlike ExceptionUtil)

Juuxel (approved):

PR #1470, opened 2025-12-09
chore: download with virtual thread

MukjepScarlet:
DownloadExecutor probably can use virtual thread too.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/download/Download.java
+++ b/src/main/java/net/fabricmc/loom/util/download/Download.java
@@ -46,10 +46,12 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.IntConsumer;
 import java.util.zip.GZIPInputStream;
 
+import org.jspecify.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/fabricmc/loom/util/download/Download.java
```diff
@@ -61,6 +63,7 @@ public final class Download {
 	private static final Logger LOGGER = LoggerFactory.getLogger(Download.class);
 	private static final Duration TIMEOUT = Duration.ofMinutes(1);
 	private static final HttpClient HTTP_CLIENT = HttpClient.newBuilder()
+			.executor(Executors.newVirtualThreadPerTaskExecutor())
```
modmuss50, line 66: Whats the benefit of this? Have you confirmed that its faster/better in any way?

We have already have:

https://github.com/FabricMC/fabric-loom/blob/0dbe660ed0d27cd8945fd196eaca6cc32d60b4d9/src/main/java/net/fabricmc/loom/util/download/DownloadExecutor.java#L40-L41

to limit the number of threads used for downloading. If a user has many cores (32+) its not helpful to use them all for downloads.

In https://github.com/FabricMC/fabric-loom/pull/1468 I tried using virtual threads but found it was much slower than regular threads, im not quite sure why tbh.

MukjepScarlet: The default executor of HttpClient (OpenJDK 21) is `Executors.newCachedThreadPool(new DefaultThreadFactory(id));`
As for your PR, I'm not sure you indicated to is `newWorkStealingPool` part or other executor. For this, I can only assume the internal method uses synchronized, which causes blocking on Java 21 (fixed on Java 25)

MukjepScarlet: https://github.com/ebarlas/project-loom-comparison/blob/1869f2084060e8757e470e8bb1ad84d4ccd9ad3e/src/main/java/loomtest/Frontend.java#L46

Here is a comparison.

MukjepScarlet: I just searched `.executor(Executors.newVirtualThreadPerTaskExecutor())`, many people use like this. By the way, I don't know why we need to limit the downloading thread count or concurrency unless we will be banned as spider software.

modmuss50 (approved):
I think this is fine in principle.

PR #1471, opened 2025-12-09
chore: remove all LinkedList usages

MukjepScarlet:
All of these usages are not used as Deque (for head-appending). In this case LinkedList will be slower than ArrayList.


modmuss50 (approved):
Sure, i guess this is fine. Likely doesnt make a lot of difference though.

PR #1482, opened 2025-12-22
Add task to validate mod-provided javadoc

Juuxel:
Probably needs some integration tests.

The setup is as simple as with the Enigma task: just register this task with the mapping file(s) in question, and add it as a dependency of `check`.


modmuss50:
Changes look good, please do add some form of testing. A collection of unit tests would likely be the best solution.

Juuxel:
There's a set of unit tests now.

modmuss50 (approved):

PR #1502, opened 2026-01-05
Use OS path resolution for xvfb-run executable and make useXvfb task property public

### LINKED ISSUE #1501: Don't hardcode xvfb-run binary path to `/usr/bin/xvfb-run`
The `AbstractRunTask` currently hardcodes the `xvfb-run` path to `/usr/bin/xvfb-run`: https://github.com/FabricMC/fabric-loom/blob/9e3837eedd03935321f4ee99aeb1ac481b76d846/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java#L215

This prevents using `xvfb-run` on platforms where it is installed to a non-default location. Examples:
- NixOS: Binary is somewhere in `/nix/store` and only appended to `PATH`
- Custom Xorg builds installed to `/usr/local`
- XQuartz for macOS

Another thing I noticed is that the task property to control xvfb usage is protected, and only set to true when the `CI` environment variable is set and it's running on Linux. I'd like it to be public so advanced users can control it themselves.

Suggested solution:
Just use the binary name (`xvfb-run`) and let the OS handle path resolution.

I'll create a PR to fix this shortly.

silenium-dev:
Changes:
- Use OS path resolution for xvfb-run executable by only specifying the executable name
- Make useXvfb task property public to allow users to use xvfb outside of the fixed convention (CI + Linux)

Closes #1501

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
+++ b/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
@@ -84,7 +84,7 @@ public abstract class AbstractRunTask extends JavaExec {
 	// We use a string here, as it's technically an output, but we don't want to cache runs of this task by default.
 	protected abstract Property<String> getArgFilePath();
 	@Input
-	protected abstract Property<Boolean> getUseXvfb();
+	public abstract Property<Boolean> getUseXvfb();
 
 	@Nested
 	@Optional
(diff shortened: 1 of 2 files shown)


modmuss50:
Hi, Thanks for the PR. Please could you re-target this against `exp/1.15`? I previously made a change to this in 1.15 so it will only use XVFB if it exists: https://github.com/FabricMC/fabric-loom/pull/1478

The changes look good otherwise.

silenium-dev:
Sure, one question though: would it be useful to make the xvfb path configurable instead of having a hard-coded executable name/path?

silenium-dev:
Replaced by #1508

PR #1507, opened 2026-01-06
Print Maven artifact URL on publish

hugeblank:
A small build script change that should make getting the correct maven artifact for a corresponding GitHub action build easier. Assuming a successful publish, the presence of `ENV.MAVEN_URL`, and `ENV.EXPERIMENTAL` being true; prints the URL to each artifact.

There is almost certainly a better way of doing this, but with my limited insight into GitHub actions, and how to take advantage of it within Gradle, this feels good enough.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -344,8 +344,7 @@ if (ENV.MAVEN_URL && ENV.EXPERIMENTAL) {
 						.append(plugin.id.replace(".", "/"))
 						.append("/")
 						.append(project.version)
-						.append("/")
-						.append("\n")
+						.append("/\n")
 			}
 			println("Published to:\n" + urls)
 		}
--- a/build.gradle
+++ b/build.gradle
@@ -333,6 +333,25 @@ publishing {
 	}
 }
 
+if (ENV.MAVEN_URL && ENV.EXPERIMENTAL) {
+	tasks.named("publish") {
+		doLast {
+			StringBuilder urls = new StringBuilder()
+			gradlePlugin.plugins.forEach { plugin ->
+				urls.append(ENV.MAVEN_URL)
+				if (!ENV.MAVEN_URL.endsWith("/")) urls.append("/")
+				urls
+						.append(plugin.id.replace(".", "/"))
+						.append("/")
+						.append(project.version)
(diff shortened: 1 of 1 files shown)


OroArmor:
1. Why? This can be done by just checking which action was run. Either the publish or publish experimental action. 

2. GHA sanitizes the output to not include secrets. Which you are then printing.

modmuss50:
No this isnt right. Ideally we should be tagging each commit with the version when publishing, as we do with fabric api and I think loader.

modmuss50:
Sorry, didnt mean to close the PR so abruptly. I think this is better if look into doing this, I think I can maybe use my publish plugin.

hugeblank:
> Why? This can be done by just checking which action was run. Either the publish or publish experimental action.

I don't understand what you mean by this. I tried finding any relation to the artifact from the actions and the best I could do was find the build number incrementing. Which *is* helpful, but that's also not the explicit version string, which was primarily what I wanted to get from the action. The link to the maven artifact was an extra bit that I could imagine being helpful, so I added it. 
 
> GHA sanitizes the output to not include secrets. Which you are then printing.

TIL, good feature.

hugeblank:
> Sorry, didnt mean to close the PR so abruptly. I think this is better if look into doing this, I think I can maybe use my publish plugin.

All good, this doesn't even really pass my own bar for what should be a contribution, and like I said in the initial message:

> There is almost certainly a better way of doing this...

modmuss50:
I dont think my publish plugin is sutible as we would need to create a release for every single build. The Gradle build likely needs to export the version number in such a way that the github action run can create a new tag.

OroArmor:
> I don't understand what you mean by this. I tried finding any relation to the artifact from the actions and the best I could do was find the build number incrementing. Which _is_ helpful, but that's also not the explicit version string, which was primarily what I wanted to get from the action. The link to the maven artifact was an extra bit that I could imagine being helpful, so I added it.

Normal: https://github.com/FabricMC/fabric-loom/actions/workflows/publish.yml

Experimental: https://github.com/FabricMC/fabric-loom/actions/workflows/publish-exp.yml

PR #1508, opened 2026-01-07
Use OS path resolution for xvfb-run executable and make useXvfb task property public

### LINKED ISSUE #1501: Don't hardcode xvfb-run binary path to `/usr/bin/xvfb-run`

silenium-dev:
Changes:
- Use OS path resolution for xvfb-run executable by only specifying the executable name
- Make useXvfb task property public to allow users to use xvfb outside of the fixed convention (CI + Linux)

Closes #1501
Replaces #1502

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
+++ b/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
@@ -63,10 +63,10 @@ import net.fabricmc.loom.configuration.ide.RunConfig;
 import net.fabricmc.loom.task.prod.TracyCapture;
 import net.fabricmc.loom.util.Constants;
 import net.fabricmc.loom.util.Platform;
+import net.fabricmc.loom.util.XVFBExistsValueSource;
 
 public abstract class AbstractRunTask extends JavaExec {
 	private static final Logger LOGGER = LoggerFactory.getLogger(AbstractRunTask.class);
-	private static final String XVFB_PATH = "xvfb-run";
 
 	@Inject
 	protected abstract ExecOperations getExecOperations();
@@ -149,7 +149,7 @@ public abstract class AbstractRunTask extends JavaExec {
 				getProject().getProviders().environmentVariable("CI")
 						.map(value -> Platform.CURRENT.getOperatingSystem().isLinux())
 						.zip(config, (enabled, runConfig) -> enabled && runConfig.environment.equals("client"))
-						.map(enabled -> enabled && Files.exists(Path.of(XVFB_PATH)))
+						.flatMap(enabled -> enabled ? XVFBExistsValueSource.exists(getProject()) : getProject().getProviders().provider(() -> false))
 						.orElse(false)
 		);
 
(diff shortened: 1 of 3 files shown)


silenium-dev:
@modmuss50 New PR, as it's not possible to change the source branch for an existing PR

src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
modmuss50: Will this change break this? I'm not sure we have any test coverage on this.

silenium-dev: Hmm, we probably need to remove/modify the presence check

modmuss50: Possibly yeah, its likely fine to remove as the line above checks to see if its a client run. The only issue is it is posisble to have a client run that can run headless (client datagen).

modmuss50: Another option would be to run `xvfb-run --help` to see if it exists, before starting the game.

silenium-dev: > Another option would be to run `xvfb-run --help` to see if it exists, before starting the game.

That would be my preferred solution, as it wouldn't sacrifice functionality

modmuss50: I hope you dont mind but I had a go at making this change, I was unable to commit it to your PR for some reason.

See: https://github.com/modmuss50/fabric-loom-1/commit/0162013103883f57ab0a552bae256bfefa8e0cc9

silenium-dev: Looks good, I'll apply that change on my branch (with credit of course).
I don't know though why you can't commit to my fork, maybe because it's not under my account, but under an org?

modmuss50: Thanks, my only guess would be becuase it wasnt a new branch?

deirn: It doesn't work for PR from organization.
https://github.com/orgs/community/discussions/5634

silenium-dev: I've pulled your commit into my fork

modmuss50 (approved):

PR #1512, opened 2026-01-19
fix Unpick constants for layered mappings with Unpick metadata V1

SpaceWalkerRS:
For regular mappings dependencies with Unpick metadata V1, Loom uses the maven dependency notation to construct the dependency for the Unpick constants. But this breaks when using layered mappings, since the mappings dependency points to the layered mappings file, not the original mappings dependency.

This PR addresses this by constructing a fallback constants dependency notation based off the maven dependency, and passing this to the `FileMappingsLayer`, where the Unpick metadata is patched before it is passed to the layered mappings processor.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java
@@ -165,6 +165,6 @@ public record LayeredMappingsFactory(LayeredMappingSpec spec) {
 		}
 
 		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_DEFINITIONS_PATH, unpickData.definitions());
-		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_METADATA_PATH, unpickData.rawMetadata());
+		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_METADATA_PATH, LoomGradlePlugin.GSON.toJson(unpickData.metadata()));
 	}
 }
--- a/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java
@@ -164,7 +164,9 @@ public record LayeredMappingsFactory(LayeredMappingSpec spec) {
 			return;
 		}
 
+		byte[] data = UnpickMetadata.toJson(unpickData.metadata()).getBytes(StandardCharsets.UTF_8);
+
 		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_DEFINITIONS_PATH, unpickData.definitions());
-		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_METADATA_PATH, LoomGradlePlugin.GSON.toJson(unpickData.metadata()));
+		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_METADATA_PATH, data);
 	}
 }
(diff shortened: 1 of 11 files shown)


src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java
```diff
@@ -165,6 +165,6 @@ private void writeUnpickData(LayeredMappingsProcessor processor, List<MappingLay
 		}
 
 		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_DEFINITIONS_PATH, unpickData.definitions());
-		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_METADATA_PATH, unpickData.rawMetadata());
+		ZipUtils.add(mappingsFile, UnpickMetadata.UNPICK_METADATA_PATH, LoomGradlePlugin.GSON.toJson(unpickData.metadata()));
```
modmuss50: Does this work? I think the `version` is never written.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpecBuilderImpl.java
```diff
@@ -82,6 +84,18 @@ public FileMappingsSpecBuilder containsAnnotations() {
...
+		if (fileSpec instanceof MavenFileSpec mavenFileSpec) {
+			String dependencyNotation = mavenFileSpec.dependencyNotation();
+			String[] notationParts = dependencyNotation.split("[:]");
+
+			if (notationParts.length == 4) {
+				dependencyNotation = dependencyNotation.substring(0, dependencyNotation.lastIndexOf(':'));
+			}
+
+			fallbackUnpickConstants = Optional.of(dependencyNotation + ":constants");
+		}
```
modmuss50, line 97: ~~This feels like a massive hack that isnt documented or tested. Would it not be better to add a method to FileMappingsSpecBuilder to specify the constants location.~~

Sorry I misread this.

modmuss50 (commented):
Really needs a test to make sure that this is working correctly, as im not sure it is atm. Let me know if you need some help with that.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpecBuilderImpl.java
```diff
@@ -82,6 +84,18 @@ public FileMappingsSpecBuilder containsAnnotations() {
 	@Override
 	public FileMappingsSpecBuilderImpl containsUnpick() {
 		unpick = true;
+
+		if (fileSpec instanceof MavenFileSpec mavenFileSpec) {
+			String dependencyNotation = mavenFileSpec.dependencyNotation();
+			String[] notationParts = dependencyNotation.split("[:]");
```
modmuss50: Ahhh this is a regex, can you just not do ":" ?

src/main/java/net/fabricmc/loom/configuration/providers/mappings/LayeredMappingsFactory.java
SpaceWalkerRS: Oops I never committed the fix for that. No it doesn't work as is.

src/main/java/net/fabricmc/loom/configuration/providers/mappings/file/FileMappingsSpecBuilderImpl.java
SpaceWalkerRS: Yes I suppose so. I'm used to adding the brackets for splitting on periods. I'll change that.

SpaceWalkerRS:
I made one further change, which is to always convert v1 metadata to v2 (for layered mappings), regardless of whether a fallback for constants is set. If there is no fallback, it just sets the constants to `null`.

I also added some tests in `LayeredMappingsUnpickTest`. Let me know if those are alright or if more is needed.

modmuss50 (approved):

PR #1514, opened 2026-01-23
Add plugin marker for repositories plugin

jpenilla:
The legacy `fabric-loom` plugin applies this when detecting it's applied to settings. It's cleaner to allow the user to explicitly apply it, rather than copy this functionality into the new plugins.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -241,6 +241,10 @@ gradlePlugin {
 			id = 'net.fabricmc.fabric-loom-companion'
 			implementationClass = 'net.fabricmc.loom.LoomCompanionGradlePlugin'
 		}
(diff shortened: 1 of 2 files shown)


modmuss50:
Looks good, just one quick question:

Does it matter that LoomGradlePlugin applies this via the class name:

https://github.com/FabricMC/fabric-loom/blob/eb157e466be88acc1d8330c0c3f8d17a172498f6/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java#L74

I assume Gradle handles this and wont try to apply the plugin twice?

jpenilla:
Yep, the id/marker is just a pointer to the class.

PR #1517, opened 2026-02-03
Handle perfered Minecraft version from constraints.

modmuss50:


modmuss50:
TODO before merging check this is the correct behaviour, see: https://docs.gradle.org/current/userguide/dependency_versions.html#sec:rich-version-constraints

Do we need to handle the other constraints such as `strictly` and `require`?

PR #1520, opened 2026-02-08
Enable stricter plugin validation

modmuss50:


src/main/java/net/fabricmc/loom/task/ValidateAccessWidenerTask.java
```diff
@@ -49,12 +53,15 @@
...
 
+@DisableCachingByDefault
 public abstract class ValidateAccessWidenerTask extends DefaultTask {
 	@SkipWhenEmpty
 	@InputFile
+	@PathSensitive(PathSensitivity.NONE)
 	public abstract RegularFileProperty getAccessWidener();
 
 	@InputFiles
+	@Classpath
```
Juuxel, line 62: You can remove the `@InputFiles` anno in this case as just having `@Classpath` there should suffice (also in the Kotlin service below and all the other tasks above that use `@Classpath` now). The javadoc of the annotation says:

> Note: to stay compatible with versions prior to Gradle 3.2, classpath properties need to be annotated with `@InputFiles` as well.

Loom clearly doesn't stay compatible with those versions :smile:

src/main/java/net/fabricmc/loom/task/GenerateSourcesTask.java
```diff
@@ -111,12 +114,15 @@ public abstract class GenerateSourcesTask extends AbstractLoomTask {
 	public abstract Property<String> getInputJarName();
 
 	@InputFiles // Only contains a single file
+	@Classpath
 	protected abstract ConfigurableFileCollection getClassesInputJar();
```
Juuxel: Not super related to the PR, but why is this `@InputFiles` as opposed to `@InputFile`?

src/main/java/net/fabricmc/loom/task/ValidateAccessWidenerTask.java
modmuss50: Will do, thanks for pointing that out 👍

src/main/java/net/fabricmc/loom/task/GenerateSourcesTask.java
modmuss50: Becuase `RegularFileProperty` has no way to lazily set the path from a `File`, it wants a `Provider<RegularFile>`.

Juuxel: You can convert `Provider<File>` to `Provider<RegularFile>` using [`ProjectLayout.file`](https://docs.gradle.org/current/javadoc/org/gradle/api/file/ProjectLayout.html#file(org.gradle.api.provider.Provider)).

src/main/java/net/fabricmc/loom/task/launch/GenerateDLIConfigTask.java
```diff
@@ -97,6 +101,7 @@ public abstract class GenerateDLIConfigTask extends AbstractLoomTask {
 	protected abstract Property<String> getDefaultMixinRemapType();
 
 	@InputFile
+	@PathSensitive(PathSensitivity.ABSOLUTE)
```
Toffikk, line 104: why the absolute? i assume only the contents matter here

src/main/java/net/fabricmc/loom/task/launch/GenerateRemapClasspathTask.java
```diff
@@ -37,14 +37,19 @@
...
 import org.gradle.api.tasks.TaskAction;
+import org.gradle.work.DisableCachingByDefault;
 
 import net.fabricmc.loom.api.RemapConfigurationSettings;
 import net.fabricmc.loom.task.AbstractLoomTask;
 import net.fabricmc.loom.util.Constants;
 
+@DisableCachingByDefault
 public abstract class GenerateRemapClasspathTask extends AbstractLoomTask {
 	@InputFiles
```
Toffikk, line 51: this should probably be @Classpath

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
```diff
@@ -73,8 +76,10 @@
 import net.fabricmc.tinyremapper.OutputConsumerPath;
 import net.fabricmc.tinyremapper.TinyRemapper;
 
+@CacheableTask
 public abstract class RemapJarTask extends AbstractRemapJarTask {
 	@InputFiles
```
Toffikk, line 81: also should be classpath to improve cache hits, and why NAME_ONLY?

src/main/java/net/fabricmc/loom/task/launch/GenerateRemapClasspathTask.java
modmuss50: No, as this task generates a file that includes the full path of these files.

src/main/java/net/fabricmc/loom/task/RemapJarTask.java
modmuss50: No, this includes the bundled jars with the jar name. We need this task to re-run if an input changes its name.

src/main/java/net/fabricmc/loom/task/launch/GenerateRemapClasspathTask.java
Toffikk: gotcha👍

src/main/java/net/fabricmc/loom/task/GenerateSourcesTask.java
lukebemish: You can also just call `fileProvider` on it: https://docs.gradle.org/current/javadoc/org/gradle/api/file/RegularFileProperty.html#fileProvider(org.gradle.api.provider.Provider)

modmuss50: Thats what I was after! What a silly name...

PR #1522, opened 2026-02-10
Stop using File collections for single files.

modmuss50:


src/main/java/net/fabricmc/loom/task/service/MappingsService.java
```diff
@@ -104,7 +104,7 @@ public static Provider<MappingsService.Options> createForRemapTask(AbstractRemap
 		final Project project = remapJarTask.getProject();
 
 		return project.provider(() -> {
-			if (remapJarTask.getCustomMappings().isEmpty()) {
+			if (remapJarTask.getCustomMappings().isPresent()) {
```
Toffikk: i think this is a regression and it should be fixed by having an `!` here, as previously it printed it when no custom mappings were defined and now with this change its the opposite

modmuss50: Yes, thanks. Thankfully all of the tests exploded as well :D

PR #1532, opened 2026-03-01
chore: Fix small typo on Intellij being Intelij

Awakened-Redstone:
Making a PR to fix 1 letter feels wrong


Awakened-Redstone:
huhh, that's a lot of test failures, but the logs don't seem to include loom anywhere
What just happened with them

PR #1539, opened 2026-03-27
Run tests on arm64 platforms.

modmuss50:
Extended tests run on both x64 and arm64 windows, reproducable build tests now only run on Java 25 but across all supported platforms.


modmuss50:
Arm64 windows issues are almost certinally caused by: https://bugs.openjdk.org/browse/JDK-8358645 nothing to do with loom.

PR #1555, opened 2026-05-15
Fix file name collision for artifacts with the same artifactId but different group

Eufranio:
When nesting jars, only the file name was being considered, causing jars with the same artifactId but different groups end up overwritten. This should fix the issue by adding a prefix to the generated file name in case of a name collision.


modmuss50:
Please target the `exp/1.17` branch.

PR #1557, opened 2026-05-16
feat: Add option to control registration of default run configurations

isXander:
This pull request adds support for disabling the automatic creation of default `client` and `server` run configurations (and their associated tasks) in Loom via a new Gradle property. It also introduces integration tests to verify this behavior and updates the documentation for the new property.

**New feature: Disable default run configs via Gradle property**
- Added the `fabric.loom.disableDefaultRunConfigs` property to `Constants.Properties`, allowing users to opt out of generating default `client` and `server` run configurations and their tasks. This is checked at plugin apply time to prevent the tasks from ever being registered. [[1]](diffhunk://#diff-8fc8f7ff8e3c01b1e4f4f5b068cc7988aa943b42a1ef52043ab2a4716ebd0156R171-R178) [[2]](diffhunk://#diff-d0fb60efa43f5e3c711a10dd42f532a263ef88523dfc6bbf07263a02dd99b068R216-R219)

**Task registration logic changes**
- Updated `LoomTasks.java` to conditionally skip creating the default run configs and associated tasks if the new property is set.
- Minor refactor to clarify logic for removing client/server run configs based on jar configuration. [[1]](diffhunk://#diff-d0fb60efa43f5e3c711a10dd42f532a263ef88523dfc6bbf07263a02dd99b068L227-L230) [[2]](diffhunk://#diff-d0fb60efa43f5e3c711a10dd42f532a263ef88523dfc6bbf07263a02dd99b068R242)

**Testing**
- Added integration tests in `RunConfigTest.groovy` to verify that the default run configs are registered by default and are properly disabled when the property is set.
- Imported the new property constant into the test file for use.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/LoomTasks.java
+++ b/src/main/java/net/fabricmc/loom/task/LoomTasks.java
@@ -213,29 +213,32 @@ public abstract class LoomTasks implements Runnable {
 			});
 		});
 
-		extension.getRunConfigs().create("client", RunConfigSettings::client);
-		extension.getRunConfigs().create("server", RunConfigSettings::server);
+		// Skip creating the default client/server run configs entirely when opted out via the gradle
+		// property. This must be checked at apply time so the corresponding tasks are never registered;
+		// once whenObjectAdded fires for a config, its task exists for the lifetime of the build.
+		if (!GradleUtils.getBooleanProperty(getProject(), Constants.Properties.DISABLE_DEFAULT_RUN_CONFIGS)) {
+			extension.getRunConfigs().create("client", RunConfigSettings::client);
+			extension.getRunConfigs().create("server", RunConfigSettings::server);
 
-		// Remove the client or server run config when not required. Done by name to not remove any possible custom run configs
-		GradleUtils.afterSuccessfulEvaluation(getProject(), () -> {
-			String taskName;
+			// Remove the client or server run config when not required. Done by name to not remove any possible custom run configs
+			GradleUtils.afterSuccessfulEvaluation(getProject(), () -> {
+				String taskName;
 
-			boolean serverOnly = extension.getMinecraftJarConfiguration().get() == MinecraftJarConfiguration.SERVER_ONLY;
-			boolean clientOnly = extension.getMinecraftJarConfiguration().get() == MinecraftJarConfiguration.CLIENT_ONLY;
+				boolean serverOnly = extension.getMinecraftJarConfiguration().get() == MinecraftJarConfiguration.SERVER_ONLY;
+				boolean clientOnly = extension.getMinecraftJarConfiguration().get() == MinecraftJarConfiguration.CLIENT_ONLY;
 
-			if (serverOnly) {
-				// Server only, remove the client run config
-				taskName = "client";
-			} else if (clientOnly) {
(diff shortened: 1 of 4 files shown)


src/main/java/net/fabricmc/loom/task/LoomTasks.java
```diff
@@ -214,29 +213,32 @@ private void registerRunTasks() {
...
-				return;
-			}
-
-			extension.getRunConfigs().removeIf(settings -> settings.getName().equals(taskName)
-					|| settings.getName().equals(taskName + "RenderDoc"));
-		});
+		// Skip creating the default client/server run configs entirely when opted out via the gradle
+		// property. This must be checked at apply time so the corresponding tasks are never registered;
+		// once whenObjectAdded fires for a config, its task exists for the lifetime of the build.
+		if (!GradleUtils.getBooleanProperty(getProject(), Constants.Properties.DISABLE_DEFAULT_RUN_CONFIGS)) {
```
modmuss50, line 219: Prefer early return.

modmuss50, line 219: This will break the Fabric API extension:

https://github.com/FabricMC/fabric-loom/blob/89c2717ee79720da9e1368399035efb23ab85c17/src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiDataGeneration.java#L96

modmuss50: This will break the sandbox configuration:
https://github.com/FabricMC/fabric-loom/blob/89c2717ee79720da9e1368399035efb23ab85c17/src/main/java/net/fabricmc/loom/configuration/sandbox/SandboxConfiguration.java#L93

modmuss50: This will also break the Fabric API extension here:

https://github.com/FabricMC/fabric-loom/blob/89c2717ee79720da9e1368399035efb23ab85c17/src/main/java/net/fabricmc/loom/configuration/fabricapi/FabricApiTesting.java#L110

modmuss50: This will break with versions older than 1.6:

https://github.com/FabricMC/fabric-loom/blob/89c2717ee79720da9e1368399035efb23ab85c17/src/main/java/net/fabricmc/loom/task/DownloadAssetsTask.java#L95-L97

modmuss50: This is not an easy change to make, a lot of things assume these run configs exist. Im not sure I tottaly agree that having this option is really what we want. Especially as there is already a supported way to safely disable/hide them.

modmuss50 (changes requested):

isXander:
Alright no worries. Didn't realise quite how much this broke. 

I will open an alternative PR that makes the added gradle property automatically *disable* and hide them, much like how it already does this with clientOnly(). Would this be considered?

modmuss50:
> automatically disable and hide them, much like how it already does this with clientOnly(). Would this be considered?

No, becuase you can already do that from your own buildscripts. We shouldnt have 2 ways to do the same thing.

PR #1559, opened 2026-05-16
fix: defer game JVM arg resolution to execution time

### LINKED ISSUE #1558: `runClient` fails when IntelliJ's coroutine debug agent is enabled in a Gradle composite build
Running `runClient` or `runServer` from IntelliJ in a Gradle composite build fails when
IntelliJ's "Enable coroutine agent" setting is enabled.

Tested with:
- Gradle 9.4.0
- Fabric Loom: 1.16-SNAPSHOT
- IntelliJ IDEA:  2026.1.2

**Error:**

    > Failed to query the value of property 'jvmArguments'.
       > Resolution of the configuration ':included-mod:runtimeClasspath' was attempted
         without an exclusive lock. This is unsafe and not allowed.

**What appears to happen:**

IntelliJ injects an init script (`ijKotlinCoroutineJvmDebugInit1.gradle`) that calls
`task.getJvmArgs()` during the `afterProject` callback to check whether
`-Dkotlinx.coroutines.debug=off` is already present.

In Gradle 9, `getJvmArgs()` bridges to `getJvmArguments().get()`, which forces evaluation
of the provider added in `AbstractRunTask`:

    getJvmArguments().addAll(getProject().provider(this::getGameJvmArgs));

`getGameJvmArgs()` calls `getInternalClasspath().getFiles()`, which resolves
`runtimeClasspath`. In a composite build, this crosses the included-build lock boundary
and triggers Gradle 9's unsafe-resolution guard.

**To reproduce:**

1. Create or open a Gradle composite build in IntelliJ:
   - the root build includes a Loom-based mod with `includeBuild("../included-mod")`

2. Ensure IntelliJ's coroutine debug agent is enabled:

   Settings → Build, Execution, Deployment → Debugger → Kotlin →
   Enable "Enable coroutine agent"

3. Attempt to run `runClient` from IntelliJ.

    ./gradlew runClient --dry-run

**Workaround:**

Disable IntelliJ's coroutine debug agent:

Settings → Build, Execution, Deployment → Debugger → Kotlin →
Disable "Enable coroutine agent"

**Possible fix:**

Changing the `getJvmArguments().addAll(provider)` call to use a JVM argument provider
appears to avoid forcing the classpath resolution during configuration:

    getJvmArgumentProviders().add(this::getGameJvmArgs)

This defers evaluation until task execution time, where classpath resolution should be safe.

`execWithXvfb()` would also need to include `getJvmArgumentProviders()` when manually
assembling the JVM command line, since it currently handles `getJvmArguments()` directly.

apmason13:
`AbstractRunTask` previously wired `getGameJvmArgs()` into `jvmArguments` via a provider:

    getJvmArguments().addAll(getProject().provider(this::getGameJvmArgs));

In Gradle 9, `getJvmArgs()` bridges to `getJvmArguments().get()`, which forces the
provider to evaluate. IDE tooling that calls `getJvmArgs()` during configuration
(e.g. IntelliJ's coroutine-debug init script) therefore triggers `getGameJvmArgs()` at
configuration time. That method resolves `runtimeClasspath`, which in a Gradle composite
build crosses the included-build lock boundary and fails with:

    Resolution of the configuration ':included-mod:runtimeClasspath' was attempted
    without an exclusive lock. This is unsafe and not allowed.

**Changes:**
- Move `getGameJvmArgs()` into `jvmArgumentProviders`, which Gradle only evaluates at
  execution time — not when `getJvmArgs()` is queried.
- Update `execWithXvfb()` to iterate `getJvmArgumentProviders()` alongside
  `getJvmArguments()`, since it manually assembles the JVM command line and would
  otherwise drop the argfile entry when xvfb is active.

Fixes #1558

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
+++ b/src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
@@ -147,7 +147,7 @@ public abstract class AbstractRunTask extends JavaExec {
 			}
 		});
 		getMainClass().set(config.flatMap(RunConfiguration::getDevLaunchMainClass));
-		getJvmArguments().addAll(getProject().provider(this::getGameJvmArgs));
+		getJvmArgumentProviders().add(this::getGameJvmArgs);
 
 		getInternalRunDir().set(config.flatMap(RunConfiguration::getRunDirectory));
 		getInternalEnvironmentVars().set(config.flatMap(RunConfiguration::getEnvironmentVars));
@@ -246,6 +246,11 @@ public abstract class AbstractRunTask extends JavaExec {
 		commandLine.add("--auto-servernum");
 		commandLine.add(javaExec);
 		commandLine.addAll(getJvmArguments().get());
+
+		for (CommandLineArgumentProvider provider : getJvmArgumentProviders()) {
+			provider.asArguments().forEach(commandLine::add);
+		}
+
 		commandLine.add(getMainClass().get());
 		commandLine.addAll(getArgs());
 


src/main/java/net/fabricmc/loom/task/AbstractRunTask.java
```diff
@@ -225,6 +225,9 @@ private void execWithXvfb() {
 		commandLine.add("--auto-servernum");
 		commandLine.add(javaExec);
 		commandLine.addAll(getJvmArguments().get());
+		for (CommandLineArgumentProvider provider : getJvmArgumentProviders()) {
```
modmuss50, line 250: Please make sure to build and test this, it wont pass checkstyle atm.

apmason13: sorry about that, fixed and pushed.

modmuss50:
I think this is acceptable, however we are currently working on the `exp/1.17` branch. I have rewritten a lot of this code, I believe this fix still applies but it may need some changes.

apmason13:
> I think this is acceptable, however we are currently working on the `exp/1.17` branch. I have rewritten a lot of this code, I believe this fix still applies but it may need some changes.

I've retargeted the PR to `exp/1.17` and reworked the fix against the rewritten code. Verified the unsafe-resolution error reproduces on the new branch and the fix resolves it.

modmuss50 (approved):

PR #1560, opened 2026-05-16
Implement per-jar include configurations

isXander:
### Changes

- ~~Added `loom.createIncludeConfigurations(SourceSet)`.~~
- Added `loom.nestJars(TaskProvider<? extends Jar>, Configuration)`
- Extracted all code that set up `include` & co into `IncludeConfigurations`
- Use the new `IncludeConfigurations` to create `include` for the main sourceset
- Implement unit tests

### Verification

- [x] I ensured that existing relevant tests relating to `include` still pass.
- [x] ~~Used same naming convention as `createRemapConfigurations` for public API.~~


modmuss50:
This doesnt make sense at all, the `include` configuration maps to a jar task, not the sourceset. There is a valid issue here, that needs to be fixed twice I think (obf and none obf).

I think for obf it needs to be something like:

```
remapJar {
includeConfiguration = configurations.myInclude
}
```

and none obf:

```
loom.incldueJars(myJarTask, configurations.myInclude)
```

This is just a rough idea off the top of my head, let me know if this doesnt seem correct.

isXander:
Implemented as you described.

```groovy
configurations {
	myInclude
}

loom.nestJars(tasks.customJar, configurations.myInclude)
```

You can also do `nestJars(myRemapJar, myInclude)` and it will auto-configure the RemapJar task specifically.

I have also removed the `includeInternal` configuration in favour of a resolveable detached configuration. This way the configuration is unnamed and can be programmatically created for every custom configuration given to `nestJars`. Behaviour is identical.

src/test/groovy/net/fabricmc/loom/test/unit/IncludeConfigurationsTest.groovy
```diff
@@ -0,0 +1,69 @@
...
+		IncludeConfigurations.nestJars(project, project.tasks.named("jar", Jar.class), customInclude)
+
+		then:
+		def processTask = project.tasks.findByName("processJarCustomIncludeJars")
+		processTask != null
+
+		def jar = project.tasks.named("jar", Jar.class).get()
+		jar.taskDependencies.getDependencies(jar).contains(processTask)
+	}
+}
```
modmuss50: This should use GradleTestUtil and isnt really testing anything at all.

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionApiImpl.java
```diff
@@ -605,5 +607,15 @@ public MixinExtension getMixin() {
...
+
+		@Override
+		public void nestJars(TaskProvider<? extends Jar> jarTask, Configuration configuration) {
+			throw new RuntimeException("Yeah... something is really wrong");
+		}
+
+		@Override
+		public void nestJars(TaskProvider<? extends Jar> jarTask, NamedDomainObjectProvider<? extends Configuration> configuration) {
+			throw new RuntimeException("Yeah... something is really wrong");
+		}
```
modmuss50, line 619: This is wrong, this causing a compile error is a real problem. You need to impliment these functions in LoomGradleExtensionApiImpl and not LoomGradleExtensionImpl

src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
```diff
@@ -0,0 +1,165 @@
...
+import net.fabricmc.loom.build.nesting.NestableJarGenerationTask;
+import net.fabricmc.loom.task.NestJarsAction;
+import net.fabricmc.loom.task.RemapJarTask;
+import net.fabricmc.loom.util.Strings;
+
+/**
+ * Sets up include configurations for the jar task that consumes them.
+ */
+public final class IncludeConfigurations {
+	private IncludeConfigurations() {
```
modmuss50, line 55: Instead of passing the project into each method, pass it via the constructor.

src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	public static void nestJars(Project project, TaskProvider<? extends Jar> jarTask, Configuration configuration) {
+		final String taskName = getProcessIncludeJarsTaskName(jarTask.getName(), configuration.getName());
```
modmuss50: Why does this not use getUniqueTaskName?

src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
```diff
@@ -0,0 +1,165 @@
...
+		nestJars(project, jarTask, configuration, taskName);
+	}
+
+	public static void nestJars(Project project, TaskProvider<? extends Jar> jarTask, Configuration configuration, String taskName) {
+		nestJars(project, jarTask, project.provider(() -> configuration), taskName);
+	}
+
+	public static void nestJars(Project project, TaskProvider<? extends Jar> jarTask, Provider<? extends Configuration> configuration, String taskName) {
+		final TaskProvider<NestableJarGenerationTask> processTask = createProcessTask(project, configuration, taskName);
+		final FileCollection outputJars = getOutputJars(project, processTask);
```
modmuss50: nit: dont make locals final.

src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
```diff
@@ -0,0 +1,165 @@
...
+			task.from(internalConfiguration);
+			task.getOutputDirectory().set(project.getLayout().getBuildDirectory().dir(task.getName()));
+			task.getUncompressNestedJars().set(extension.getUncompressNestedJars());
+		});
+	}
+
+	private static Configuration createInternalConfiguration(Project project, Provider<? extends Configuration> include) {
+		final Configuration internal = project.getConfigurations().detachedConfiguration();
+		internal.setCanBeConsumed(false);
+		internal.setCanBeResolved(true);
```
modmuss50: Use LoomConfigurations.Role.RESOLVABLE.apply

modmuss50 (commented):
generally looks fine, just a few points.

src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
isXander: Why? This class is a static utility class that is used across various points around Loom. Unless you want me to share the instance somewhere?

src/main/java/net/fabricmc/loom/extension/LoomGradleExtensionApiImpl.java
isXander: I don’t really understand the point of having Api, ApiImpl, and Impl. what is the point of ApiImpl?

The existing nestJars task that takes a file collection is implemented in the same place I implemented the new ones.

modmuss50 (approved):

PR #1562, opened 2026-05-21
Support GStrings in runs VM args

NerjalNosk:
Avoids a gradle crash when loading a loom project in IDEA with a GString in a run config's VM args

e.g. to base a VM arg off a project property

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
+++ b/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java
@@ -149,7 +149,7 @@ public class RunConfig {
 
 		// Custom parameters
 		runConfig.programArgs.addAll(settings.getProgramArgs());
-		runConfig.vmArgs.addAll(settings.getVmArgs());
+		runConfig.vmArgs.addAll(settings.getVmArgs().stream().map(Object::toString).toList());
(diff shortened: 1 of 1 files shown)


modmuss50:
Is this fixed in the exp/1.17 branch? I rewrote a lot of this code to use Gradle properties.

NerjalNosk:
> Is this fixed in the exp/1.17 branch? I rewrote a lot of this code to use Gradle properties.

I honestly only looked at the 1.16, not noticing the 1.17 one.

Not sure how to set up an environment with the 1.17, not sure how I can test, but from what I can see the vmArgs list is directly passed to the `ExecSpec` instance's `args` method, which seems to just accept objects of any object iterable, so likely calling `toString` internally. So all things considered, wouldn't be surprised if that wasn't an issue in `exp/1.17`

modmuss50:
Have you got an easy way to reproduce this? I can have a go, or maybe even add a test to cover it.

NerjalNosk:
Any `"whatever${something}"` in a run config's vmArgs will raise an issue (gstrings not being cast-able to string)

modmuss50:
If you do want to test this yourself set the version to `1.17.0-alpha.8`

NerjalNosk:
I can confirm my guess, it works just fine with loom 1.17

PR #1565, opened 2026-05-23
Initial provider migration of Configurations

Toffikk:
As per discussion on discord, this PR migrates `extendsFrom` calls to use Providers, along with a bit of other laziness improvements in regards to Configurations.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/mods/ArtifactRef.java
+++ b/src/main/java/net/fabricmc/loom/configuration/mods/ArtifactRef.java
@@ -29,13 +29,13 @@ import static net.fabricmc.loom.configuration.mods.ModConfigurationRemapper.repl
 
 import java.nio.file.Path;
 
+import org.gradle.api.NamedDomainObjectProvider;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.artifacts.ModuleDependency;
 import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.dsl.DependencyHandler;
-import org.gradle.api.provider.Provider;
 import org.jspecify.annotations.Nullable;
 
 import net.fabricmc.loom.util.Checksum;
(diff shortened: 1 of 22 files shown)


src/main/java/net/fabricmc/loom/configuration/mods/ArtifactRef.java
```diff
@@ -86,7 +87,7 @@ public void applyToConfiguration(Project project, Configuration configuration) {
 				moduleDependency.setTransitive(false);
 			}
 
-			dependencies.add(configuration.getName(), dep);
+			dependencies.add(configuration.get().getName(), dep);
```
isXander: pointless laziness to just `.get()`.

Use a `NamedDomainObjectProvider<? extends Configuration>` which includes `.getName()`

isXander (changes requested):
All usages of `Provider<Configuration>` should at *least* be changed to `Provider<? extends Configuration>`. This allows use of the new configuration types such as `ConsumableConfiguration`.

You also do a lot of resolution just to get the configuration name. I don't really know when you'd ever be given a plain `Provider` for a configuration. Use `NamedDomainObjectProvider` to get the configuration name without resolving it.

src/main/java/net/fabricmc/loom/configuration/mods/ArtifactRef.java
Toffikk: In some places, sadly, you can't use the `NamedDomainObjectProvider` type due to provider mutation resulting in the Configuration losing its named aspect, and becoming a `Provider`

Toffikk: I switched to the `NamedDomainObjectProvider` wherever it was possible without huge refactors, the main culprit for needing to use the plain provider is `sourceConfig`, using recursive copy which turns it into that

isXander (approved):

src/main/java/net/fabricmc/loom/configuration/mods/dependency/SimpleModDependency.java
```diff
@@ -30,17 +30,18 @@
...
+import org.gradle.api.provider.Provider;
 import org.jspecify.annotations.Nullable;
 
 import net.fabricmc.loom.configuration.mods.ArtifactMetadata;
 import net.fabricmc.loom.configuration.mods.ArtifactRef;
 
 // Single jar in and out
 public final class SimpleModDependency extends ModDependency {
-	private final Configuration targetConfig;
+	private final Provider<? extends Configuration> targetConfig;
```
modmuss50: Can this be a NamedDomainObjectProvider, you could then just do .getName() instead of having to resolve the whole provider in applyToProject. Same applies to SplitModDependency.

PR #1566, opened 2026-05-27
Be resilient to strict version resolution for Minecraft dependencies

isXander:
When someone does this:

```kt
dependencies {
	minecraft("com.mojang:minecraft") {
		version {
			prefer("26.1.2")
			strictly("[26.1.2,26.2)")
		}
	}
}
```

Loom previously tried to use the version range and failed to find a matching Minecraft version:

```
Caused by: java.lang.RuntimeException: Failed to find minecraft version: [26.1.2,26.2)
	at net.fabricmc.loom.configuration.providers.minecraft.MinecraftMetadataProvider.getVersionEntry(MinecraftMetadataProvider.java:130)
```

This PR tries to use `prefer` instead.

### Why do this in the first place?

Yes, doing the example above is stupid because strictly doesn't do anything.

But when using Gradle version catalogs:

```toml
[versions]
minecraft = { prefer = "26.1.2", strictly = "[26.1.2,26.2)" }

[libraries]
minecraft = { module = "com.mojang:minecraft", version.ref = "minecraft" }
```

You can then make use of the strictly version in other ways:

```kotlin
publishingMetadata {
    supportedMinecraftVersionRanges = libs.versions.minecraft.strictly.get()
}
```

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftMetadataProvider.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/MinecraftMetadataProvider.java
@@ -32,6 +32,9 @@ import java.util.List;
 import java.util.function.Function;
 
 import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.ExternalModuleDependency;
+import org.gradle.api.artifacts.VersionConstraint;
 import org.gradle.api.provider.Property;
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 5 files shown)


modmuss50:
This looks very similar to my PR: https://github.com/FabricMC/fabric-loom/pull/1517 I wasnt too sure it was the correct approach to take. It may well be better than nothing tho.

lukebemish:
Some notes on your impl (copied from my comment on discord):
- `version.startsWith("[") || version.startsWith("(");` is not enough to detect version ranges. `+` exists. This is Gradle, not Maven. You also should account for status-attribute-backed versions as dynamic versions (aka, `latest.release`) and probably just error if you see one, TBH. Version ranges can also start with `]`, not just `[` and `(`. See <https://docs.gradle.org/current/userguide/dependency_versions.html#sec:single-version-declarations>
- `prefer` is soft. If there is a non-dynamic `strictly` or `require` specified, `prefer` will be ignored. Your current impl is wrong -- it ignores `strictly` if there's a `prefer`. See <https://docs.gradle.org/current/userguide/dependency_versions.html#sec:rich-version-constraints>.

PR #1567, opened 2026-05-28
Use Gradle to resolve minecraft dependency

isXander:
- This PR allows Loom to resolve the Minecraft dependency via Gradle, rather than doing it itself.
- This allows for proper use of version ranges like `net.minecraft:minecraft:[1.21.1,1.21.4]`
- Loom publishes a metadata-only maven repository into global cache, so Gradle can resolve through it
- Loom uses Fabric Loader's `McVersionLookup` utility to normalise versions so Gradle can parse semver and use version ranges correctly
- A new `minecraftVersionNormalizerClasspath` configuration where you can add a future version of Fabric Loader onto it to get its normalization improvements for future version formats.
- Retain current behaviour on `com.mojang:minecraft` for backwards compatibility, since normalized versions may not match, moved to `net.minecraft:minecraft` for this new resolution behaviour.

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/FabricLoaderMinecraftVersionNormalizer.java
+++ b/src/main/java/net/fabricmc/loom/configuration/providers/minecraft/FabricLoaderMinecraftVersionNormalizer.java
@@ -1,0 +1,158 @@
+/*
+ * This file is part of fabric-loom, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2026 FabricMC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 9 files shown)


src/main/java/net/fabricmc/loom/configuration/providers/minecraft/FabricLoaderMinecraftVersionNormalizer.java
```diff
@@ -0,0 +1,158 @@
...
+	private FabricLoaderMinecraftVersionNormalizer(URL[] urls) {
+		this.classLoader = new FabricLoaderClassLoader(urls, FabricLoaderMinecraftVersionNormalizer.class.getClassLoader());
+
+		try {
+			MethodHandles.Lookup publicLookup = MethodHandles.publicLookup();
+			Class<?> lookup = this.classLoader.loadClass("net.fabricmc.loader.impl.game.minecraft.McVersionLookup");
+			this.getRelease = publicLookup.findStatic(lookup, "getRelease", MethodType.methodType(String.class, String.class));
+			this.normalizeVersion = publicLookup.findStatic(lookup, "normalizeVersion", MethodType.methodType(String.class, String.class, String.class));
+
+			Class<?> semanticVersion = this.classLoader.loadClass("net.fabricmc.loader.api.SemanticVersion");
```
modmuss50, line 69: Oh no, this is the eaxact thing I didnt want. Please discuss the approach before doing shit like this.

sylv256:
I'm not a fan of using `net.minecraft` to resolve Minecraft considering it's always been `com.mojang` and all of its dependencies are at `com.mojang`.

isXander:
> I'm not a fan of using `net.minecraft` to resolve Minecraft considering it's always been `com.mojang` and all of its dependencies are at `com.mojang`.

Agree, wanted to maintain backwards compatibility though as non-release versions are normalised and are now different to what they would be.

src/main/java/net/fabricmc/loom/configuration/providers/minecraft/FabricLoaderMinecraftVersionNormalizer.java
isXander: easily removed if you just use fabric loader on loom, this is so users can override fabric version.

maybe we could join the classpaths and use service loaders or something, idk

PR #1579, opened 2026-06-10
fix unresolved %ECLIPSE_PROJECT% reference in Eclipse run configs

SpaceWalkerRS:
The `%ECLIPSE_PROJECT%` variable was resolved before the last reference was added, leading to an error when running the config. Simply swapping two lines around fixes this.


modmuss50 (approved):
Good spot, thanks 👍

PR #1592, opened 2026-07-06
Add task for validating injected interfaces

Juuxel:
Mostly for use within Fabric API, but added here so anyone can benefit from it.

Also includes a new wrapper for Gradle's problem reporting API that
- hopefully wraps any breakages in a way that doesn't need changes in task classes
- lets tasks report multiple problems and throw one exception that lists all the problems at once for better UX
- lets us print GitHub Actions annotations directly from Gradle if requested.

## Example setup
```gradle
tasks.register('validateInjectedInterfaces', ValidateInjectedInterfacesTask) {
    modJar = tasks.named('jar').flatMap { it.archiveFile }

    // Optional, results in better error messages
    sourceRoots.from(sourceSets.main.java.srcDirs)

    // To enable GitHub Actions annotations for errors
    problemReportingOptions.displayGithubActionsAnnotations = true
}
```

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
+++ b/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
@@ -203,14 +203,23 @@ public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
 
 	private void checkInjectedInterface(byte[] classBytes, Consumer<Violation> violationConsumer) {
 		ClassVisitor visitor = new ClassVisitor(Opcodes.ASM9) {
-			private @Nullable String className;
+			private @Nullable String packageName;
+			private @Nullable String simpleClassName;
 			private @Nullable String sourceFile;
 
 			@Override
 			public void visit(int version, int access, String name, @Nullable String signature, @Nullable String superName, String @Nullable [] interfaces) {
-				// Strip the package to make the error messages more concise.
+				// Only the simple name is used in the error messages to make them more concise.
+				// The package is needed for resolving the source code file based on package + file name.
 				int slashIndex = name.lastIndexOf('/');
-				className = slashIndex >= 0 ? name.substring(slashIndex + 1) : name;
+
+				if (slashIndex >= 0) {
+					simpleClassName = name.substring(slashIndex + 1);
+					packageName = name.substring(0, slashIndex);
+				} else {
+					simpleClassName = name;
+					packageName = null;
+				}
 			}
 
 			@Override
@@ -223,7 +232,7 @@ public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
 			@Override
 			public @Nullable MethodVisitor visitMethod(int access, String name, String descriptor, @Nullable String signature, String @Nullable [] exceptions) {
 				if ((access & Opcodes.ACC_ABSTRACT) != 0) {
-					violationConsumer.accept(new Violation(className, name, descriptor, resolveSourceFile(className, sourceFile)));
+					violationConsumer.accept(new Violation(simpleClassName, name, descriptor, resolveSourceFile(packageName, sourceFile)));
 				}
 
 				return null;
@@ -232,14 +241,12 @@ public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
 		new ClassReader(classBytes).accept(visitor, ClassReader.SKIP_CODE | ClassReader.SKIP_FRAMES);
 	}
 
-	private @Nullable File resolveSourceFile(String className, @Nullable String sourceFileName) {
+	private @Nullable File resolveSourceFile(@Nullable String packageName, @Nullable String sourceFileName) {
 		if (sourceFileName == null) {
 			return null;
 		}
 
-		int slashIndex = className.lastIndexOf('/');
-		String directory = slashIndex >= 0 ? className.substring(0, className.lastIndexOf('/') + 1) : "";
-		String relativeSourcePath = directory + sourceFileName;
+		String relativeSourcePath = packageName != null ? packageName + File.separator + sourceFileName : sourceFileName;
 
 		for (File sourceRoot : getSourceRoots()) {
 			File sourceFile = new File(sourceRoot, relativeSourcePath);
--- a/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
+++ b/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
@@ -117,7 +117,7 @@ public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
 
 	public ValidateInjectedInterfacesTask() {
 		setGroup(JavaBasePlugin.VERIFICATION_GROUP);
-		problemReportingOptions = ProblemReportingOptions.createDefault(getProject());
+		problemReportingOptions = getProject().getObjects().newInstance(ProblemReportingOptions.class);
 
 		// Ignore outputs for up-to-date checks as there aren't any (so only inputs are checked)
 		getOutputs().upToDateWhen(task -> true);
--- a/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
+++ b/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
@@ -77,7 +77,7 @@ import net.fabricmc.loom.util.problem.ProblemReportingOptions;
  *
  * <p>{@snippet lang=groovy :
  * tasks.register('validateInjectedInterfaces', net.fabricmc.loom.task.ValidateInjectedInterfacesTask) {
- * 	modJar = tasks.jar.flatMap { it.archiveFile }
+ * 	modJar = tasks.named('jar').flatMap { it.archiveFile }
  * 	sourceRoots.from(sourceSets.main.java.srcDirs)
  * }
  * }
@@ -97,6 +97,8 @@ public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
 	/**
 	 * A collection of source code roots where the mod jar was built from.
 	 * This is used for resolving the corresponding source code files where report details are attached.
+	 *
+	 * <p>Adding source roots is optional. If not added, the file paths simply won't show up in error reports.
  	 */
 	@InputFiles
 	@PathSensitive(PathSensitivity.ABSOLUTE)
--- a/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
+++ b/src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
@@ -224,9 +224,7 @@ public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
 
 			@Override
 			public void visitSource(@Nullable String source, @Nullable String debug) {
-				if (source != null) {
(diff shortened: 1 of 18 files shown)


src/main/java/net/fabricmc/loom/util/problem/ProblemReportingOptions.java
Juuxel, line 1: This interface is technically part of the public API for tasks where it's used so I'm not sure if I should move it to the task package 🤔

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+		};
+		new ClassReader(classBytes).accept(visitor, ClassReader.SKIP_CODE | ClassReader.SKIP_FRAMES);
+	}
+
+	private @Nullable File resolveSourceFile(String className, @Nullable String sourceFileName) {
+		if (sourceFileName == null) {
+			return null;
+		}
+
+		int slashIndex = className.lastIndexOf('/');
```
modmuss50: Unless im silly, className will never have a / in it as its stripped above on line 213?

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+	public void problemReportingOptions(Action<? super ProblemReportingOptions> action) {
+		action.execute(getProblemReportingOptions());
+	}
+
+	@TaskAction
+	protected void check() throws IOException {
+		List<Violation> violations = new ArrayList<>();
+		Path modJar = getModJar().get().getAsFile().toPath();
+		FabricModJson fabricModJson = FabricModJsonFactory.createFromZip(modJar);
+		Set<String> injectedInterfaces = new HashSet<>();
```
modmuss50: As annoying as it is, I would perfer if you could use a worker executor. It then allows the task to run in parallel with other tasks within the same project. It seems we dont do it for the aw validation (we really should, I might change this in another PR).

If you really dont want to do this let me know and I can take a look, maybe after this is merged.

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+
+		// Look for injected interfaces in fabric.mod.json "custom" section
+		for (InterfaceInjectionProcessor.InjectedInterface injectedInterface : InterfaceInjectionProcessor.InjectedInterface.fromMod(fabricModJson)) {
+			injectedInterfaces.add(injectedInterface.ifaceName());
+		}
+
+		try (var zip = new ZipFile(modJar.toFile())) {
```
modmuss50: I dont think its safe to use ZipFile here, as it may cause race conditions with other threads opening the same jar. Use ZipUtils or FileSystemUtil.getJarFileSystem as this keeps a ref count of the open zips.

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+
+		try (InputStream in = zip.getInputStream(ctEntry)) {
+			ClassTweakerReader.create(visitor).read(in.readAllBytes());
+		} catch (IOException e) {
+			throw ExceptionUtil.createDescriptiveWrapper(UncheckedIOException::new, "Could not read class tweaker " + classTweaker, e);
+		}
+	}
+
+	private void checkInjectedInterface(byte[] classBytes, Consumer<Violation> violationConsumer) {
+		ClassVisitor visitor = new ClassVisitor(Opcodes.ASM9) {
```
modmuss50: Constants.ASM_VERSION

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+			@Override
+			public void visitSource(@Nullable String source, @Nullable String debug) {
+				if (source != null) {
+					sourceFile = source;
+				}
+			}
+
+			@Override
+			public @Nullable MethodVisitor visitMethod(int access, String name, String descriptor, @Nullable String signature, String @Nullable [] exceptions) {
+				if ((access & Opcodes.ACC_ABSTRACT) != 0) {
```
modmuss50: Modifier.isAbstract

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+			public void visit(int version, int access, String name, @Nullable String signature, @Nullable String superName, String @Nullable [] interfaces) {
+				// Strip the package to make the error messages more concise.
+				int slashIndex = name.lastIndexOf('/');
+				className = slashIndex >= 0 ? name.substring(slashIndex + 1) : name;
+			}
+
+			@Override
+			public void visitSource(@Nullable String source, @Nullable String debug) {
+				if (source != null) {
+					sourceFile = source;
```
modmuss50: is the source != null correct here? I imagine it doesnt matter but it doesnt feel right.

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+				return null;
+			}
+		};
+		new ClassReader(classBytes).accept(visitor, ClassReader.SKIP_CODE | ClassReader.SKIP_FRAMES);
+	}
+
+	private @Nullable File resolveSourceFile(String className, @Nullable String sourceFileName) {
+		if (sourceFileName == null) {
+			return null;
+		}
```
modmuss50, line 245: Objects.requireNonNull and throw? This should never happen.

src/main/java/net/fabricmc/loom/util/problem/Problem.java
```diff
@@ -0,0 +1,133 @@
...
+ * @param solution        a user-facing solution message, can be null
+ * @param fileLocation    the file where this error originated, can be null
+ * @param cause           a throwable cause, can be null
+ */
+public record Problem(ProblemId id, String message, @Nullable String details, Severity severity, @Nullable String contextualLabel, @Nullable String solution, @Nullable FileLocation fileLocation, @Nullable Throwable cause) {
+	public static Builder builder(ProblemId problemId) {
+		return new Builder(problemId);
+	}
+
+	public static final class Builder {
```
modmuss50, line 51: I wonder if a record + builder is the best solution when we have gradle's object factory. Would an interface with abstract properties be cleaner?

src/main/java/net/fabricmc/loom/util/problem/ProblemReportingOptions.java
modmuss50: Eh, leave it here. The task API surface is really hard to cleanly model.

src/test/groovy/net/fabricmc/loom/test/integration/InterfaceInjectionTest.groovy
```diff
@@ -112,4 +118,69 @@ class InterfaceInjectionTest extends Specification implements GradleProjectTestT
...
+
+ 			tasks.named('check') {
+ 				dependsOn 'validateInjectedInterfaces'
+ 			}
+			"""
+		new File(gradle.projectDir, 'src/main/java').mkdirs()
+		new File(gradle.projectDir, 'src/main/resources').mkdirs()
+		new File(gradle.projectDir, 'src/main/java/TestItf.java').text = INJECTED_INTERFACE_WITH_ABSTRACT_METHOD
+
+		if (useCt) {
```
modmuss50, line 144: Would it not be cleaner to just make 2 seperate tests at this point?

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
```diff
@@ -0,0 +1,257 @@
...
+public abstract class ValidateInjectedInterfacesTask extends DefaultTask {
+	private static final Logger LOGGER = LoggerFactory.getLogger(ValidateInjectedInterfacesTask.class);
+	private static final ProblemId ABSTRACT_METHOD_IN_INJECTED_INTERFACE = LoomProblems.problemId("abstract-method-in-injected-interface", "Abstract method in injected interface");
+
+	/**
+	 * The mod jar to check.
+	 */
+	@InputFile
+	@PathSensitive(PathSensitivity.NONE)
+	public abstract RegularFileProperty getModJar();
```
modmuss50, line 108: Should this have a default value of the `jar` / `remapJar` output?

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
Juuxel: Oops, I added the class name stripping after I added that method... Will split into `className` and `simpleClassName` to fix this.

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
Juuxel: Other threads won't be opening the same zip currently because it's not using WorkerExecutor (which I will make it use, which needs me to use FileSystemUtil anyway :smile:)

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
Juuxel: It doesn't really matter as this class visitor is only run once, but the check is technically pointless so I'll simplify the code and remove it.

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
Juuxel: It might happen with other JVM languages/compilers, mod code with stripped debug info (Proguard), or generated code. I can add a log warning if wanted, but I think throwing is a bad idea.

src/test/groovy/net/fabricmc/loom/test/integration/InterfaceInjectionTest.groovy
Juuxel: It's mostly the same code (the rest of the setup + all of when, then, where) which is why I thought one test method is better.

src/main/java/net/fabricmc/loom/task/ValidateInjectedInterfacesTask.java
Juuxel: Added. (Also added the main and client source sets to `sourceRoots` by default.)

src/main/java/net/fabricmc/loom/util/problem/Problem.java
Juuxel: Not necessarily considering there are a lot of util methods here and the lazy properties aren't needed as the record is consumed immediately without passing through Gradle systems where it needs to be serialised.

modmuss50: Fair enough, its not a major issue.

modmuss50 (approved):

PR #1596, opened 2026-07-10
Fix implementation from #1560 to allow configuration of dep resolution for included jars

lukebemish:
Use of a detached configuration as #1560 was implemented introduces a regression: it prevents any configuration of the resolution process by modders, including dependency locking, preventing changing/dynamic versions, adding attributes, or configuring handling of version or capability conflicts.

The ability to configure attributes of this configuration is necessary for, for example, [multiloader template](https://github.com/jaredlll08/MultiLoader-Template/blob/178d1abb85e7d4ef7903bdd8fe9f4f9e0748c80c/fabric/build.gradle#L33-L40), which uses the `io.github.mcgradleconventions.loader` attribute in resolvable configs to handle multiloader dependencies which distinguish loader by attribute.

However, the uses that use of a detached configuration rules out, that worked previously, are _much_ broader, and don't even have to invoke `includeInternal` by name, since the configuration in question is not affected by `configurations.configureEach` and thus any attempt to enforce defaults for resolution across all configs -- in terms of various resolution strategy rules or conflict resolution, preventing dynamic or changing versions, etc. -- will fail

### DIFF
--- a/src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
+++ b/src/main/java/net/fabricmc/loom/configuration/IncludeConfigurations.java
@@ -57,20 +57,22 @@ public final class IncludeConfigurations {
 
 	public static void nestJars(Project project, TaskProvider<? extends Jar> jarTask, Configuration configuration) {
 		String taskName = getUniqueTaskName(project, getProcessIncludeJarsTaskName(jarTask.getName(), configuration.getName()));
-		nestJars(project, jarTask, project.provider(() -> configuration), taskName);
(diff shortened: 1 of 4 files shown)


lukebemish:
For examples of the regression that don't involve references to `includeInternal` by name: previously, the following
```gradle
configurations.configureEach {
    resolutionStrategy.failOnChangingVersions()
}
```
or
```gradle
configurations.configureEach {
    resolutionStrategy.capabilitiesResolution.all { selectHighestVersion() }
}
```
or
```gradle
configurations.configureEach {
    resolutionStrategy.activateDependencyLocking()
}
```
would correctly configure the resolution of jar-in-jar jars from `include`. Now, they no longer do.

modmuss50 (approved):

PR #1597, opened 2026-07-11
perf: process included jars in batch

MukjepScarlet:
In my case (41 included jars) has about 10x speed increment.

### DIFF
--- a/src/main/java/net/fabricmc/loom/build/nesting/JarNester.java
+++ b/src/main/java/net/fabricmc/loom/build/nesting/JarNester.java
@@ -26,9 +26,11 @@ package net.fabricmc.loom.build.nesting;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.LinkedHashMap;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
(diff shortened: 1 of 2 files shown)


MukjepScarlet:
I have some other small changes want to apply to this file, should I append to this PR?

src/main/java/net/fabricmc/loom/build/nesting/JarNester.java
```diff
@@ -56,15 +58,31 @@ public static void nestJars(Collection<File> jars, File modJar) {
...
+				nestedJarEntries.put(nestedJarPath, Files.newInputStream(file.toPath()));
+			}
+
+			try {
+				ZipReprocessorUtil.appendZipEntries(modJar.toPath(), nestedJarEntries);
+			} finally {
+				for (var is : nestedJarEntries.values()) {
+					try {
+						is.close();
+					} catch (IOException ignored) {
```
Toffikk, line 78: u could use a simple `IOException _` instead of ignored

MukjepScarlet: Sadly it's a Java 25 feature...

Toffikk: oh i thought you targeted the exp/1.18 branch which is on java 25.

modmuss50:
Superseeded by my PR, we may want the split improvement later if there are still performance issues.

MukjepScarlet:
Okay I will test new `jar` task later, thank you!

PR #1599, opened 2026-07-12
Increase copy buffer to 8KiB in Download.java

MukjepScarlet:


modmuss50:
Have you noticed an issue to suggest we need this?

MukjepScarlet:
The 1KiB buffer can cause more disk write operations. This might be worse on the computer which has old disks. On my PC I haven't found noticable issues about it.

Basically 8KiB is the most common value in JVM libs including Java/KT stdlib and other IO libs, so I think it's a good practice.

modmuss50:
No, the network is almost certinally going to be slower. The write will first go to the file system cache in memory before ever hitting the disk. Unless you can prove that this makes a reasonable speedup I wont be merging this.

PR #1600, opened 2026-07-13
Microsoft Authentication

modmuss50:
This PR adds support for authenticating with your Microsoft account to play a fully licensed copy of the game in dev, nothing changes if you do not wish to login. The following steps can be taken to login:

1. run the `microsoftLogin` task
2. grant access via your OS to allow loom to store its encryption key in a secure location
3. perform the Microsoft device authentication flow
4. refresh token get securely saved

Any following client runs (via the gradle task) will use the refresh token to acquire a Minecraft access token. If this fails an error is logged but the game will still start. If you run the game often (I have been told once a week, but dont quote me) you should never need to go through the full login flow.

As far as I am aware this is the most secure implimentation of storing the secrect refresh tokens in any Minecraft launcher (please correct me if im wrong!). On windows we store an encryption key in the TPM and require user interaction (with an optional password) to read this key. The key is used to encrypt the refresh token along with some other data into a json file. On macos we use the system keychain. This is to try and make it much much harder (but not impossible) for malware to read these tokens and take over your account.

Running the `microsoftLogout` task will delete the refresh token and remove the encryption key from the TPM/keychain.

If it looks like enterprise java code, thats becuase I wanted to make sure we could unit test as much of this as possible. There is a lot of complexity here that cannot easily be tested end-to-end. Yes, I have used AI to help me write some of this (especially tests) but a vast amount of manual research, design and effort has gone into this, I have been thinking of adding this for a long while. Do please point out issues. Its only really useful now that there is an option for the intelij runs to use the Gradle tasks.

### DIFF
--- a/src/main/java/net/fabricmc/loom/task/LoomTasks.java
+++ b/src/main/java/net/fabricmc/loom/task/LoomTasks.java
@@ -45,6 +45,8 @@ import net.fabricmc.loom.configuration.providers.minecraft.MinecraftVersionMeta;
 import net.fabricmc.loom.task.launch.GenerateDLIConfigTask;
 import net.fabricmc.loom.task.launch.GenerateLog4jConfigTask;
 import net.fabricmc.loom.task.launch.GenerateRemapClasspathTask;
+import net.fabricmc.loom.task.launch.auth.MicrosoftLoginTask;
+import net.fabricmc.loom.task.launch.auth.MicrosoftLogoutTask;
 import net.fabricmc.loom.util.Check;
 import net.fabricmc.loom.util.Constants;
 import net.fabricmc.loom.util.LoomVersions;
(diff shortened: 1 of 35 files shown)


sylv256:
re: weird hardware i have a weird linux ideapad laptop (idk how old it is) that i could test this on. i think it has tpm since it ran windows 11 but i don't know if it'll even work on linux correctly.

modmuss50:
I haven't looked at Linux, I'm unsure if it needs a TPM or not. I mostly want to test on windows without a TPM.

CallMeEchoCodes:
linux will not need a TPM. the secret service should support any hardware

it may choose to use the TPM if available, that is up to the implementation

all the api specifies is that a secret will be stored securely

modmuss50:
Not possible, I have spent ages trying to create a MS oauth application, it then turns out I need to get that verified to be able  to do this for Minecraft. I really hate how locked down/undocumented this whole thing is. Sorry I tried, what a great waste of time.

modmuss50:
Microsoft have fixed their website, I now have a client id. Just need it approved by Mojang now.

PR #1605, opened 2026-07-17
JarNester performance improvement

modmuss50:
Same approach as https://github.com/FabricMC/fabric-loom/pull/1597 just a slightly different impl.

### DIFF
--- a/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
+++ b/src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
@@ -33,6 +33,7 @@ import java.nio.file.StandardCopyOption;
 import java.util.Calendar;
 import java.util.Comparator;
 import java.util.GregorianCalendar;
+import java.util.Map;
 import java.util.zip.CRC32;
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/loom/util/ZipReprocessorUtil.java
```diff
@@ -211,6 +212,45 @@ public static void appendZipEntry(Path file, String path, InputStream data) thro
 		Files.move(tempFile, file, StandardCopyOption.REPLACE_EXISTING);
 	}
 
+	/**
+	 * Appends files to a zip file, preserving the existing entry order and time stamps.
+	 * The new entries are appended in map iteration order with constant time stamps to ensure reproducibility.
+	 * This method should only be used when a reproducible output is required, use {@link ZipUtils#add(Path, Iterable)} normally.
+	 */
+	public static void appendZipEntries(Path file, Map<String, Path> paths) throws IOException {
```
LlamaLad7, line 220: Very late and nit but `SequencedMap` might be better

PR #1607, opened 2026-08-03
Don't regenerate layered mappings if they exist already

ScoreUnder:
I was having trouble with the Gradle configuration cache being invalid on every build due to the layered mappings being eagerly regenerated.

This guards the regeneration similarly to how we guard dependency remapping, which has fixed the configuration cache issue for me.


modmuss50 (approved):
Seems reasonable

modmuss50:
Humm, this seems to have broken the LayeredMappingsTests 🤔

PR #1618, opened 2026-08-15
Remove Mixin AP support

modmuss50:
Needs discussion if this is something we want to do right now, and if this is the right way to go about it. 

The entire mixin extension was marked as experimental, so removing it like this is technically correct but might be a bit extreme to do with no warning.


modmuss50:
We need a solution for Mixin Squared and other 3 part mixin extensions first. Runtime remapping is the main blocker behind removing this.

PR #1620, opened 2026-08-18
Add support for the shadow Gradle plugin

### LINKED ISSUE #762: Add support for shading libraries into the remapped jar.
I had a look at adding support for the very popular shadow plugin, however there are a number of major issues making it hard to recommend.

- By default it shades your entire runtime classpath, meaning you end up with Minecraft and a host of other things you really **DON'T** want in your mod jar.
- It does not enforce the use of the reallocate feature, meaning if you and another mod dont use it correctly (its easy to miss a package if you have many jars) you will end up with duplicate classes.
- It is not developed in a way that would allow Loom to tweak the defaults to better use the usecase, for example uses a `List<FileCollection>` [here](https://github.com/johnrengelman/shadow/blob/d8117e942b7d98e3274cb1cc9dbb4e4e978a2a58/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java#L34) instead of `ConfigurableFileCollection` that would allow loom to alter the convention value.
- I dont want to maintain support for this plugin in loom, it may become harder to support different version in the future.

For now using `include` is the better way to go, but I do see the need to shade libraries in the jar I suggest we add a shadow feature to the remap jar task, that automatically repackages the classes based on the modid. It should be none transitive to prevent you from accidentally including libraries that you dont want. 

If you must use the shadow plugin please make sure to check that you arent including any libraries that you aren't expecting, and ensure that **ALL** of the shaded classes have been repackaged. In most cases I highly reccomend using 'include' as it should cover most use cases.

NonSwag:
Closes #762

Since I am very unfamiliar with the groovy syntax, my clanker _helped_ me with the tests.
In addition to the tests, I manually verified everything still works as intended.

I am still working on getting JIJ and shadow to work together, so W.I.P.

### DIFF
--- a/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
+++ b/src/main/java/net/fabricmc/loom/LoomGradlePlugin.java
@@ -39,6 +39,7 @@ import net.fabricmc.loom.api.fabricapi.FabricApiExtension;
 import net.fabricmc.loom.configuration.CompileConfiguration;
 import net.fabricmc.loom.configuration.LoomConfigurations;
 import net.fabricmc.loom.configuration.MavenPublication;
+import net.fabricmc.loom.configuration.ShadowConfiguration;
 import net.fabricmc.loom.configuration.fabricapi.FabricApiExtensionImpl;
 import net.fabricmc.loom.configuration.ide.idea.IdeaConfiguration;
 import net.fabricmc.loom.configuration.sandbox.SandboxConfiguration;
@@ -95,6 +96,8 @@ public class LoomGradlePlugin implements Plugin<PluginAware> {
 		for (Class<? extends Runnable> jobClass : SETUP_JOBS) {
 			project.getObjects().newInstance(jobClass).run();
 		}
+
+		ShadowConfiguration.setup(project);
 
 		project.apply(Map.of("plugin", LoomCompanionGradlePlugin.NAME));
 	}
--- a/src/main/java/net/fabricmc/loom/configuration/LoomConfigurations.java
+++ b/src/main/java/net/fabricmc/loom/configuration/LoomConfigurations.java
@@ -31,6 +31,8 @@ import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ConfigurationContainer;
 import org.gradle.api.artifacts.dsl.DependencyHandler;
+import org.gradle.api.attributes.Attribute;
+import org.gradle.api.attributes.AttributeContainer;
 import org.gradle.api.plugins.JavaPlugin;
 
 import net.fabricmc.loom.LoomGradleExtension;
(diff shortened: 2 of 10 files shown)


modmuss50:
Can you explain what this does and why its needed? I dont think its really what I had in mind when I made the original issue.

NonSwag:
When using shadow in combination with Loom, it produces a rather unwanted output with the entire _decompiled_ Minecraft source code included in the final artifact.
When only using loom to build the artifact, it didn't include any libraries.

With this change it is possible to shade runtime dependencies into mods.

modmuss50:
My rough idea for this was to not depend on the shadow plugin, and use tiny remapper to handle the repacking. Loom would also have some custom logic to ensure that you dont end up shading libraries used by minecraft or loader.

Many other projects already use the shadow plugin with loom, I dont see what this PR is doing that you couldnt already do.

NonSwag:
I have not seen a single mod that really uses shadow and loom to produce an artifact with runtime libraries, nor was I able to get such a setup to produce proper outputs myself.
Could you link a project that does that?

NonSwag:
> My rough idea for this was to not depend on the shadow plugin, and use tiny remapper to handle the repacking. Loom would also have some custom logic to ensure that you dont end up shading libraries used by minecraft or loader.

I understand what you mean, but this sounds like a rather complex solution to a simple problem _in my opinion_.

Ever since I started working with loom it has been somewhat of a pain for me, just because it doesn't obey to the clearly defined dependency scopes.

https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management
https://gradleup.com/shadow/api/shadow/com.github.jengelman.gradle.plugins.shadow.tasks/-shadow-jar/configurations.html

The only _more or less_ reliable workaround to this that I could find (without relying on unofficial loom plugins, or publishing each library as a standalone mod) was to manually expand the source set of my mods with the sources of the given library.
This workaround however, is also quite annoying as I have to include all transitive dependencies manually and also only works with local modules that are already present in form of actual sources (not artifacts).

I am sure there are better, more reliable workarounds to this problem, but I do not care to invest more time into finding such solutions, if this could just be fixed/supported by loom itself.

modmuss50:
> to manually expand the source set of my mods with the sources of the given library.

This sounds like the correct solution to me, I dont see how this is a work around. 

I think the real issue the shadow plugin has defaults that are not a good fit for mods, but its easily configured to work. But you have to take special care to ensure that everything is repackaged and that you are not including any libraries that the game already provides.

Loom needs its own solution that solves the above two issues.

cputnam-a11y:
> I have not seen a single mod that really uses shadow and loom to produce an artifact with runtime libraries, nor was I able to get such a setup to produce proper outputs myself. Could you link a project that does that?

I have an example based on the fabric template mod if you wish to copy from there. Getting shadow working with loom really just requires flipping it into include mode vs exclude mode, and wiring included jar if unobf, or remapping if obf

PR #1623, opened 2026-08-24
Use Strict Gson for JSON serialization.

modmuss50:
Should help (but not tottaly prevent) mods from generating fmj's with "invalid" json that can be hard to parse in other languages.


modmuss50:
Going to leave this for now, lets support this fully for v1 jsons, when we do FMJ v2 we can parse them with strict mode here and in loader.

PR #1628, opened 2026-08-28
Fix: Strip whitespace from Mixin strings before looking them up in refmap.

LlamaLad7:
This matches Mixin's own behaviour when writing the refmap entries (https://github.com/FabricMC/Mixin/blob/0a83ef1f5621bd7db10c497665ec5990b80c8d45/src/main/java/org/spongepowered/asm/mixin/refmap/ReferenceMapper.java#L223)

There is not currently much reason to have whitespace in selectors but that will change with the new `->` stuff


LlamaLad7:
Failed test is due to Maven Central ratelimiting.

PR #1631, opened 2026-09-03
Adjust mod javadoc warnings based on where the javadoc came from.

modmuss50:


src/main/java/net/fabricmc/loom/configuration/processors/ModJavadocProcessor.java
```diff
@@ -121,6 +123,7 @@ public record ModJavadoc(String modId, MemoryMappingTree mappingTree, String map
 		@Nullable
 		public static ModJavadoc create(FabricModJson fabricModJson, MappingsNamespace productionNamespace, MarkdownJavadocOption markdownJavadocOption) {
 			final String modId = fabricModJson.getId();
+			final boolean remoteMod = fabricModJson.getSource() instanceof FabricModJsonSource.ZipSource;
```
modmuss50, line 126: I think this is fine, the other option is to pass the context into the source.

PR #1633, opened 2026-09-13
Key `JarManifestService` by project path instead of name

SoKnight:
`JarManifestService.get` registers its shared build service as `"LoomJarManifestService:" + project.getName()`. Build services are shared across the whole build, so two Loom projects with the same name under different parents, e.g. `:platform:fabric:v1` and `:platform:vanilla:v1`, end up with the same service. Whichever project registers it first supplies the parameters, and the other project's jar gets its `Fabric-Minecraft-Version`, `Fabric-Loader-Version` and `Fabric-Mixin-*`.

This PR keys the service by `project.getPath()`, which is unique within a build.

### Testing

Tested on a build with five Loom subprojects, where `fabric:v1`/`vanilla:v1` and `fabric:v2`/`vanilla:v2` have the same names. The vanilla projects have no Fabric Loader dependency.

| Project | Minecraft | Before: `Fabric-Minecraft-Version` / `Fabric-Loader-Version` | After |
|---|---|---|---|
| `:platform:vanilla:v1` | 1.12.2 | 1.14.4 / 0.16.0 (from `:platform:fabric:v1`) | 1.12.2 / unknown |
| `:platform:vanilla:v2` | 1.14.4 | 26.1 / 0.18.6 (from `:platform:fabric:v2`) | 1.14.4 / unknown |

The `fabric:*` projects are unchanged. Since the vanilla projects no longer borrow another project's data, they now log `Could not determine Mixin version for jar manifest`, the same as any project without Fabric Loader already does.

`./gradlew build -x test` passes. I didn't add an integration test because it would need two Minecraft setups; happy to add one if you want it.


modmuss50 (approved):
Great spot, thanks.

